Issue #129, opened 2019-05-29
Research API documentation tools


kreek:
## User Story
As a tools team member I want to research tools to create/generate and maintain beautiful OpenAPI (_or something else?_) docs so that they're easy to update and navigate.

## Acceptance Criteria
We have a path for:
  - Generating code from docs
  - Generating docs from code
  - Rendering docs that look nicer/more intuitive than swagger-ui (target [Stripe]
  - Research api doc retail tools/standards
  - Identify min AC for tool's output
(https://stripe.com/docs/api) quality)

PR #289, opened 2019-06-17
RFC: API docs

kreek:
RFC for documenting alternate solutions to our current workflow for generating API docs https://github.com/department-of-veterans-affairs/va.gov-team/issues/129


kreek:
@patrickvinograd does your team's API docs get generated via swagger blocks? Either way feel free to add anyone you think should comment on this.

patrickvinograd:
Nope we started writing them in yaml and have a controller that serves them as JSON. We lose some of the spec integration that way but 🤷‍♂ . I told the API team to take a look if they're interested

kfrz:
[Here's](https://github.com/department-of-veterans-affairs/vets.gov-team/issues/18618) some background discovery on OAS and what's broken from a standpoint of our swagger_blocks integration generating valid OAS (for documentation, tests, etc).

bastosmichael:
The idea of referencing a raw github file in the source for the swagger docs concerns me a bit, have you guys looked over all of the options available, @ATeal did some pretty in depth research into various options for us and can give you the scoop on other alternatives out there.

kreek:
@bastosmichael it doesn't have to be a github referenced file it could be relative to the project. That came out of my test to see if you could just sub in a JSON schema file for the schema and have it run in the swagger online editor (which it did) 😄- happy to learn how you guys reference schemas

ATeal:
So definitely agree that referencing a raw file on github doesn't feel optimal.

As far as our team, we have a hodgepodge, monolithic yaml files written in openapi spec 3, and some `swagger_blocks`. Neither of which is great.

I have two options on good solutions:

For the first I did a small proof of concept (I used our appeals api as an example):
1) Serve static json openapi spec files: 
```
        def swagger
          file = "#{params[:file]}.json"
          swagger = JSON.parse(File.read(AppealsApi::Engine.root.join('swagger/', file)))
          render json: swagger
        end
```
2) Organize your swagger files like so:
image

3) Reference sibling files like so:
```
"200": {
  "description": "Appeals retrieved successfully",
  "content": {
    "application/json": {
      "schema": {
        "required": [
          "data"
        ],
        "properties": {
          "data": {
            "$ref": "components.json#/schemas/Appeal"
          }
        }
      }
    }
  }
},

```
4) Load into swagger ui given your preferred displayer (An important note is that in this case we're depending on whatever you're using to load swagger based ui to know how to derefence)
  - For this I used our developer portal (developer.va.gov)
```
        name: 'Appeals Status',
        openApiDocUrl: `${
          process.env.REACT_APP_VETSGOV_SWAGGER_API
        }/services/appeals/docs/v0/testing.json`,
        shortDescription: 'Track appeals',
        urlFragment: 'appeals',
        vaInternalOnly: true,
```
5) Swagger ui hits the ref and assumes based off of the loading path + reference path that a file exists (in this case it hits `/services/appeals/docs/v0/components.json`)
6) UI is built and refs are dereferenced


---

A second option would be to preemptively dereference any `$ref`s in the json. There are implementations for this in other languages, and one exist in ruby, but it appears to have been stale for a long time: https://github.com/tzmfreedom/json_refs 

We could either update it, roll our own using other language implementations as an example etc. This would have the benefit of less calls when you load swagger ui, and not dealing with serving static assets leaving the logic in a rails controller.

I would say either of those two ways feels like the right call. It seems like the first way is the most standard, but I would say that either is an improvement on the current process.

kreek:
thanks @ATeal ! I like how option 1 works

PR #299, opened 2019-06-17
RFC: code isolation

jholton:
RFC for ideas around contractor code isolation and organization. #326


Practice Areas/Engineering/RFCs/API Code Isolation.md
```diff
@@ -0,0 +1,50 @@
+# RFC: Contractor code isolation and organization
```
gunsch: location: I don't love this "RFCs" folder, especially since the peer folders are "Backend", "Frontend", and "Infrastructure". maybe under "Backend/Notes", and date the file with YYYY-MM-DD for file ordering?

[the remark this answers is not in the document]
jholton: VSP engineers tasked with turning `vets-api` from a more traditional Rails API codebase into a "platform" to support multiple VFS teams working concurrently in the codebase.

jholton: Makes sense to me.  @kreek  will probably want to follow suit in his [API docs RFC PR](https://github.com/department-of-veterans-affairs/va.gov-team/pull/289)

patrickvinograd:
Other thoughts: You will likely need a custom generator instead of the default rails engine generator that generates a VSP module scaffold.

patrickvinograd:
More other thoughts (hit comment too soon):
- `current_user` is a good one to call out, also consider how to handle authorization - the pundit stuff  that's in place  is good but how do we ensure things stay aligned across modules.
- logging - figure out how to proactively drive teams towards effective logging (event-based logging?)
- Is some sort of namespacing/declaration of exported metrics useful?

kreek:
> What other changes do we need to consider to enable us to allow VFS teams to implement end points in self-contained Engines?

Early on we had thought about a SDK so the patterns we have are encapsulated and there’s less to keep track of/go wrong when wiring up an endpoint. Things like having the right middleware in the right order etc. However, we have a pretty aggressive timeline of real VSP teams hitting the platform by the end of the summer. Engines plus a generator or cloning a starter app can get us there quicker. The thinking is also to not let perfect be the enemy of the good. Pause on the SDK for now and get real developers building things so they can provide feedback (that informs what the SDK eventually looks like).

> How do we ensure that VFS teams are working within our established norms (semantics, error handling)?

For now have easy to follow docs and examples then enforce things via code review. Later a SDK can put up guard rails at the code level.

> Do we need to create a directory structure that separates VFS team work from Lighthouse API team work since by default all Rails Engines exist at the same level in the modules/ directory?

IMO it's fine as is. Code separation from the base app in the main concern. But I could see being split into internal and external dirs if we're worried about it getting unmanageable.

jholton:
Thanks for the comments.   I have updated the RFC to reflect them.

As an experiment, I made an attempt to get a current module (`vba_documents`) to be able to run specs from within it’s directory.
It just reinforced that our current path (given time constraints) is the way to go. (I gave up before I was successful)

Issue #108, opened 2019-05-29
[Discovery] Investigate options for access to internal tools


gunsch:
# Description

Currently, granting access to internal tools requires manually requesting access in Slack, with a lot of ops-team toil per-person per-service. We should investigate ways to simplify this process to reduce toil on ops-team members.

# AC
- [x] Enumerating all services that should be encompassed in onboarding, with criteria for which new members need access to which systems
- [x] Get buy-in from DSVA + VSP leadership in terms of what systems can be self-service, which can be exposed publicly, and which must be VA-private

wyattwalter:
Notes from previous ticket in vets.gov-team:

> After discussion with platform team after onboarding new devs during prototype phase [for compass project], we see the following improvements as warranted:
> - Make a public Jenkins instance for continuous integration (i.e. testing of branches), that doesn't require socks proxy access. There's no risk to doing this because our repos are already open source.
> - Make review instances accessible without requiring the socks proxy by proxying through paths under dev.vets.gov
> - Enable GitHub OAuth access for Sentry and Grafana - reducing the number of accounts that need to be set up and that a user needs to keep track of - they will have one-click log in based on their membership in the department-of-veterans-affairs Github org.

wyattwalter:
Submitted PR with an RFC for decision here. I'm leaning towards getting requests in for the tools at {tool}.vfs.va.gov internally so keep things cleanly separated, and using dev.va.gov/_vfs/review/{hash} for review instances.

wyattwalter:
re: the first requirement here:

Systems we need are:

- Jenkins
- Grafana
- Sentry

All of them would be required for VFS developers.

PR #310, opened 2019-06-18
RFC for internal tools access

wyattwalter:
Documentation for decision on access to tools meant for consumption by developers on the platform.

ref: https://github.com/department-of-veterans-affairs/va.gov-team/issues/108

### DIFF
--- a/Practice Areas/Engineering/RFC/001-internal-tools-access.md
+++ b/Practice Areas/Engineering/RFC/001-internal-tools-access.md
@@ -17,7 +17,11 @@ Below is a chart with tools we maintain today along with perceived risk profile 
 
 **Jenkins could be split into multiple parts to separate the risk profile, but it's being used as a deployment mechanism right now which raises the risk profile.
 
-## Proposal 1
+## Proposal
+
+While Option 2 outlined below has a higher number of cons, the items in that list are much smaller impact than those in the pro list, and we believe Option 2 should be the path forward.
+
+## Option 1
 
 Expose developer tools on https://dev.va.gov/_vfs/{toolname} ({toolname} here would be like "grafana", "sentry", etc).
 
@@ -33,8 +37,9 @@ cons:
 
 - not as clean a separation between lower and production environments (example: there's one grafana instance that can switch between dev, staging, and prod by design -- this could be undone but not without a major project)
 - risk of mistakes exposing something publicly that wasn't intended to be
+- potentially confusing behavior where a specific hostname may or may not act the same way between internal and external clients
 
-## Proposal 2
+## Option 2
 
 Setup internal-only hostnames for tools we use {toolname}.vfs.va.gov.
 
--- a/Practice Areas/Engineering/RFC/001-internal-tools-access.md
+++ b/Practice Areas/Engineering/RFC/001-internal-tools-access.md
@@ -1,0 +1,53 @@
+# Internal Tool Access
+
+The following tools are currently being used by developers and operators for the VSP, currently exposed by the SOCKS proxy. This proxy adds a good deal of toil for operations staff, is not scaleable for onboarding non-VSA engineers to the platform, and is unnecessary now that we have access to VAEC.
+
+Below is a chart with tools we maintain today along with perceived risk profile and exposure tolerance. Risk profile is meant to describe the amount of risk to Veterans and VA if information or access were compromised. Exposure tolerance is meant to describe whether VA might tolerate a service being exposed internally or externally from a compliance perspective.
+
+| Tool | Risk Profile | Exposure Tolerance |
+| --- | --- | --- |
+| Grafana / Prometheus | low | external |
+| Sentry* | high | internal |
+| Jenkins | high** | internal |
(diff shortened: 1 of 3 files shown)


Practice Areas/Engineering/RFC/001-internal-tools-access.md
```diff
@@ -0,0 +1,56 @@
...
+
+pros:
+
+- fewer SSL certificates to maintain
+- same URL structure for internal- and external-facing tools (ie, there's not an internal-only URL for one tool and a public URL for another)
+- simplified configuration for tools like Sentry where error reporting is sent through the publicly-exposed endpoint
+
+cons:
+
+- not as clean a separation between lower and production environments (example: there's one grafana instance that can switch between dev, staging, and prod by design -- this could be undone but not without a major project)
```
U-DON: What could be undone here? The ability to switch between environments?

Practice Areas/Engineering/RFC/001-internal-tools-access.md
```diff
@@ -0,0 +1,56 @@
...
+- risk of mistakes exposing something publicly that wasn't intended to be
```
U-DON: How easy is it to accidentally publicly expose an internal tool?

Practice Areas/Engineering/RFC/001-internal-tools-access.md
```diff
@@ -0,0 +1,56 @@
...
+
+**Jenkins could be split into multiple parts to separate the risk profile, but it's being used as a deployment mechanism right now which raises the risk profile.
+
+## Proposal
+
+While Option 2 outlined below has a higher number of cons, the items in that list are much smaller impact than those in the pro list, and we believe Option 2 should be the path forward.
+
+## Option 1
+
+Expose developer tools on https://dev.va.gov/_vfs/{toolname} ({toolname} here would be like "grafana", "sentry", etc).
```
annaswims: Did you pick `dev.va.gov` because its a subdomain already in place?  I'd expect `dev.va.gov` and `staging.va.gov` to have the same content, just different environments.  What about something like `devtools.va.gov`?

Practice Areas/Engineering/RFC/001-internal-tools-access.md
```diff
@@ -0,0 +1,56 @@
+# Internal Tool Access
+
+The following tools are currently being used by developers and operators for the VSP, currently exposed by the SOCKS proxy. This proxy adds a good deal of toil for operations staff, is not scaleable for onboarding non-VSA engineers to the platform, and is unnecessary now that we have access to VAEC.
+
+Below is a chart with tools we maintain today along with perceived risk profile and exposure tolerance. Risk profile is meant to describe the amount of risk to Veterans and VA if information or access were compromised. Exposure tolerance is meant to describe whether VA might tolerate a service being exposed internally or externally from a compliance perspective.
+
+| Tool | Risk Profile | Exposure Tolerance |
```
annaswims: Is authentication for these tools a separate discussion?  Sentry still has accounts for people who no longer work on VA projects, but it didn't matter because they couldn't reach the url without socks and account creation is wide open on Grafana, I think.

annaswims:
overall option 1 seems preferable to me but I have no strong opinion.

Practice Areas/Engineering/RFC/001-internal-tools-access.md
wyattwalter: We'd have to split up Grafana, Sentry, and Jenkins into multiple installs/instances to make it a bit cleaner. Right now they're all in a "utility" network that sort of runs disconnected from any of the live apps.

Practice Areas/Engineering/RFC/001-internal-tools-access.md
wyattwalter: In this case it would be one misconfiguration away. We can monitor for it and send an alert, but mistakes could happen. It's also possible that a change upstream of us at the TIC to cause our view of the real IP address of our client could cause a configuration we use to stop working correctly without us knowing (again, monitoring).

Practice Areas/Engineering/RFC/001-internal-tools-access.md
wyattwalter: Yeah, picked it because we already have the domain provisioned. Getting a new domain exposed is a good deal of work.

Practice Areas/Engineering/RFC/001-internal-tools-access.md
wyattwalter: Yeah, we're working on GitHub OAuth that will ensure a user is part of the VA org on GitHub. Grafana is now setup that way, we ran into some issues with Sentry due to what appears to be a bug in pyOpenSSL. We may get some pressure to attempt to get these setup with AD authentication, but I don't know if that's possible or what it looks like yet.

wyattwalter:
Decision after internal discussion on Slack as well was to use the `{tool}.vfs.va.gov` hostname and move forward with migration.

PR #602, opened 2019-07-16
Adding AWS console question to issue template

b00klegger:
Reworked the AWS question to include a descriptive paragraph regarding console access and to remove the either / or console vs ssh.


ricetj (approved):
LGTM

PR #652, opened 2019-07-23
Update OnCall.md

kfrz:
Add schedules, quicklinks


omgitsbillryan (approved):
LGTM.

Didn't like "OnCall"? Looks like it's been largely replaced by "on-call" 🙂

PR #721, opened 2019-07-28
remove copy of va-website

annaswims:
unclear why there was a copy of va-website in this repo, but it is unnecessary


cvalarida (approved):
Approving under the assumption this is the result of a `rm -rf Archive/va-website`; I didn't actually look at all the files removed :see_no_evil:

PR #905, opened 2019-08-09
Adds guidance about `va.gov-team-sensitive` repository to `va.gov-team`.

gunsch:
Adds guidance about `va.gov-team-sensitive` repository to `va.gov-team`.

This content is copied from https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/blob/master/README.md, with only minor changes at the top for context.


gunsch:
cc @leahbannon @batemapf for visibility

PR #911, opened 2019-08-12
Add prohibition on gmail.com addresses to template

batemapf:
As a analytics access requester, I need to know the type of email addresses for which I may not request access so I can gain access to GA faster and with less effort.

### DIFF
--- a/.github/ISSUE_TEMPLATE/analytics-access-request-template.md
+++ b/.github/ISSUE_TEMPLATE/analytics-access-request-template.md
@@ -10,7 +10,9 @@ assignees: joanneesteban, jonwehausen, nedierecel
 # Analytics Access Request
 
 ## User Email Address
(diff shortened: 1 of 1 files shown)


joanneesteban (changes requested):
_Provide email address(es) for user(s) requesting access to the Insights team on the DSVA Slack channel: #VSP-Analytics. Please also link this issue. Email addresses ending in `@gmail.com` _**may not**_ be used. If the user has a VA.gov or work email address that is not linked to a Google account, they will need to create one by visiting [this link](https://accounts.google.com/signup/v2/webcreateaccount?continue=https%3A%2F%2Faccounts.google.com%2FManageAccount&gmb=exp&biz=false&flowName=GlifWebSignIn&flowEntry=SignUp). VA.gov users and external users with email accounts that are not Google based should select the option "Use my current email address instead"._

_If the user already has a Google account that is not yet linked to their VA.gov email, they can link it using [these instructions](https://support.google.com/accounts/answer/176347?co=GENIE.Platform%3DDesktop&hl=en)._

batemapf:
@joanneesteban updated to match the text of your comment. Let me know if it is good to merge.

joanneesteban (approved):

joanneesteban:
Good to go, @batemapf !

PR #1080, opened 2019-08-19
Updating the Definition of done verbiage to better reflect its actual …

alexpappasoddball:
…purpose


lisac-usds (approved):
This makes sense to me!

One thing I noticed, on `epic-issue.md` - if it's the case that those Issues should be labeled with `Epic`, it's possible to have GitHub set that default, by adjusting this line: https://github.com/department-of-veterans-affairs/va.gov-team/pull/1080/files#diff-da9a58d17cd578829b71e6180f6c8c77R5

PR #1156, opened 2019-08-21
Update position on PII in GitHub

lisac-usds:
## Summary of changes

Removes language allowing special cases of PII in GitHub.

Why: it's my understanding that GitHub is approved at a low impact SaaS rating (LI-SaaS), and that PII cannot be stored in products with this rating. 

GitHub approved as LI-SaaS: 
  - https://github.blog/2018-10-24-github-is-fedramp-authorized/

Specs on LI-SaaS: 
  - https://tailored.fedramp.gov/policy/#fedramp-tailored-lisaas-requirements 
  - https://www.fedramp.gov/understanding-baselines-and-impact-levels

### DIFF
--- a/Administrative/Policies and work norms/sensitive-guidance.md
+++ b/Administrative/Policies and work norms/sensitive-guidance.md
@@ -10,7 +10,7 @@ All content related to development on VA.gov belongs in the public [`va.gov-team
 
 ### PII
 
-Veteran PII should generally not be checked into GitHub, but in some cases (i.e. investigating bugs related to specific users) can be checked in here if needed.
+Veteran PII should not be checked into GitHub. 
 
 [OMB memorandum M-07-16](https://www.whitehouse.gov/sites/whitehouse.gov/files/omb/memoranda/2007/m07-16.pdf) includes extensive details about securing PII, including this brief definition:
 


batemapf (approved):
LTGM. Thanks for catching this @lisac-usds!

gunsch:
Blah. So that means if we're investigating issues about a specific veteran, we can't (for example) even paste their email into an issue in a private repo?

That's still a use case we need to be able to handle, even if that's not in GitHub. I think we need a positive recommendation of where to put these things --- not just "don't put them in GitHub", but "put them [here] instead".

Has the team had another solution for this kind of use case in the past that we can reuse? Do we need to come up with something new?

lisac-usds:
Unfortunately for this PII sitn, I think the notion the content being stored on a non-VA system is the main consideration, more than whether a repo is private or public.  

Whether an email is PII - that's my read from this Apr 2010 [NIST writeup](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-122.pdf):  
> Examples of PII include, but are not limited to:
> - Name, such as full name, maiden name, mother‘s maiden name, or alias
> - Personal identification number, such as social security number (SSN), passport number, driver‘s license number, taxpayer identification number, or financial account or credit card number
> - Address information, such as street address or email address
> [...]

But if someone can find another position (or an update to this doc), that could change things.


> That's still a use case we need to be able to handle, even if that's not in GitHub. I think we need a positive recommendation of where to put these things --- not just "don't put them in GitHub", but "put them [here] instead".

+1.  I haven't been in situations where I needed to hold onto PII long-term, so I don't have recs there. But I'd think something like an issue tracker hosted on-prem or in VAEC might have more leeway than a LiSaaS. 

For sharing PII for short-term troubleshooting efforts, what I've done on other teams is use encrypted email, VA's Skype/Lync messaging, and plain phone calls.

gunsch:
> But I'd think something like an issue tracker hosted on-prem or in VAEC might have more leeway than a LiSaaS.

Do we have an existing issue tracker that meets these requirements? (maybe accessible via SOCKS proxy)

> For sharing PII for short-term troubleshooting efforts, what I've done on other teams is use encrypted email, VA's Skype/Lync messaging, and plain phone calls.

All of these options are super unstructured, though --- that is, I see them getting in the way more than helping for tracking/debugging over time if more people get involved. if people start saving these things to their own machines and sending them to each other over encrypted messaging systems, that feels like an even worse security posture than where we're starting from :/

lisac-usds:
@gunsch I agree with your points about the the workarounds - I'm sorry not to offer anything firm. :/  

> Do we have an existing issue tracker that meets these requirements? 

It's possible! I'll do some homework on this this week. @batemapf and @leahbannon are you aware of any self-hosted issue trackers used by the va.gov teams? 

Another idea: self-hosted sentry might capture details helpful for triage and might be accessible via socks proxy, and those links could be noted in any issue tracking done in GitHub.com.

leahbannon:
There's a little nuance to PII rules that may give us some space to work with.

Pulling from the [Lighthouse team's guidance](https://github.com/department-of-veterans-affairs/vets-contrib/blob/master/lighthouse/program/training/PIITraining.md#where-cant-it-go) -- a name/email alone, for example, does not constitute PII. They must be combined with "any other information that is linked or linkable to an individual, such as medical, educational, financial, and employment information." 

I'm doing some research into what constitutes "medical, educational, financial, and employment information." For example, I think it's possible that, in a private repo, an email address PLUS "they can't see their claims status" would be acceptable. Where unallowable would be posting email/name PLUS "their appeal related to diabetes from Agent Orange exposure isn't showing up."

Let me know if y'all know of resources on this or you think I'm way off.

PR #1436, opened 2019-08-29
Created discovery epic template.md

smoonpeyes:
Created a Discovery Epic template. Please let me know if you have feedback.


emilywaggoner:
Awesome! Looks great. My only question is on this sentence:

"Designers should also plan to do research readouts at sprint planning."

Should it be the end-of-sprint demo rather than planning?

PR #1467, opened 2019-09-02
Create 2019-09-02 Vets-Website CMS Integration

ncksllvn:


Practice Areas/Engineering/RFCs/2019-09-02 Vets-Website CMS Integration.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Broken-Link Checker
+
+__The broken-link-checker should no longer execute during a website build, but during a later stage of the CI.__ It should execute similarly to the `prearchive` stage of our CI, where the website has already been successfully built, and occur only when the CI is executing on the master branch. The outputted HTML is then processed and evaluated for broken links. We can summarize the result into a "report", with criteria that could include number of pages processed, number of links processed, any broken links found, etc. 
+
+We could also define some build-breaking criteria, where we have reached a certain threshold that our is determined by our build to not be deployable. For example, perhaps 100 broken links is _okay_ (considering the website will eventually be composed of many, many pages), but if there are 100 broken links to the same resource (such as `/health-care`), then this is not an acceptable state for the website to be deployed, because a key resource is missing. In this case, the build or deployment will fail.
```
cvalarida: Who would be responsible for determining these "allowable degradation" rules and thresholds?

I'm definitely not a fan of intentionally allowing any degradation, but I think this needs to be a DSVA decision.

Practice Areas/Engineering/RFCs/2019-09-02 Vets-Website CMS Integration.md
```diff
@@ -0,0 +1,85 @@
...
+
+Instead of failing a build when an accessibility issue has been saved into the CMS, we should introduce "warnings" into the CMS Preview Server - the CMS Preview Server injects and executes `axe-core`, then when an issue(s) is found, an alert banner is rendered to the author with the results. We could also enable [best practices](https://www.deque.com/axe/axe-for-web/documentation/api-documentation/) to help encourage higher quality accessibility practices than we do currently. 
+
+This proposal has already been captured into [issue 1794](https://github.com/department-of-veterans-affairs/va.gov-team/issues/1794).
+
+
+## Risks
+_What are the risks of the proposed changes?_
+
+The largest risk by far is degraded website quality overtime. If broken links become non-breaking and the parties responsible for maintaining the content aren't made aware, we could see broken links and small accessibility issues becoming widespread throughout the site.
```
cvalarida: Yeah, I feel like adding rules and thresholds to allow a certain level of "badness" into the site could put VA.gov in a position where the same broken links are never fixed, which essentially puts us back in the same situation that we're in now, but where the status quo is > 0 broken links and accessibility issues.

(I can probably expand on that if that doesn't make sense right now.)

cvalarida (commented):
If I understand it, the thrust of this RFC is to:
1) Move the CMS content validation into a different step (not the `Build` step) of the Jenkins pipeline
2) Allow for certain rules and thresholds (to be determined later) to decide whether the content is "good enough"
    - This is a change from our current rules and thresholds of "no broken links or accessibility issues"

The **Motivation** section does a good job of outlining the issues we face right now, but it could be clearer on why we want to allow a certain level of "badness" into the site. Just reading through this, I'm not thrilled with the idea of letting _any_ known bad links or accessibility issues in. Can you expand on why we would want to?

ncksllvn:
Closing. Per convo w/@cvalarida, VSP has their own vision to execute on, that will separate the concern of erroneous HTML to the content team.

cvalarida:
**Context:** VSP is probably going to split out the content building into a process outside of `vets-website`. I'm still drafting up an RFC for that right now, though.

Separating the build from the validation would be useful still, I'm sure, but that won't affect the vets-website <-> CMS integration at that point, since invalid content (broken links, etc.) won't affect vets-website deploys.

PR #1958, opened 2019-09-20
Draft RFC: Manage alerts from the CMS

batemapf:
## What is this?
A draft Request for Comment (RFC) regarding how to leverage the VA.gov CMS to quickly publish alert information. This is expected to be a starting point and is intended to reflect several recent discussions between the platform and CMS teams. This RFC *does not* include the related discussion of how email alerts will be tied to website alerts. Though we have discussed them in the same sittings, I believe they can (and should) be handled independently.

## How do I comment?
For specific comments, please make inline comments to the markdown file inside this PR. For general comments, please leave a comment on this PR. If you'd like to comment on other peoples' comments, please do.

## Do I need to comment right now?
No, but I'd like to get this wrapped up by 9/27, EOD, so if you are going to comment, please make some time to do so between now and then.

## What happens to my comments?
I'll incorporate any comments as best as I can. I may reach out for additional information from commenters. Once complete, I'll ask all commenters for their review/approval and then merge.

## ...then what?
I'll call a meeting of interested / necessary parties, make and document a decision, and then move on to the fun part 🎉

### DIFF
--- a/platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
+++ b/platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
@@ -57,7 +57,7 @@ Jenkins makes a GraphQL query the CMS content API based on given parameters, pul
 |Pros|Cons|
 |----|----|
 |Familiar architecture and process|Somewhat weighty process for minor updates|
-|Can apply existing tests (e.g. link checker, 508)|Inherently slower b/c of tests and checks|
+|Can apply existing tests (e.g. link checker, 508)|Publication is inherently slower b/c of tests and checks|
 |No additional client requests needed after page download|
 |Leverages benefits of GraphQL|
 
--- a/platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
+++ b/platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
@@ -79,6 +79,6 @@ A React component makes an asynchronous call with page-specific parameters to ap
 |----|----|
 |Supports decoupled approach|Big departure from current architecture and process|
 |Removes website/CMS interdependency if extended to all CMS data|Requires at least one more client request after page download
-|Allows generic content to be easily consumed by other applications||
+|Allows generic content to be easily consumed by other applications|Creates new threat vector for the CMS|
 |Doesn't require tracking of additional asset|
 |Leverages benefits of GraphQL||
--- a/platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
+++ b/platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
@@ -1,0 +1,84 @@
+# RFC: Manage alerts from the CMS
+
+Date: 2019-09-20
+
+## Background
+
+### Alerts, in general
+Veterans utilize VA.gov to learn about changes in operating status, event-based information, issues with tools and services on VA.gov, and other atypical interactions VA business lines. This includes everything from planned and unplanned changes in facility operating hours to form validation warnings.
+
+VA.gov users view this information as "alerts", which are presented in various ways across the site, including page-width banners, inline alert boxes, and dedicated operating status pages. There are at least four different ways alerts become available to the user on VA.gov:
+
+1. **Existing BRD process**. Included in page templates or React components that are delivered via the automated build, release, deploy (BRD) process (e.g. data validation alerts in a benefits application); [1]
+1. **Teamsite**. Published on legacy VA.gov pages via the Teamsite CMS (e.g. facility operating statuses);
+1. **Sourced from `va.gov-content` on load**. Rendered at page load with data pulled from the `va.gov-content` GitHub repository (e.g. the page-width [emergency banner on the homepage](https://github.com/department-of-veterans-affairs/vets-website/blob/225dcfc5cbfae4c28405e98300b66c8de734c16f/src/applications/static-pages/renderHomepageBanner.js)); and
+1. **Sourced from Vets API on load**. Rendered at page load with data pulled from Vets API(e.g. automated or scheduled [downtime alerts](https://github.com/department-of-veterans-affairs/vets-website/pull/10446/files#diff-85b5e6f872d2b0f86442f96d50218eab) for various VA backends)).
+
+With the exception of the Teamsite option for legacy pages, all of these implementations require an engineer to make a code change to update the content, behavior, or visibility of an alert. [2] Of course, Teamsite is not an option for updating alerts on Web Brand Consolidated (WBC) VA.gov pages and, as more content migrates to WBC VA.gov, Teamsite will become less relevant.
+
+### Specific case
+DSVA is working with the Pittsburgh VA Medical Center (PVAMC) to redesign the PVAMC facility pages on VA.gov. The redesign is being implemented in the Drupal-based VA.gov content management system (CMS). Once complete, PVAMC content managers will be able to utilize WYSIWYG tooling in the VA.gov CMS to make updates to the PVAMC facility pages, without engineer intervention. [3]
+
+A common and high priority PVAMC content management task is to update operating status pages and banners to inform users of changes in facility hours or provide emergency response information. Because this delivery of this information to users is time sensitive, content managers must be able publish updates in ten minutes or less.
+
+In addition to PVAMC, DSVA will be migrating the remaining ~170 VA Medical Centers' facility pages to the WBC VA.gov and training their respective content managers to use the VA.gov CMS. This significantly increases the likelihood that many similarly structured VA.gov pages derived from a shared content models and templates will be concurrently updated by disparate, isolated content managers across many VAMC's. This is particularly true in the case of regional or national emergencies.
+
+--
+
+[1] *These are not super relevant for this RFC, but worth noting because of sharing similar/same look and feel which may cause a Veteran to have similar expectations about how they might behave.*
+
+[2] *Technically, a non-engineer comfortable with git and markdown or YAML could update the GitHub-based homepage banner. But those folks are few and far between in the VA content manager ranks.*
+
+
+[3] *PVAMC's legacy VA.gov facility pages are managed via the Teamsite CMS.*
+
+--
+
+## Motivation
+To satisfy this user story:
+
+> *As a VA.gov content manager, I need to reliably publish content updates to VA.gov in less than ten minutes without relying on an engineer in order to deliver time-sensitive information to Veterans.*
+
+This functionality does not currently exist. Each of the alert implementations discussed above has significant limitations in attempting to satisfy the story:
+
+1. **Existing BRD process**. This takes more than ten minutes to run and rebuilds the entire site. Feels very weighty for what may regularly be a small update on a single page.
+1. **Teamsite**. This is going away once legacy VA.gov pages are migrated to WBC VA.gov or retired.
+1. **Sourced from `va.gov-content` on load**. Requires content managers to be comfortable with GitHub and markdown or YAML.
+1. **Sourced from Vets API on load.** No logical connection between the CMS and Vets API exists.
+
+## Possible design approaches
+
+For the first two possibilities, assume that a change to a specific area of the content model causes the CMS to tell Jenkins kick off a specific pipeline with certain parameters. For the third, Jenkins is out of the loop.
+
+### Single page BRD
+
+Jenkins makes a GraphQL query the CMS content API based on given parameters, pulls _only_ the relevant content and flags, builds _only_ the corresponding templates, runs _only_ the relevant tests and checks, and uploads the HTML asset to the existing S3 bucket, overwriting the existing asset. As soon as upload is complete, the page begins being served to users. Alerts are rendered along with the rest of the static content on the page.
+
+|Pros|Cons|
+|----|----|
+|Familiar architecture and process|Somewhat weighty process for minor updates|
+|Can apply existing tests (e.g. link checker, 508)|Inherently slower b/c of tests and checks|
+|No additional client requests needed after page download|
+|Leverages benefits of GraphQL|
+
+### Sourced from S3 bucket on load
+
+Jenkins makes a GraphQL query to the CMS content API based on given parameters, pulls _only_ the relevant content and flags, builds one or more JSON, YAML, or other type of structured data file, and uploads to a dedicated S3 bucket, possibly overwriting the existing asset. On page load, a React component makes an asynchronous call to the S3 bucket, parses the payload for any relevant alert information, and renders the result (if applicable).
+
+|Pros|Cons|
+|----|----|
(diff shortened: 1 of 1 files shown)


jefflbrauer:
Hi @batemapf - this is awesome! Thank you so much for outlining these approaches. I have 2 comments (that are really just regurgitating what @robbiethegeek-usds has said):

- We might want to add something about the fact that we have two distinct content types - Alert banners and Operating Status pages. It is possible we choose a different option per type. 

- The third option is certainly intriguing to me as it moves us towards a future of a completely dynamic site, however Robbie has warned another _con_ on this approach is that the ATO may need to be revisited, and/or a new one created. 

- We should include in here our options around Preview. I don't know if that needs to be completely separate options or if it should be baked into the pros/cons. Let's definitely make sure it gets included in the discussion either way. 

Thanks again!
Jeff

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
```diff
@@ -0,0 +1,84 @@
...
+
+Jenkins makes a GraphQL query to the CMS content API based on given parameters, pulls _only_ the relevant content and flags, builds one or more JSON, YAML, or other type of structured data file, and uploads to a dedicated S3 bucket, possibly overwriting the existing asset. On page load, a React component makes an asynchronous call to the S3 bucket, parses the payload for any relevant alert information, and renders the result (if applicable).
+
+|Pros|Cons|
+|----|----|
+|Lightweight and fast|Need to track another kind of asset outside of source control|
+|Leverages benefits of GraphQL|Adds more architectural and process complexity website management|
+| |Requires at least one more client request after page download|
+
+### Sourced from CMS content API on load
```
robbiethegeek: Missing a con for opening up the internal CMS to the outside work and creates a threat vector that hasn't be availabe til this point, a system that this team owns that is on the inside of the VA network or has access to it from the website via the public API.

jefflbrauer: Yes - this is what I was alluding to that would trigger the ATO in my second bullet :) Thanks Robbie!

batemapf: @robbiethegeek-usds is the threat markedly different than the one posed to any other internal VA system that is accessed through the vets-api layer?

robbiethegeek-usds: It is taking on a different risk profile since it is a system that will be owned by the VSP team that will be accessed from the outside.  It isn't a different risk but is for this team to take the responsibility for putting a system in the path that is owned by them.

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
```diff
@@ -0,0 +1,84 @@
...
+|Pros|Cons|
+|----|----|
+|Familiar architecture and process|Somewhat weighty process for minor updates|
+|Can apply existing tests (e.g. link checker, 508)|Inherently slower b/c of tests and checks|
+|No additional client requests needed after page download|
+|Leverages benefits of GraphQL|
+
+### Sourced from S3 bucket on load
+
+Jenkins makes a GraphQL query to the CMS content API based on given parameters, pulls _only_ the relevant content and flags, builds one or more JSON, YAML, or other type of structured data file, and uploads to a dedicated S3 bucket, possibly overwriting the existing asset. On page load, a React component makes an asynchronous call to the S3 bucket, parses the payload for any relevant alert information, and renders the result (if applicable).
```
swirtSJW: To minimize processing issues, it should use the format of the data that graphql provides.    It should not have to build or preprocess the data before saving.   The data provided by graph ql should just be saved and only replace a copy of the existing file.

Will need to add some simple logic to make sure that a "null response" does not cause an existing file to be replaced with an empty file.

swirtSJW: We should make sure the graphql response includes the time of the response to make it easy to investigate/determine if the pull system is failing.

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
```diff
@@ -0,0 +1,84 @@
...
+
+|Pros|Cons|
+|----|----|
+|Lightweight and fast|Need to track another kind of asset outside of source control|
+|Leverages benefits of GraphQL|Adds more architectural and process complexity website management|
+| |Requires at least one more client request after page download|
+
+### Sourced from CMS content API on load
+
+A React component makes an asynchronous call with page-specific parameters to api.va.gov (or some other middleware that shares network or environment access with the CMS content API) which, in turn, makes a GraphQL call to the CMS content API for _only_ the relevant content and flags. The payload is returned to the React component which renders the result (if applicable).
```
wyattwalter: The data from the CMS could be cached in the database for the API. This would be similar to the way we do maintenance windows from PagerDuty.

- We have a sidekiq (background) job that queries for maintenance windows from the PagerDuty API, stores them in the database.
- A react component makes a call to the API to load relevant data (does NOT pass through to the Drupal backend)

There's no reason that this would have to be done in realtime as described here unless i'm missing something.

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
```diff
@@ -0,0 +1,84 @@
...
+
+|Pros|Cons|
+|----|----|
+|Supports decoupled approach|Big departure from current architecture and process|
```
wyattwalter: Disagree that loading content from the API is a departure of the current architecture and process :)

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
wyattwalter: Perhaps ^^ should be a separate potential implementation?

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
```diff
@@ -0,0 +1,84 @@
...
+|Removes website/CMS interdependency if extended to all CMS data|Requires at least one more client request after page download
```
wyattwalter: What does it mean to "extend to all CMS data"?

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
robbiethegeek-usds: This is the first time the vets-api in production would talk to the CMS.  May need to update the ATO to acknowledge a new system integration and new data stored in the API (even though it is "cached" only meaning it is transitive and not a permanent source of truth or contain PII.

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
robbiethegeek-usds: Please see above comment, first time that CMS would be accessed from outside of the VA internal network.

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
robbiethegeek-usds: I think this was...make request to GraphQL for alerts + status pages + page dependency graph for each & the Granicus IDs and email payloads initially and pass along via parameterized builds.

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
```diff
@@ -0,0 +1,84 @@
...
+|Allows generic content to be easily consumed by other applications|Creates new threat vector for the CMS|
+|Doesn't require tracking of additional asset|
+|Leverages benefits of GraphQL||
```
wyattwalter: 🤔 GraphQL is great and all, but why is this a "pro"?

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
robbiethegeek-usds: Also not saying this is not possible or wrong, just pointing out that it is a change from what I know.

wyattwalter: Except for the content that we render to HTML and store in S3? I don't see that as being functionally different. If we cache the data, in the API it's just that we're storing content who's source of truth is the CMS in a new way. I don't feel the need to argue about this, but I disagree that it's a new thing and that's it's really not different than other models we have.

robbiethegeek-usds: The Jenkins server is within the VA infrastructure and that server is where the files are rendered and then pushed to S3. So with PIV / SSOi access on the CMS not sure if there is any implications for this since it wasn't a pattern or paradigm that @JeffBarnesUSDS or anyone discussed with the IAM team.

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
wyattwalter: A con here: the build process for vets-website already includes an implementation to pull from the CMS, we'd have to duplicate code in order to make this work via the API.

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Sourced from S3 bucket on load
+
+Jenkins makes a GraphQL query to the CMS content API based on given parameters, pulls _only_ the relevant content and flags, builds one or more JSON, YAML, or other type of structured data file, and uploads to a dedicated S3 bucket, possibly overwriting the existing asset. On page load, a React component makes an asynchronous call to the S3 bucket, parses the payload for any relevant alert information, and renders the result (if applicable).
+
+|Pros|Cons|
+|----|----|
+|Lightweight and fast|Need to track another kind of asset outside of source control|
+|Leverages benefits of GraphQL|Adds more architectural and process complexity website management|
+| |Requires at least one more client request after page download|
```
wyattwalter: A HUGE con to me is that this moves business logic into vets-website code.

wyattwalter: I say that because we also want to make sure that the alert gets published before the call out to send the email alerts.

I imagine other facilities will want to do something similar but probably not the same thing. They may have other email sending providers, want to tweet, etc. Do we need to have a generic "actions" editor for various facilities? And does having this encapsulated somewhere outside of the CMS make it more complicated?

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
wyattwalter: api.va.gov is also hosted within the VA network :)

authentication could be handled the same way as the vets-website builds pull content.

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
```diff
@@ -0,0 +1,84 @@
...
+|Leverages benefits of GraphQL|
+
+### Sourced from S3 bucket on load
+
+Jenkins makes a GraphQL query to the CMS content API based on given parameters, pulls _only_ the relevant content and flags, builds one or more JSON, YAML, or other type of structured data file, and uploads to a dedicated S3 bucket, possibly overwriting the existing asset. On page load, a React component makes an asynchronous call to the S3 bucket, parses the payload for any relevant alert information, and renders the result (if applicable).
+
+|Pros|Cons|
+|----|----|
+|Lightweight and fast|Need to track another kind of asset outside of source control|
+|Leverages benefits of GraphQL|Adds more architectural and process complexity website management|
```
wyattwalter: What benefits of GraphQL are we gaining here? The alerts in this implementation would be loaded from a JSON file from S3.

Or maybe this means that we're leveraging tooling already written in vets-website build process, just with a slightly different result?

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
wyattwalter: To tag onto the "requires at least one more client request" con, this also means that the page will likely render without the alert and be loaded in after the fact. Users will likely see that banner flash in and move the page a bit.

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
```diff
@@ -0,0 +1,84 @@
...
+For the first two possibilities, assume that a change to a specific area of the content model causes the CMS to tell Jenkins kick off a specific pipeline with certain parameters. For the third, Jenkins is out of the loop.
+
+### Single page BRD
+
+Jenkins makes a GraphQL query the CMS content API based on given parameters, pulls _only_ the relevant content and flags, builds _only_ the corresponding templates, runs _only_ the relevant tests and checks, and uploads the HTML asset to the existing S3 bucket, overwriting the existing asset. As soon as upload is complete, the page begins being served to users. Alerts are rendered along with the rest of the static content on the page.
+
+|Pros|Cons|
+|----|----|
+|Familiar architecture and process|Somewhat weighty process for minor updates|
+|Can apply existing tests (e.g. link checker, 508)|Inherently slower b/c of tests and checks|
```
wyattwalter:
```suggestion
|Can apply existing tests (e.g. link checker, 508)|Publication is inherently slower b/c of tests and checks|
```

wyattwalter: Adding this suggestion as performance of the site is unaffected.

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
rianfowler: what are granicus ids?

platform/engineering/request-for-change/2019-09-20-rfc-cms-based-alerts.md
gunsch: I'd say it's also a different risk, because "accessed via vets-api layer" means the add'l surface area is only whatever we decide to expose through vets-api, whereas if we make Drupal more publicly accessible then that likely means the _entire_ Drupal surface area is a potential for attack.

gunsch: missing a con for performance/latency/scalability: I don't think anyone here really wants Drupal in the _serving path_ of VA.gov, and that it'd likely be a regression (TBD how significant, but I expect a nontrivial one) compared to what we do today (loading everything from Drupal at build time, then serving it as static content).

batemapf:
Closing this as it was overcome by events :-)

PR #1463, opened 2019-08-30
Adds initial holiday release freeze schedule.

gunsch:
Initial proposal for dates for holiday release suspensions.

Note that I'm proposing a few days around Thanksgiving, as well as an extended two-week freeze over the holidays. I'm open to pushback here but I really think we should be avoiding while most people are out, and set the expectation to others not to be expecting to release functionality over the holidays.

PR #1987, opened 2019-09-23
updated deployment policy with holiday policy

accbjt:
Updated docs according to @gunsch holiday policy.

### DIFF
--- a/platform/working-with-vsp/policies-work-norms/deployment-policy.md
+++ b/platform/working-with-vsp/policies-work-norms/deployment-policy.md
@@ -1,26 +1,38 @@
 Last updated 8-30-19
 # Automated Deploys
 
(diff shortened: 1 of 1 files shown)


gunsch:
Ah hey, I think #1463 should actually cover this, but got dropped while I was on vacation

accbjt:
@gunsch Do we want to also include any retreat days? Since half of the team might be out during a retreat.

gunsch:
> @gunsch Do we want to also include any retreat days? Since half of the team might be out during a retreat.

Leaning no --- "half the team out" is very different than "100% of the team is out because holiday" (or likely 75%+ for Christmas week) in terms of coverage. There should still be oncall with plenty of engineers supporting in that kind of case.

PR #2103, opened 2019-09-30
Research practice area move

accbjt:
Moving the research documentation from vets.gov-team repo to va.gov-team.


mnorthuis:
@accbjt I haven't looked at PR in detail, but the description says we are moving files to va.gov-vfs-teams, which isn't accurate.  Looks like you did move them to the right place, put just wanted to clarify.

thatmattjohnson:
move completed and next steps defined per today's working session

Issue #1425, opened 2019-08-29
Discovery on feasibility of micro frontends


rianfowler:
## Overview

- [Micro frontends](https://martinfowler.com/articles/micro-frontends.html) are an approach to managing multiple front end apps in an ecosystem that prioritizes team autonomy
  - covers patterns that allow every team to have their own build / deployment pipeline 
  - addresses concerns with bundle size, css, etc 
- Problems this could address: 
  - reduce the overhead with managing `vets-website` front end platform by separating the front end apps from the platform (e.g. not using a monorepo)
  - allow apps teams to modify or introduce additional tooling, modules, or development flows without impacting the platform
  - remove the need for the platform team to review application specific code 

## Goal

- determine feasibility of micro frontend approach for `vets-website`

## Potential tasks 

- focus on process for accommodating design system- this specific challenge is not addressed in the micro frontends article
- determine how the platform group would manage critical metrics like performance, bundle size, user experience, test coverage

rianfowler:
@cvalarida Can you link to your design document here? I think this is the next step after separating the cms build.

timwright12:
closing since we're doing this as part of the app isolation discovery

PR #2150, opened 2019-10-01
Add CMS Integration RFC

cvalarida:
# Purpose
This RFC proposes a way to separate the task of building the content from the `vets-website` build to increase CI stability by isolating the effect of content-related issues such as broken links or Drupal instability.

The full RFC with diagrams, rendered tables, and everything can be found [here](https://github.com/department-of-veterans-affairs/va.gov-team/blob/cms-integration-rfc-cv/platform/engineering/request-for-change/2019-09-30-cms-integration.md). That might be a little easier to read.


rianfowler (commented):
got through most of it- will follow-up with an additional review on the rest.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,182 @@
...
+- Sitemap accessibility tests will be done during the content build,
+  not the `vets-website` build
+- The `vets-website` Jenkins build will have the following changes
+  - There will be no cache fallback
+  - There will be no content caching
+  - There will be no sitemap accessibility testing
+  
+### Serving React apps
+`vets-website` should be responsible for making the pages where the
+React apps live, but not responsible for building the header, footer,
```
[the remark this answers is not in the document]
rianfowler: we added the date hash for debugging. maybe we just hav ethe js apps insert an html comment / adds something to the global state that outlines their version / build date info.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,182 @@
...
+the sitemap.
+
+Whenever the accessibility check fails, the Jenkins build fails.
+
+When broken links are discovered during a normal Jenkins build, we
+send a Slack notification with a link to the build. If Jenkins was
+building the production branch, the build will also fail, to prevent
+broken links in production.
+
+## Motivation
```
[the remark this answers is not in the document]
cvalarida: :man_shrugging: This is the order [in the template](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/request-for-change/RFC-template.md).

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,182 @@
...
+
+![separate build diagram](images/separated-content-build-proposal.png)
+
+**Note:** Pulling the already-built content _can still_ have two
+different results, much like the current state, but since the content
+it's pulling will have already been validated, the likelihood of a
+failure is drastically reduced.
+
+### Changes
+- `vets-website` will no longer build the content
```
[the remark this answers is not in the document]
cvalarida: Yes. :sweat_smile:

platform/engineering/request-for-change/2019-09-30-cms-integration.md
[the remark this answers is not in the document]
cvalarida: >  I don't think we could get to a webpack only build if we have to pull in all of the content to inject our asset names into the markup.

I don't think we'd start with a Webpack-only `vets-website` build since it's a more drastic change, but I'd like to get there shortly after the separation. I don't know what that would look like right now, though, for the same reasons you mentioned.

> This means that the CMS would be soley responsible for managing the references to the webpack output: js / css / webfont files.

I don't think this would work. In my diagram, you can see I have the `vets-website` build pull the pre-built content. Because this content is pre-built and then saved in S3, it won't know anything about the references to the Webpack output. It also _shouldn't_ know about them. The separation of concerns goes both ways. I can definitely write more about this, but will defer that soliloquy for now. :stuck_out_tongue: 

> An alternative approach is to not create new asset names. The purpose of adding hashes to file names is for cache busting and for debugging (to a lesser extent). This means that the CMS would be soley responsible for managing the references to the webpack output: js / css / webfont files. We can cache-control header + etags to do cache busting. There's a slight performance trade off to use this too.

Adding hashes to the asset filenames aside, we'd still have to deal with adding `build/${env}/path/to/react/app/index.html` and ensuring it has the correct assets in that file. As I said above, this shouldn't happen during the CMS build, since the React apps are `vets-website`'s responsibility. Additionally, the paths to those applications live in their respective `manifest.json` files, so the CMS build would have to be aware of those, which then creates a fairly tight coupling between the two, which we're trying to avoid as much as possible.

That said, I think you're right about the cache busting. I'm not super familiar with the reasoning behind the way we do it now, but I like where you're going with that. It's probably tangential to this particular RFC, though, and can be done with or without this.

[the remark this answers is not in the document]
rianfowler: I forgot about the environment targeting. I guess that means there will always be a need for js build to inject something into the static markup. 

I guess we can profile it and see. My initial thought is that pulling in all of the content and adding the script and link tag is going to be time consuming. it would be nice if we could just fetch the content and use it as is- it would also be nice to not have to take an output from one process and put it into another. especially if that output rarely changes e.g. our js common bundle name has probably never changed besides the cache busting hashing.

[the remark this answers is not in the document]
cvalarida: Actually, now that I think about it, if we're not already injecting asset names into the HTML files during the Webpack step, we probably won't need to do it once we separate them. We currently don't have a separate Metalsmith plugin for this.

I haven't dug into our Webpack setup to figure out how this is handled right now. I was making some assumptions... :grimacing: 

I'll dig into this today.

rianfowler (approved):

rianfowler:
I think this is good- I would like to know what the next steps are. I assume that we can make a separate pipeline job for this for testing i.e. we don't have to fiddle with anything in the existing build process. I'm not sure but I think we can trigger jenkins builds off of a branch other than master. Either way, that might be a good thing to summarize. 

You may also want to add your own review to bring focus to the items you'd most like to have comments on. This is a lot of complicated stuff so it might be hard to get people to engage the whole thing.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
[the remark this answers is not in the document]
robbiethegeek-usds: Seems like over time there will be more and more apps that will be using the platform and shared common files (js) for auth / login as an example and likely leveraging existing code that is maintained by the platform team and (css) for shared common styles is something that would need to be managed as part of the platform and not have to be part of the new app teams boundary.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,226 @@
...
+    - Will not fall back to the cache
+  - Uses the JavaScript and CSS assets from the latest full build
+  - Runs a limited suite of tests
+
+### Metalsmith
+Currently, `vets-website`'s build is managed by Metalsmith, a
+plugin-based static site generator. Metalsmith uses a number of custom
+plugins used to manipulate the static content in various ways. One
+such plugin fetches the content from Drupal if we pass the
+`--pull-drupal` flag to the build script or if the local cache isn't
```
omgitsbillryan: What "build script"? Provide a cmd line usage example perhaps?

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,226 @@
...
+found, another checks for broken links, etc.
+
+This build script is run in a number of ways:
+- The Jenkins pipeline for a full build
+- The Jenkins pipeline for a content-only build
+- Locally, with `yarn build` or `yarn watch`
+
+Separate, but related, the preview server has a modified version of
```
omgitsbillryan: What's the preview server? (I _kind of_ know what it is, but maybe a screenshot, link to the server, or document link would help)

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,226 @@
...
+1. Because of this latest content, the `vets-website` `master` build
+   fails, but not because of any changes in `vets-website` code
+
+The goal of this RFC is to determine an approach for separating the
+content build from the `vets-website` Webpack build, so when
+`vets-website` pulls the content to serve up, it's already been built
+and validated. **Build failures due to content issues will be isolated
+to the content build and will not affect `vets-website` builds or
+deploys.** This will improve the overall stability of the
+`vets-website` CI.
```
omgitsbillryan: > This will improve the overall stability of the `vets-website` CI.

🎉

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,226 @@
...
+## Design
+_Explain the proposed changes in enough detail so that a team member
+with working knowledge of the project could implement the change
+themselves. Bonus points if this is easily mappable to user stories or
+specs._
+
+I propose rearchitecting the `vets-website` build to separate the
+static content building into a process distinct from the website.
+
+![separate build diagram](images/separated-content-build-proposal.png)
```
omgitsbillryan: What about going full "BRD" for content builds? 

Meaning, that when "Save to S3" happens, it's labeled in S3 with either with a commit SHA or with a release tag. Then in the `vets-website` build where you say, "Should pull in latest content", you could instead pass a commit SHA or a release tag as an input param. You could even have jenkins get the HEAD commit SHA on the `master` branch for content to "get the latest".

...All that said, I'm actually not even sure if content builds use GitHub 😬 , so I could be way off.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,226 @@
...
+failure is drastically reduced.
+
+### Changes
+- The `vets-website` build script will no longer build the content
+  - **Possible exception:** We may need to update some things in the
+    pre-built content such as references to the new cache-busted
+    filenames for JS and CSS assets
+  - **Possible exception:** It should still be responsible for building
+    the React pages, which would need to have the surrounding markup
+    come from the templates
```
omgitsbillryan: Formatting recommendation:
  - **Possible Exceptions**
    - We may need to update some things in the....
    - It should still be responsible for building...

platform/engineering/request-for-change/2019-09-30-cms-integration.md
cvalarida: This is a great question. The short answer is in the sentence following, but I realized yesterday that I'm actually not sure where content editors actually _go_ to see their changes. They're likely not running the preview server locally. :man_shrugging:

That said, it's mostly tangential to this RFC. Just thought it would be worth mentioning since it exists and is related.

I'll see if we have any documentation floating around for the preview server and link to it if we do. :+1:

platform/engineering/request-for-change/2019-09-30-cms-integration.md
cvalarida: Yeah, I'm not sure if GitHub releases could play a role in this process. But then, my understanding of BRD is pretty limited, I'll admit.

_Theoretically,_ we might be able to do something like:

**Building the content**
1. Every time a content build is triggered, generate a hash of all the content pulled in
1. Save the built content into S3 using the hash in the filename

**Building the website**
1. Every time a website build is triggered, look up the latest content hash
    - How? Where can we store it?
1. Use that hash to grab the latest content

I like being explicit with what we're putting into the build like that, but I don't understand the benefit or the cost yet.

Also keep in mind that we'll need to be able to grab the latest content locally, not just within the context of Jenkins. Not sure how that'll affect the implementation.

This is exactly the kind of feedback I'm looking for, though. Thank you! :raised_hands:

platform/engineering/request-for-change/2019-09-30-cms-integration.md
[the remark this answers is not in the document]
rianfowler: @robbiethegeek-usds yeah I think that's the direction we're moving. we have a [task](https://github.com/department-of-veterans-affairs/va.gov-team/issues/1425) to investigate the micro frontend [pattern](https://martinfowler.com/articles/micro-frontends.html) that thoughtworks is pushing. I think they do a good summary of the tradeoffs. 

I think the ideal is for the apps teams have their own deployment pipeline and use common dependencies provided by the platform (maybe through an npm module). There's a lot to work through to get to that point though including defining / enforcing core quality metrics (e.g. performance budgets).

I think that autonomy includes static content deployment but I'm not sure what that looks like. Right now this process is managing a giant graphQL query and there's a lot of `content model -> template model` transformations that don't seem very clearcut / clean (i.e. easy to separate into team owned deployment pipelines). I think separating the process from the js build is a first step to supporting the principles in that micro frontends article but there's a lot of unknowns specific to `vets-website` we need to work through.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,230 @@
...
+| Accessibility | True | True | False |
+
+The Metalsmith script performs the check for broken internal
+links. This happens before the accessibility tests in the Jenkins
+build pipelines.
+
+When broken links are discovered during a full Jenkins build, we send
+a Slack notification with a link to the build. If Jenkins was building
+the production (`master`) branch, the build will also fail, to prevent
+broken links in production.
```
gunsch: maybe add this info to the table? (i.e. does the build fail on errors?)

platform/engineering/request-for-change/2019-09-30-cms-integration.md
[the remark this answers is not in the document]
gunsch: if it improves clarity, don't feel you have to follow the template. it's there as a starting point / guideline.

or add a summary of this motivation at the top.

the fact that you have a line below that says "the point of this RFC is..." is a good hint that it should be at the top ;)

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,230 @@
...
+## Design
+_Explain the proposed changes in enough detail so that a team member
+with working knowledge of the project could implement the change
+themselves. Bonus points if this is easily mappable to user stories or
+specs._
+
+I propose rearchitecting the `vets-website` build to separate the
+static content building into a process distinct from the website.
+
+![separate build diagram](images/separated-content-build-proposal.png)
```
gunsch: if it's not too much effort to do,  this diagram might be a lot clearer with a similar "current build process" diagram, to visually see what's changing

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,230 @@
...
+and validated. **Build failures due to content issues will be isolated
+to the content build and will not affect `vets-website` builds or
+deploys.** This will improve the overall stability of the
+`vets-website` CI.
+
+## Design
+_Explain the proposed changes in enough detail so that a team member
+with working knowledge of the project could implement the change
+themselves. Bonus points if this is easily mappable to user stories or
+specs._
```
gunsch: delete?

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,230 @@
...
+- The `vets-website` Jenkins build will have the following changes
+  - There will be no cache fallback
+  - There will be no content caching
+  - There will be no sitemap accessibility testing
+  
+### Serving React apps
+`vets-website` should be responsible for making the pages where the
+React apps live, but not responsible for building the header, footer,
+and other surrounding markup. To make this possible, **I propose we
+have the content build produce an empty React page which the
```
gunsch: I don't really understand this section.  what is a "React page"?

"an empty React  page" --> one such page that gets reused for multiple React apps? or one for each React app?

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,230 @@
...
+`vets-website` can copy into each place a React app lives.** The new
+page would then be updated with references to the correct JavaScript
```
gunsch: how/when would said updating be performed?

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,230 @@
...
+
+The empty React page would then be deleted or otherwise prevented from
+being served to production.
+
+### Beneficial side effects
+- The `vets-website` build will be faster both locally and in Jenkins
+  - Currently, performing transformations to the content accounts for
+    about half of the time in `yarn build` / `yarn watch` when the
+    cache is used
+  - Pulling the content from Drupal currently takes upwards of a full
```
gunsch: reword as a benefit (it'll be sped up? how much?)

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,230 @@
...
+    minute
+	- This is because Drupal has to parse the _massive_ GraphQL query,
+      put together the result, and send back the ~3.7MB response
+  - The sitemap accessibility tests will be performed in the content
```
gunsch: why is this a benefit?

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,230 @@
...
+the deploy may not have all the latest changes.
+
+## Risks
+_What are the risks of the proposed changes?_
+
+- The current integration is _so_ complex; it would be easy to
+  overlook something while implementing the changes
+
+## Questions
+- Who owns the content build repo?
```
gunsch: make a suggestion?

platform/engineering/request-for-change/2019-09-30-cms-integration.md
cvalarida: I added a glossary term for "React app page" and (hopefully :crossed_fingers:) clarified this a bit by using the terms "template" and "landing page."

The idea here is that the content build makes a template which the `vets-website` build can use to make landing pages for all its applications. This ensures:
- The content build doesn't have to know about the React apps
- The `vets-website` build doesn't have to know about the header, footer, or any other page chrome; its only concern is updating the references to the individual JS bundles and CSS files to bootstrap the application
  - I still have to dig into how we do this right now, though, to ensure we don't have a cleverer solution already

platform/engineering/request-for-change/2019-09-30-cms-integration.md
cvalarida: Done! I put it in the **Background** section, though. I can move it if desired.

cvalarida:
Thanks for the comments so far, @rianfowler, @omgitsbillryan, and @gunsch! I made a bunch of changes that hopefully clarify the things you all mentioned.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
cvalarida: Done!

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,230 @@
...
+`vets-website`.
+
+### Content validation
+Some basic validations are performed on the content from the CMS and
+vagov-content repo.
+
+| Validation | Full build | Content-only build | Preview server |
+| ---------- | ---------- | ------------------ | -------------- |
+| Broken link checking | True; will fail the build on the `master` branch only | True; will fail the build | False |
+| Accessibility | True; will fail the build | True; will fail the build | False |
```
1Copenut: @ncksllvn and @1Copenut are working on a way to add axe errors and warnings to the CMS preview server: https://github.com/department-of-veterans-affairs/vets-website/pull/10836

I think this inline error checking should go forward, regardless of combined/split build tasks.

cvalarida: Agreed. That's a great addition to the preview server! As it stands now, though, the preview server doesn't do that checking, right?

1Copenut: @cvalarida Correct, as it stands right now, the preview server does not do this checking.

cvalarida: Cool. I'll go ahead and leave that out of this for now, then, since I want it to reflect the current state. This RFC shouldn't affect adding that functionality.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,237 @@
...
+  - Has the normal header and footer
+  - Points to the appropriate assets for the application
+
+### Metalsmith
+Currently, `vets-website`'s build is managed by Metalsmith, a
+plugin-based static site generator. Metalsmith uses a number of custom
+plugins used to manipulate the static content in various ways. One
+such plugin fetches the content from Drupal if we pass the
+`--pull-drupal` flag to the build script (`yarn build --pull-drupal`)
+or if the local cache isn't found, another checks for broken links,
```
rianfowler:
```suggestion
or if the local cache isn't found. Another plugin checks for broken links,
```

patrickvinograd:
My question was I think along the same lines as Bill's - is there typically only one HEAD version of the content that the vets-website build should use, or is there a need to identify/build against multiple versions. What about for deployment - is there ever a need to do a coordinated deploy of both components that would be difficult under this scheme?

platform/engineering/request-for-change/2019-09-30-cms-integration.md
patrickvinograd: I don't it's essential that the content release artifact (e.g. a tarball) be stored in GitHub (as opposed to S3). Or perhaps it is stored in GitHub for posterity but the actively used copy (for both Jenkins and local builds) is in S3 for ease of download.

rianfowler:
>is there ever a need to do a coordinated deploy of both components that would be difficult under this scheme?

- markup references a new js or css files
- markup references new css classnames 
- markup references deprecated css classnames 

Deploying css / js first would require us to continue using the hashed filenames (i.e. we couldn't rely on just the cache-control header). 

Deploying the static content first would probably require a combination of process and toggles (e.g. removing deprecated css in later release)- probably not feasible especially for new static pages with new js apps.

How do we do this now?

cvalarida:
@patrickvinograd Great questions! I'll break them out:

> Is there typically only one HEAD version of the content that the `vets-website` build should use?

Basically, yes. The HEAD version, as you put it, should be the latest content that's been successfully built and validated. This should represent the latest "known good" content.

> Is there a need to identify/build against multiple versions.

Not that I know of, but that may be more a question for @ElijahLynn. (He'll be back next week, I think.) The CMS has three environments that content passes through, but—and this confuses me too—new content _starts_ in production and makes its way _down_ to staging and dev. So using the latest content from the production CMS environment is typically what we want to do. I can't think of a use case for `vets-website` using the other two.

> What about for deployment - is there ever a need to do a coordinated deploy of both components that would be difficult under this scheme?

I can't think of why that would ever be necessary, but paranoia tells me yes, probably. We do have control over what pages are published from within Drupal, so if we needed to coordinate a content deploy with a specific code deploy, that's at least one switch we can flip. I don't know exactly how much control we have over that, though. Like, if we needed to deploy changes to an existing page alongside a code deploy, that might be tricky.

That said, I don't know how we'd manage it even if we weren't using a single HEAD... :thinking: Unless we just didn't publish any new content until the code is ready to be deployed. That might work.

cvalarida:
@rianfowler The only assets that the static HTML files should be referencing is `static-pages.css`, which is never cache-busted, I think, and `static-pages.js`, which we'll always have to update in the `vets-website` build anyhow.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,237 @@
...
+Accessibility tests run the aXe checker on all pages found in the
+sitemap. Whenever the accessibility check fails, the Jenkins build
+fails. This happens after the content build and link checker.
+
+## Design
+I propose rearchitecting the `vets-website` build to separate the
+static content building into a process distinct from the website.
+
+![separate build diagram](images/separated-content-build-proposal.png)
+
```
wyattwalter: One of the end states here (after validation) is "throw an error stating why the content is invalid". Do we have a mechanism to get this back to a writer that may have added a broken link? Could we possibly do this as part of a save action in Drupal?

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,237 @@
...
+- Who owns the content build repo?
+  - The CMS team?
+  - VSP?
+  - Global UX
+    - Given their relationship with the content, this may make the most sense
+- Who owns managing the content build Jenkins pipeline?
+  - The same team that owns the repo?
+- What triggers a content build?
+  - Could be a daily job
+  - Could be triggered manually from inside Jenkins or even Drupal
```
wyattwalter: Drupal can already trigger builds on Jenkins for updating content, I'd say that's probably the only primary way we'd need to support.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,237 @@
...
+  
+### Serving React apps
+`vets-website` should be responsible for making the pages where the
+React apps live, but not responsible for building the header, footer,
+and other surrounding markup. To make this possible, **I propose we
+have the content build produce an empty React app landing page
+template which the `vets-website` build script can use to make landing
+pages for each React application.** The `vets-website` build script
+would then update the landing page with references to the correct
+JavaScript bundle and CSS files to bootstrap the application.
```
wyattwalter: A question that's not necessarily related to this RFC specifically, but behavior may change with this method inadvertently: what do we do in the event that there's a conflict between content and react pages? i.e. someone creates a `/health/` page in Drupal, but there's already a `/health/` page in vets-website (or vice versa)

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,237 @@
...
+
+The CMS team can also trigger a content-only deploy from inside
+Drupal. If this is done before the latest content build is finished,
+the deploy may not have all the latest changes.
+
+## Risks
+- The current integration is _so_ complex; it would be easy to
+  overlook something while implementing the changes
+
+## Questions
```
wyattwalter: There's some questions in the diagram that aren't represented here, I think.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
cvalarida: What we're doing now is sending a link to the broken build to Slack. The next step is sending the CSV file listing the broken links (which I already have saved in a separate file; the slack plugin just doesn't support uploading files, and I don't even know where to start investigating that one).

Ideally, this validation would be done in Drupal as part of the save action, but according to the CMS team, that's not possible. (cc @ElijahLynn @swirtSJW for confirmation.)

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,237 @@
...
+- The preview server will live in the new content-building repo
+- Sitemap accessibility tests will be done during the content build,
+  not the `vets-website` build
+- The `vets-website` Jenkins build will have the following changes
+  - There will be no cache fallback
+  - There will be no content caching
+  - There will be no sitemap accessibility testing
+  
+### Serving React apps
+`vets-website` should be responsible for making the pages where the
```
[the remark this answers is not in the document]
cvalarida: Ha! Nope. You didn't miss that at all. That's a great point which I hadn't considered yet. :sweat_smile:

If we had access to `vets-website`'s `manifest.json` files, we could either construct a list of exceptions or create temporary files at the appropriate paths right before doing the link validation. Either way introduces some awkward overhead to get to access the `manifest.json` files. :thinking: I'm not sure yet.

Thank you for pointing this out!

platform/engineering/request-for-change/2019-09-30-cms-integration.md
cvalarida: :thinking: Are you saying that when content is updated in Drupal, it can trigger a Jenkins build?

I think if we go that route, that might quickly overwhelm Jenkins as we get _tons_ (hundreds?) of content editors making changes throughout the day.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
cvalarida: Right, so currently, this isn't a thing we're handling at all. Drupal pages are set to override both React landing pages and `vagov-content` pages. We'd probably want to throw up a warning to some place visible (Slack?) and use the React landing page. It would then be on a case-by-case basis what the resolution looks like.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
swirtSJW: Drupal can't fully handle the cases of link validation upon a save in the CMS because when creating multiple new pages, only one can be created at a time.   So it is likely that links may be added that point to a page that does not exist yet.  Requiring a link to be checked before a save could be completed would result in a very stunted workflow for the editors.

cvalarida: Ah! Right. That's what it was. Thank you for that context, @swirtSJW. :raised_hands:

I don't know enough about Drupal to know if there's an alternative there, so I'll defer to @swirtSJW here and say link checking from within Drupal isn't a viable solution for now.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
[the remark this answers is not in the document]
cvalarida: Alright, so I took some time to hammer out some possible solutions. I don't know which is the _best_ way, but I think I'm leaning toward using the manifest files in the content build since I prefer to keep the dependencies explicit.

## Problem statement

Currently, these landing pages seem to all be in vagov-content, but as we move away from that, we need to have another solution. We have a Metalsmith plugin which looks at all the manifest.json files for the paths to the React apps and builds a page there for them. When we separate the builds, the content build will be in charge of validating broken links, but won't have access to the React app landing page locations because those are generated from within vets-website.

## Possible solutions

### Keeping the landing pages in Drupal

****Pros****

-   The link checker works with no modification

****Cons****

-   vets-website React apps depend on the CMS having their landing pages
    -   This dependency is implicit
    -   This would also mean that to develop locally, we'd either have to pull down content or still generate temporary landing pages for the applications if none is found...and if the manifest file can't reliably point to where the application lives, how can we tell?
-   The vets-website build would have to update all the landing pages to point to the appropriate assets
    -   How will it know where the landing pages are?
        -   The manifest.json files could point to the wrong landing page if it isn't in charge of where that page actually ends up

### Using the manifest files in the content build

There are at least a couple ways we could validate that the links to React apps aren't broken if we had access to the manifest files.

1.  Add plugins before and after the broken link checker to create and remove dummy files where the React apps would go
2.  Modify the broken link checker to ignore links that are found in an exceptions list that we'd populate from paths found in the manifest files

****Pros****

-   vets-website would be in charge of where the React apps go

****Cons****

-   The content build would be somewhat more complicated
-   The content build would depend on vets-website

### Use blank pages in the content build, but override in the vets-website build

This would look something like the following:

1.  Insert essentially blank pages during the content build
    -   This could either be adding blank pages in Drupal, or as part of the build script itself
2.  The content build validates links to React apps against the blank pages
3.  The vets-website build overrides the blank landing pages
    -   And issues a warning if that page didn't already exist to make sure it gets added into the content build for the content validation
4.  The vets-website production build searches for blank pages
    -   If any are found, alert somehow before deleting

****Pros****

-   The content build wouldn't be strongly tied to vets-website
-   The vets-website build wouldn't be strongly tied to the content

****Cons****

-   The content build would have kind of an implicit dependency on vets-website
    -   It needs to know where the apps are
-   If pages in the content pointed to the blank landing page, but the path was changed from the manifest.json in vets-website, the links will be broken after the blank landing page is removed
    -   Broken links :(
-   The vets-website build would have kind of an implicit dependency on the content build
    -   It needs to know that the dummy landing pages are blank

cvalarida:
I took like an hour to whip up a webpack-only build proof of concept: https://github.com/department-of-veterans-affairs/vets-website/pull/10898.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
jbalboni: I believe the current process is that you can manually trigger a content build from Drupal. It's not automatic after every change.

jbalboni:
@cvalarida I don't have much to add. I don't see any hidden gotchas. We've talked before, the reason we didn't go down a similar route before is because I didn't want the markup separated from styles/js, which would force us to coordinate changes across multiple repos. But now that seems like a better tradeoff than the build process instability.

The preview server will definitely need some tweaks, but I believe they're minimal since it already fetches the Webpack manifest to figure out how to do the asset hash injection. I think you'll have duplicate code between the content repo and vets-website for that, which I don't think is a big deal.

I think it's worth considering the impact to the speed that content can get from Drupal into production. You'll now have two builds to trigger, which will inevitably be slower. This shouldn't matter for the daily deploy, but for emergency/time-sensitive content that needs to go out immediately it could be an issue.

ncksllvn:
No strong opinions from me. I'm glad we're talking about all this. It seems like this ultimately moves us towards a single source of truth for the content area of the site, so that React applications can deploy and integrate without worrying about rebuilding the whole website. I think this is the right direction, but here are my first thoughts.

- We're still rebuilding the whole website on either a content deployment or app deployment, so there will probably be a performance hit for deployments. We know we have requirement up ahead for time-sensitive deployments - like, less than a minute to go from "clicked this button in Drupal" to being live on the site - does this further that goal at all?
- Are there opportunities to do any health-check on the deployments before they go live? It seems like we'd still be in the same place safety-wise, and will probably see an additional risk with the two chunks - apps and content - converging at some point to populate the S3 bucket.

I'm still semi-interested in exploring the really extreme direction that we've talked about a few times, where vets-website operates conceptually as a live-reload cache layer for Drupal  - as in, any CMS author at any given time can publish a change to Page X, and now only Page X is rebuilt and redeployed - while app-code continues to need a release in order to be redeployed. That's more than likely just wishful thinking, but I'm just brainstorming before we embark on a major technical effort.

platform/engineering/request-for-change/2019-09-30-cms-integration.md
```diff
@@ -0,0 +1,237 @@
...
+- The `vets-website` build will be faster both locally and in Jenkins
+  - Because the build script wouldn't perform any transformations on
+    the content, this would save about half the build time for
+    developers on their local machines (~20s)
+	- Assuming the cache and the `localhost` buildtype are used
+  - All the assets from Drupal (PDFs, JPEGs, etc.) will be bundled up
+    in the single content tarball, so instead of making 475 individual
+    requests (as of the time of this writing), there will be only a
+    single request
+	- Saving upwards of 40 - 50 seconds
```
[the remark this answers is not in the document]
cvalarida: Oh, interesting. I have no idea how long it'll actually take. I just created a tarball of my localhost build (with the JS + CSS) and it ended up at 395 MB, so if downloading it from S3 _is_ slow, then also caching it on the Jenkins server seems like a great way to optimize the build time...if we can swing it. I don't know how that works if the content build and `vets-website` build happen on two different Jenkins swarm nodes, or if there's anything we can do there.

We do still want to upload it to S3 so it's accessible for engineers to download to their local machine, though.

cvalarida:
@jbalboni Yeah, separating the markup from the styles isn't ideal, I'll admit.

As far as the content deployment speed goes, I'm not anticipating much by way of increased latency there, since the content-only deploy would essentially just pull the latest `vets-website` assets (like it does now) and grab the latest content build (instead of building it out), and then deploy it. There's probably some more overhead I'm not accounting for, but I think that'll be on the order of seconds. :crossed_fingers:

cvalarida:
@ncksllvn 

> We're still rebuilding the whole website on either a content deployment or app deployment, so there will probably be a performance hit for deployments. We know we have requirement up ahead for time-sensitive deployments - like, less than a minute to go from "clicked this button in Drupal" to being live on the site - does this further that goal at all?

Not exactly. Nor does it put us further from that goal either, I think. It's then just a matter of _where_ that optimization takes place so we can push out content in under a minute (or whatever time frame we land on). Like I mentioned in my comment to @jbalboni, it'll probably add some seconds to a content-only deploy since there are more moving parts, but I don't anticipate the added latency is going to be too much.

So for a content-only deploy, rather than:
1) Fetch latest webpack assets
2) Fetch latest content
3) Build HTML files
4) Deploy

It would be a two-phase operation:
**Content build**
1) Fetch latest content
2) Build HTML files
3) Upload to S3

**`vets-website` build**
1) Fetch latest webpack assets
2) Fetch latest content
3) Deploy

Again, there's probably some additional overhead I'm not accounting for right now, but this is for the speed of content-only deploys. To get content out the door in under a minute, I think it'll take further architectural changes and we may or may not even use a content-only deploy for that kind of urgent content.

So it doesn't exactly further that goal, but nor does it hinder it, I don't think.

> Are there opportunities to do any health-check on the deployments before they go live? It seems like we'd still be in the same place safety-wise, and will probably see an additional risk with the two chunks - apps and content - converging at some point to populate the S3 bucket.

:thinking: I'm not sure I understand what you're saying here. Content validation would happen in the content build. Unit and e2e tests would happen in the `vets-website` build. Once these two pieces have been validated, we'd also validate that they're integrated properly by checking that the React apps are all in place and pointing to the right bundles.

Is that the kind of health-check you're talking about? I think I'm missing something. :confused:

> ...vets-website operates conceptually as a live-reload cache layer for Drupal...

That's a really interesting idea. This RFC falls somewhat shy of that as written, but I don't think if we do the work to separate the builds, we wouldn't be able to go that route later. I'd recommend holding off on that for now, though, since we're liable to get ourselves in trouble if we get too many pieces moving at once, I think.

So I understand what you're poking at, though, what would that kind of live-reload cache layer look like?
- Would we do any content validation?
  - If so, where?
- Do we have any need to know exactly what's on the site at any given time?
  - I'm inclined to say yes, but I couldn't tell you why
- Would we ever have to revert the content?
  - If so, how?
  - Would it require content editors to revert their changes?
  - Would the reversion be validated (for broken links, mostly)?

PR #2300, opened 2019-10-08
Scripts for repo move for deprecation and url replace

accbjt:
This is 2 scripts created for post repo file move. After the files have been moved we need a script to deprecate old files from the old repo and another script that will replace the old repo url links with the new ones in the new repo file.


accbjt:
@johnpaulashenfelter Can you do a code review and also run both scripts to make sure they are working as expected. After running the scripts it will generate a log.csv that will tell you what happened with each record change. It's great for figuring out errors and what got changed.

johnpaulashenfelter:
@accbjt Should all of those markdown changes be in here? Little surprised by that.

scripts/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "scripts",
+  "version": "1.0.0",
+  "description": "",
+  "main": "repo-move.js",
```
johnpaulashenfelter: Should this file be somewhere?

PR #2325, opened 2019-10-09
rename files based on lowercase naming conventions

LindseySaari:
Renames files in the /platform/triage directory according to new naming conventions defined here:
https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/working-with-vsp/onboarding/repo-guidelines.md#guidelines-for-creating-and-naming-folders-and-files


jbritt1 (approved):
Within guidelines, so approved. Just a quick question for my own personal knowledge though, is "with" a necessary word here or could it be reduced to `rules-of-engagement-triage.md`?

PR #2356, opened 2019-10-09
Url update for vets.gov-teams migration

accbjt:
This is the url replacement generated by the script we created. It should replace the old link url to the new one located in the va.gov-team repo.


thatmattjohnson:
reviewed -- looks good.

thatmattjohnson:
Looks like only 23 items of 40 submitted are here - can you confirm that list is complete? Thanks!

thatmattjohnson (changes requested):
need to update URLs for 17 additional items submitted in original .csv (these may be associated with uploaded documents but should provide redirects regardless)

thatmattjohnson (approved):
approved per convo w @accbjt 10/11/19

PR #2368, opened 2019-10-10
Update code_review_guidelines.md

rianfowler:
- add frontend specific guidelines to PR review guidelines document

### DIFF
--- a/platform/engineering/code_review_guidelines.md
+++ b/platform/engineering/code_review_guidelines.md
@@ -9,7 +9,7 @@ Before sending your code to a VSP team for review, your code should meet all of 
 
 *Hint*: smaller, more frequent pull requests will make this easier for everyone!
 
-## Guidelines
+## General guidelines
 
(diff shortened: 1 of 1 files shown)


deweydell (commented):
@rianfowler These guidelines make sense to me. Some questions I have are:

- You mention when not to block approvals, but are there any examples of when we should we leave reviews that `Request changes` vs `Comment`?
- For first review, do we only need 1 approval to mark it ready for review by the VSP team and does the same time rule for reviewing apply (within 24 hours)?

gunsch (changes requested):
-1 for being part of this doc --- this is a completely different audience! This doc is currently aimed at VFS teams coming to us, for them to understand our process as code authors.

The other doc (the Google Doc linked in the eng-leads channel) is the VSP-team-facing one, which is where we should be putting advice for reviewers.

edit: the content itself is fine! but let's move over there and continue discussion on the doc?

PR #2412, opened 2019-10-11
Update README.md

kfrz:
I don't think it makes sense to list our ToC as "key documents," so I removed those and kept only external ones, which may be outdated anyhow.

I updated the mission to be a slight bit more terse, happy to listen to opposing viewpoints here.


teams/vsp/teams/triage/README.md
```diff
@@ -1,45 +1,37 @@
 # Triage
 
+Repository for team information, planning docs, meeting notes, etc.
+
 ### Mission
-Identify, resolve, and delegate issues across the platform and applications for Veterans and teams building products to serve Veterans while continuously improving telemetry, tools, and documentation.
+Identify, resolve, and delegate issues across the platform and applications for Veterans and teams building products to serve Veterans while
```
LindseySaari: Did you mean to delete "continuously improving telemetry, tools, and documentation." from the sentence here?

kfrz: Nope. Fixed.


I *do* think we need to revisit this mission and clarify the language. It's not very clear to understand.

PR #2490, opened 2019-10-15
Adds security review process doc.

gunsch:
Basically just a writeup of https://dsva.slack.com/archives/CBU0KDSB1/p1567188507041800?thread_ts=1567188275.041500&cid=CBU0KDSB1, but with a checked-in process that we can link to and iterate on / improve from here.

### DIFF
--- a/platform/security/security-review-process.md
+++ b/platform/security/security-review-process.md
@@ -1,0 +1,25 @@
+# Security review process
+
+## Context
+
+Before launching an application on VA.gov, the feature should be reviewed by a VSP security engineer.
+
+This isn't intended to be a thorough audit, but a high-level overview of your application, with a focus on any sensitive areas.
+
+**Note: this is extremely low-process today and will be more fleshed out in the future.**
+
+## Process
+
+1. File a [GitHub issue in `va.gov-team`](https://github.com/department-of-veterans-affairs/va.gov-team/issues/new):
+    * Provide the title "Security Review for [feature name]"
+    * Tag with the label `security-review`
+    * Provide the following in the ticket:
+        * an architecture diagram, showing involved systems + data flows
+        * description of any new publicly-exposed endpoints (`vets-api` or otherwise)
+        * description of any new interactions with dependent services (i.e. `vets-api` --> ???)
+        * description of any other security hotspots you're concerned about / want extra attention on
+1. Schedule an hour "Security Review" meeting with Andrew Gunsch (`@gunsch`), including:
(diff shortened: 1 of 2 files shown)


gunsch:
two-day review ping, can you take a look?

platform/security/security-review-process.md
```diff
@@ -0,0 +1,25 @@
+# Security review process
+
+## Context
+
+Before launching an application on VA.gov, the feature should be reviewed by a VSP security engineer.
```
annaswims: Are we clear who we'd call a "security engineer" ? I feel like it would be an exaggeration to call myself one.

platform/security/security-review-process.md
```diff
@@ -0,0 +1,25 @@
...
+This isn't intended to be a thorough audit, but a high-level overview of your application, with a focus on any sensitive areas.
+
+**Note: this is extremely low-process today and will be more fleshed out in the future.**
+
+## Process
+
+1. File a [GitHub issue in `va.gov-team`](https://github.com/department-of-veterans-affairs/va.gov-team/issues/new):
+    * Provide the title "Security Review for [feature name]"
+    * Tag with the label `security-review`
+    * Provide the following in the ticket:
```
annaswims: Should we be clear that we intend for this documentation to live in the `va.gov-team` not the `va.gov-team` issue.

platform/security/security-review-process.md
```diff
@@ -0,0 +1,25 @@
...
+        * an architecture diagram, showing involved systems + data flows
+        * description of any new publicly-exposed endpoints (`vets-api` or otherwise)
+        * description of any new interactions with dependent services (i.e. `vets-api` --> ???)
+        * description of any other security hotspots you're concerned about / want extra attention on
```
annaswims: Should we put the concerns in `va.gov-team-sensitive`?  We probably don't want to announce our concerns of vulnerabilities.

annaswims (approved):

platform/security/security-review-process.md
gunsch: right now I consider that list to be [gunsch], but I think we're hoping to eventually hire engineers with security background

platform/security/security-review-process.md
gunsch: ooh good question! I'd be supportive of that change.

@andreahewitt-odd do you think it's OK process-wise to move security review issues into `va.gov-team-sensitive`?

platform/security/security-review-process.md
gunsch: I like that idea, reworded

andreahewitt-odd (approved):

PR #2610, opened 2019-10-21
Add triage call center automation RFC

LindseySaari:
# Purpose 

This RFC describes design solutions for automation surrounding the triage call center issue reporting process. The goal is to add further automation as this process is currently very lengthy. Risks and pain points are described in the RFC.


kfrz:
This looks fine but I believe we're putting this on the backburner. @alexpappasoddball should we close this PR? We can always re-open it in the future.

alexpappasoddball:
@kfrz @LindseySaari Lets close this one for now and re-open at a later date if needed.

PR #2646, opened 2019-10-22
Add PII guidelines for VSP teams

johnpaulashenfelter:
This, along with updated code review (completed) and new security guidelines will
be part of our new VSA team onboarding


annaswims:
I think it would help to have an explicit decision on which id's are PII.  Clearly SSN is, but is ICN or EDIPI?

annaswims (approved):
LGTM

gunsch:
> I think it would help to have an explicit decision on which id's are PII. Clearly SSN is, but is ICN or EDIPI?

image

PR #2966, opened 2019-10-30
Migrate files from communications-products

johnpaulashenfelter:


thatmattjohnson:
reviewed and discussed on standup today, looks good to go!

PR #3136, opened 2019-11-06
Inital stab at service deploy rfc

nathanhruby:
several days of wrangling broken builds have led me here.


wunderhund:
I like the basic ideas here, but we're going to need to develop these ideas in a lot more detail to create a plan for moving forward. Do you intend to flesh out those details here or did you want to merge this as-is and add detail later?

nathanhruby:
Hey Craig, I think there's lots to do here still so would like to keep this as a PR and work on it some more.  I think generally this is a high level "Meta RFC" that would set direction and would be followed up by more tactical documents (which would be outlined here before merge).

Goal for right now is to raise awareness and start the conversation because this is a big project.

nathanhruby:
Here is a rough sketch of what I think this layout woudl sort of look like

wyattwalter:
Notes from call:

- restate problems in terms of what VFS teams encounter or how this affects us
- user personas may be a good tool here
- eye towards something more like Heroku for VA teams
- Patrick: get a researcher to help build some questions, and do as many inteviews as we can

nathanhruby:
https://srvaroa.github.io/paas/infrastructure/platform/kubernetes/cloud/2020/01/02/talk-how-to-build-a-paas-for-1500-engineers.html is related and relavant

nathanhruby:
really not sure what happened to this PR but the info was valid and still being worked on.  It's a shame the branch is effectively dead.

I'm coming up with a larger set of recommendations from the Infrastructure Discovery Sprint we just did that was triggered by this RFC.  In the meantime, talk to me or TJ about this plan.

PR #3348, opened 2019-11-12
Migrate teamsite files

johnpaulashenfelter:


thatmattjohnson:
I've updated the migration plan with the inclusion of a /teamsite subdirectory to avoid the conflict you pointed out. This PR will need to be updated accordingly.

johnpaulashenfelter:
Replaced by #3491

PR #3495, opened 2019-11-14
Migrate declare-dependent documentation from vets.gov-team

johnpaulashenfelter:


thatmattjohnson (changes requested):
scripts/migrate.sh - should this file be included in the PR? haven't seen it on other batches.
Everything else looks good.

johnpaulashenfelter:
@thatmattjohnson fixed a (CLI output -- not functional) problem that only noticed when running w/ 101 other files. won't hurt anything unless we rollback this commit so pretty safe. 

Good pickup though :)

PR #3638, opened 2019-11-20
Re-create deleted research for links and buttons

1Copenut:
I documented previously with one of our awesome researchers and it came up again recently. Found it, and am porting to `va.gov-team` repo for posterity.


Clark-W-Griswold (approved):
👍

PR #3652, opened 2019-11-21
create ruby/rails norms document for current versioning

LindseySaari:
This contains our official VSP policy on currenct/up-to-dateness of our Ruby and Rails versions.


alexpappasoddball (approved):
👍

PR #3698, opened 2019-11-25
Move health taxonomy research

batemapf:
[@andrewlewandowski to provide details for the move]


andrewlewandowski (approved):
Looks good! We made this change because we have now fully fleshed out facilities products organized inside the facilities product folder.

PR #3705, opened 2019-11-25
Update current-ruby-rails-norms.md

kfrz:


alexpappasoddball (approved):
LGTM

johnpaulashenfelter (approved):
Thank you for reducing maintenance @kfrz by pointing the reader to where we already keep the latest info on versions (the Gemfile)

gunsch:
can we get some more explicit definition on what "as close to latest" means?

does it mean as soon as a new stable release is out, we're immediately prioritizing an upgrade the next sprint? are we fine sitting one version behind for a bit? (weeks? months?)

LindseySaari (approved):
➕ 1 on Andrew's comment but otherwise, Looks good! 😸

annaswims:
I'd prefer a bit of a defined lag, like 3-6 months for major releases. Being on the bleeding edge has a cost.

gunsch:
Other thoughts for this:
* differences in upgrade policy between major versions (Rails 5->6) vs minor (5.x --> 5.y) vs security releases. should we prioritize security updates much quicker?

johnpaulashenfelter:
Who's the reader for this? Seems like it would be part of the decision record or whatever (we should have had) for rails in general.

But that aside -- the only people doing an upgrade are us (VSP) and we'll prioritize it however it gets prioritized. 🤷‍♂️ If our goal is to be able to point to this to prod devops to update to newer Ruby, feels like its not that useful since they didn't agree. If its to point to our own project manager we're broken. If its to the teams working on the platform then its really just a kind heads-up.

CVEs block our build by default for security issues at that level. There's an implicit rule we already meet.

I think its fine if the underlying goal is "here's what we agree to as a team".

This also doesn't mention anything about other dependencies (Postgres, clamav, poppler, pdftk, etc etc) that the app is contingent on.

Bluntly, I'd like this to have an aspirational feel. 

"The vets-api app attempts to track the latest Rails major version within 1 month of the final release." 

There's no reason preventing that kind of work from being prioritized and the longer we wait, the worse it is, The halo effect from that aspiration is it means we have to update our existing code/gems to keep up which shows the real cost of ongoing maintenance and inevitably will be cheaper than waiting until its an emergency.

General thoughts -- not 👍 or 👎

gunsch:
the requester here is @batemapf : Patrick, can you speak to @johnpaulashenfelter 's question about who the audience for this is?

@johnpaulashenfelter "decision record" framing is interesting, do we have other examples of that?

random examples for pondering: https://github.com/joelparkerhenderson/architecture_decision_record

batemapf:
@johnpaulashenfelter a possible story:

> As a VFS developer architecting a `vets-api` integration with a new upstream service, I need to know what version of the application runtime (Ruby) and framework (Rails) I should expect to be building for so I can confidently make technical decisions that are version-dependent.

In addition, since the API is shared territory b/w us and the Lighthouse/VA API team, some documentation of norms/expectations is hypothetically useful should disagreement arise over a potential upgrade. Also valuable for (as I believe you mentioned) holding ourselves accountable so we don't fall too far behind.

PR #3718, opened 2019-11-25
Add engineering ROE

johnpaulashenfelter:
Starting a draft ROE for engineering borrowing from the recently approved design version


johnpaulashenfelter:
From slack/@gunsch

```
took a look! couple things:
security review at beginning isn't quite what we're doing currently, I'm open to having that be part of the process but we're also asking about implementation details closer to launch, so don't want to frame it just as a "beginning" thing currently.
linking to security/privacy review docs would be good
ctrl+f "design" :wink: 
I like the doc overall!```

gunsch:
the "ctrl+f "design"" comment still stands, there's a "What VSP design does" heading that should be changed to engineering

PR #3813, opened 2019-11-27
Migrate product-support-engineering-jpa documentation from vets.gov-team

billfienberg:
image

The number of changed `.md` files in this PR (45) is one less than the number of`.md` files in the pivot table (46) because there was one file conflicting file. Per the plan, https://github.com/department-of-veterans-affairs/vets.gov-team/blob/master/Practice Areas/Engineering/README.md was supposed to live in the `platform/engineering/backend` folder, but there is already an existing https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/backend/README.md file.


thatmattjohnson (approved):
approved

PR #4704, opened 2020-01-09
Create 2020-01-08-ssoe-removing-implicit-mvi-queries

patrickvinograd:


ericbuckley:
@patrickvinograd the `.md` extension is missing on the file

patrickvinograd:
> @patrickvinograd the `.md` extension is missing on the file

🤦‍♂ I was wondering why I wasn't getting a formatted preview

annaswims:
We're still in the early stages of discovery, but we're exploring what it will be like for a user to be a delegate (a user to act on behalf of a veteran). From what I understand so far, a user with permission to act on behalf of another would log in through SSOe.  Va.gov would get the standard attributes for the currently logged in user  from ssoe, but also an attribute like `is_delegate = true`. From there vets-api would query MVI to get the list of veterans that the user has a delegate relationship with and additional attributes about those veterans.

annaswims:
We've tried it insulate ourselves from MVI failures.  As stands now, when MVI is down a veteran can still log in to va.gov.  Admittedly, it's a diminished login experience if MVI is non-responsive but better than an error. When there is an MVI outage, I'm not aware of a way to contact MVI and get a prompt response.

It looks like MVI outages have become less frequent, but we've also stopped doing high-priority pages when MVI is nonresponsive
https://dsva.pagerduty.com/services/PCIPVGJ?status=acknowledged%2Ctriggered%2Cresolved

platform/engineering/request-for-change/2020-01-08-ssoe-removing-implicit-mvi-queries.md
```diff
@@ -0,0 +1,106 @@
...
+
+It's unclear how much of a problem this poses. We _could_ mitigate this by retaining the ability to do an MVI lookup on every sign-in (e.g. via a configuration setting) and enabling that capability in dev. The MVI lookup would point at the mocked MVI service. 
+
+### Attribute Priority
+There are several user traits that are currently available in both the incoming ID.me SAML assertion and in MVI (`first, middle, last, gender, birth_date, zip, ssn`). vets-api currently prefers the values from ID.me, and only falls back to the MVI values if the ID.me values are not present. 
+
+With an SSOe-based SAML payload we don't have this conflict; we'll only get one value for each of these traits. Presumably this will be the MVI value. On paper this is more "correct" as MVI is the canonical source for VA identity. But we know there are cases where this data is less accurate. VA.gov loses the ability to make this trade-off and prefer one source of another; it devolves to the IAM team to address (and ideally resolve) data mismatches and fidelity issues.
```
gunsch: reading this part scares me a bit --- it seems like we're likely have a one-time cutover then where a bunch of the data might be wrong, since we're dropping ID.me as a source-of-truth of this data? what can we do to mitigate this ahead of time?

platform/engineering/request-for-change/2020-01-08-ssoe-removing-implicit-mvi-queries.md
```diff
@@ -0,0 +1,106 @@
+# RFC: Removing implicit MVI lookups
+
+- Date: _2020-01-08_
+
+
+## Background
+At present, when a user authenticates and is determined to be Level of assurance 3, we look up that user's VA profile in MVI (aka MPI, used interchangeably hereafter). The combination of the incoming attributes from the identity provider and the attributes from MVI make up the bulk of the logical user model that is used across vets-api.
+
+As part of the overall SSO consolidation work, we are integrating VA SSOe with the VA.gov sign-in process. We will maintain the three current credential options, but the SAML request/response cycle will be intercepted by SSOe before being returned to VA.gov. Along the way, SSOe can establish a session cookie for use by other sites. But more importantly, SSOe will provide a standardized SAML payload that includes attributes from the credential provider AND from MVI.
```
gunsch: question from someone who's not close to the rest of the SSOe work: how do different services (vets-api, ...[other services using SSOe]) all get access to the SAML payload, if it's just identified by a session cookie? 

another way to ask: where is that SAML payload cached/stored? it sounds like it's used by a broader set of services than just vets-api

patrickvinograd: SSOe has all the content for the SAML payload stored in association with the user's SSOe session. So another system would make their own SAML request to SSOe and SSOe would send them a SAML response. Because a session already exists (i.e. SSOe can detect their cookie and use the token inside to look up the user's session), SSOe does not need to trigger user authentication - so there's no user interaction other than seeing their browser redirect a couple times.

platform/engineering/request-for-change/2020-01-08-ssoe-removing-implicit-mvi-queries.md
patrickvinograd: I can ask if IAM does anything today when they see an incoming identity assertion from ID.me and there's any data discrepancy. 
FWIW I know vets-api flags some discrepancies but it's mostly been in the aid of debugging, we don't have any mechanism in place to follow up/correct data in MVI if there is a discrepancy.

patrickvinograd:
@annaswims  you're correct that this is introducing a harder dependency on another system during sign-in. If we're distinguishing SSOe from MVI, then, I confirmed that if SSOe (aka the part of SSOe that does the SAML stuff) can't reach MVI,  it returns a "csp-only" response (pretty much equivalent to the limited LOA1 response). So we could sign people in at LOA1 and give them partial access.

So this seems about at par with current state relative to MVI outages; that doesn't address SSOe outages.

annaswims:
Not everybody who logs into Va.gov [has a record in mvi](http://grafana.vfs.va.gov/d/000000062/mvi?orgId=1&from=1578383907558&to=1578483686254).  ~0.5% do not.  Now, if vets-api makes a call to MVI and the user/veteran is not in MVI they will make a call to [dmdc](https://github.com/department-of-veterans-affairs/vets.gov-team/blob/ead0fe129bfee6b8e7eaae59d7f4f681a37dec15/Products/Identity/Login/MVI%20Integration/201808_MVI%20Fallback/180803%20SR-001118%20MVI%20DMDC-Aware%20Search%20for%20Vets.pptx).

Today was a pretty [good example ](http://grafana.vfs.va.gov/d/000000062/mvi?orgId=1&from=now-2d&to=now) MVI had a significant failure rate because DMDC was having database issues.  
Screen Shot 2020-01-13 at 5 34 35 PM
(In this graph MVI lookups spiked as the DMDC database issues were fixed and veterans could log in again with DS logon)

platform/engineering/request-for-change/2020-01-08-ssoe-removing-implicit-mvi-queries.md
```diff
@@ -0,0 +1,106 @@
...
+The isDelegate flag is present in SSOe SAML assertions, so the login flow an account for delegation. The next step of selecting a delegatee on behalf of whom to act, will likely require one or more additional explicit MVI requests which likewise should be separated  from the Redis cache_aside mechanism.
+
+5. **OAuth Flow - Lighthouse**
+SAML authentication and MVI resolution is also performed in the context of the Lighthouse OAuth flow. MVI is queried by an API endpoint triggered by the SAML proxy, and cached MVI data is used when serving OAuth-protected API requests.<br/>
+At this point, there is no intent to integrate SSOe with the OAuth SAML interaction. Signing in to authorize an OAuth application will not establish an IAM session.<br/>
+Current proposal for the OAuth code path is to leave the MVI caching mechanism in place. This will mean that the `User` model definitely needs to remain robust to `va_profile` being populated either from an SSOe SAML payload or from an MVI lookup for the indefinite future. We should audit all of the OAuth-protected APIs to reconcile uses of various delegated methods on `User` to make sure there are no conflicts.
```
dvogel: You've noted below that vets-api prefers ID.me values for some fields. Ignoring those, thinking just about the fields that have a "correct" value like `icn` and `dslogon_edipi`, is there any reason to expect `va_profile` to contain different values?

dvogel (commented):
For Lighthouse's needs:

1. What you've laid out here seems workable. The wiggle in `va_profile` values shouldn't affect too many users as long as the ICN values remain stable.
2. Over the next couple months, we could probably refactor away from the `User` / `UserIdentity` models. It felt appropriate when we implemented it because it was a way to be consistent. I think we can refactor toward a more direct lookup though.
3. Our longer term plan is to move the MVI lookups out of vets-api for a variety of reasons, including latency and exporting the interface to API providers.

platform/engineering/request-for-change/2020-01-08-ssoe-removing-implicit-mvi-queries.md
```diff
@@ -0,0 +1,106 @@
...
+
+- Date: _2020-01-08_
+
+
+## Background
+At present, when a user authenticates and is determined to be Level of assurance 3, we look up that user's VA profile in MVI (aka MPI, used interchangeably hereafter). The combination of the incoming attributes from the identity provider and the attributes from MVI make up the bulk of the logical user model that is used across vets-api.
+
+As part of the overall SSO consolidation work, we are integrating VA SSOe with the VA.gov sign-in process. We will maintain the three current credential options, but the SAML request/response cycle will be intercepted by SSOe before being returned to VA.gov. Along the way, SSOe can establish a session cookie for use by other sites. But more importantly, SSOe will provide a standardized SAML payload that includes attributes from the credential provider AND from MVI. 
+
+Since we will be receiving the bulk of the MVI attributes in the SAML payload, we want to determine whether we can and should do away with the separate MVI query that normally happens immediately following authentication.
```
saneshark: This is certainly going to be a lot of work in redoing specs across the board, might be an opportunity to clean up some things with how we instantiate current_user in our request specs currently. Totally makes sense. Just thinking out load.

platform/engineering/request-for-change/2020-01-08-ssoe-removing-implicit-mvi-queries.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Motivation
+Given that we will receive a SAML payload with most of the needed information from MVI already, doing a separate query to MVI is largely redundant, and adds latency to the sign-in process (though note we're likely swapping added latency during the SAML round-trip for the MVI query latency).
```
saneshark: receiving a SAML payload with *most* of the needed information might not be enough. Would we still keep the ability to query MVI when the payload is missing some needed set of attributes?

platform/engineering/request-for-change/2020-01-08-ssoe-removing-implicit-mvi-queries.md
```diff
@@ -0,0 +1,106 @@
...
+
+
+
+## Design
+In general, MVI attributes are accessed via the `User` model, either as direct delegated methods on `User` (e.g. `user.icn`), or as nested delegated methods on `User.va_profile` (e.g. `user.profile.address). Both of these eventually delegate to the `mvi.profile` attribute on `user`, which is an instance of `MviProfile`.  
+
+We can instead instantiate a MviProfile object by parsing the incoming SAML attributes and delegate to that directly from `User.va_profile`. Initially we'll want to do so selectively for SSOe authentication so we can support old and new authentication mechanisms during roll-out.
```
saneshark: Is this possibly an opportunity to have all of the MVI profile information just live directly on the user model or I guess more accurately user_identity model? I think this would simplify things quite a bit.

platform/engineering/request-for-change/2020-01-08-ssoe-removing-implicit-mvi-queries.md
patrickvinograd: ICN might come along with an MHV credential - but in that case we look up from MVI using that ICN as the primary key.  Likewise EDIPI might come along with a DSLogon credential, but in that case we use the EDIPI as the lookup key. So I don't see any case where those identifiers would not match what comes back in va_profile.

platform/engineering/request-for-change/2020-01-08-ssoe-removing-implicit-mvi-queries.md
patrickvinograd: That's a good point in terms of LOE to get this done.

patrickvinograd:
@annaswims regarding the lack of presence in MVI - that's a good point. I will ask the IAM team if the MVI lookup that occurs internal to SSOe has the pass-through to DEERS enabled. And inquire if they've done anything to insulate their system from DMDC instability.

patrickvinograd:
More generally, SSOe is capable of returning a "csponly" SAML payload for users who are definitively not in MVI.

platform/engineering/request-for-change/2020-01-08-ssoe-removing-implicit-mvi-queries.md
patrickvinograd: Yes, logically if it comes in via a SAML assertion it will just be attributes on user_identity, and then I would just change the  delegation of `user.va_profile` things to point to `identity`.

platform/engineering/request-for-change/2020-01-08-ssoe-removing-implicit-mvi-queries.md
patrickvinograd: Yes, this is described further below, there will be cases for explicit queries.

patrickvinograd:
Some repo history rewriting caused this to get closed in an un-mergeable state. I merged it in a separate PR.

PR #4760, opened 2020-01-10
Additions to VSA Slack Guidance

ncksllvn:
Adds general Slack usage suggestions as well as engineering-specific


sshein (changes requested):
I feel like we shouldn't include #va-vsa-engineering for now since it doesn't exist and this doc is going to be shared with the team on Monday all hands.

ncksllvn:
@sshein removed!

sshein (approved):

PR #4869, opened 2020-01-15
Answered some of the questions

steve-gov:


jason-gcio (approved):
These look like great additions.

sporeboy (approved):
Looks good. Some things may be worth a few minutes of future discussion.

PR #5102, opened 2020-01-23
Add list of resources about design docs

billfienberg:
Add a `resources.md` file to the `platform/engineering/design-docs` folder

### DIFF
--- a/platform/engineering/design-docs/resources.md
+++ b/platform/engineering/design-docs/resources.md
@@ -1,0 +1,12 @@
+# Design Doc Resources
+
+## Blogs
+
+- https://www.freecodecamp.org/news/how-to-write-a-good-software-design-document-66fcf019569c/
+- https://medium.com/dev-channel/design-docs-a-design-doc-a152f4484c6b
+
+## Examples
+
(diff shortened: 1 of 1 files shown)


gunsch (commented):
I'd rather not --- this loses fidelity of information from what's in the "Engineering Design Doc Culture" doc (i.e. don't just put links! put article titles / design doc titles. a reader staring at a list of "docs.google.com" URLs isn't going to find them particularly useful.). Plus also loses most of the context from that doc.

if anything, I'd rather see us port that entire doc over to GitHub (or most of it) once it's more finalized, since then we also get the philosophy / intention of how we want to use it / etc. but for now, let's table this?

billfienberg:
> I'd rather not --- this loses fidelity of information from what's in the "Engineering Design Doc Culture" doc (i.e. don't just put links! put article titles / design doc titles. a reader staring at a list of "docs.google.com" URLs isn't going to find them particularly useful.). Plus also loses most of the context from that doc.
> 
> if anything, I'd rather see us port that entire doc over to GitHub (or most of it) once it's more finalized, since then we also get the philosophy / intention of how we want to use it / etc. but for now, let's table this?

Sounds good to me. Thanks for the feedback.

PR #5248, opened 2020-01-29
Design Doc/RFC - Multi-repo Documentation Search

billfienberg:
**Status**: On Hold while we seek approval for Algolia

In December 2019, members of the Product Support, Content & IA, and Leadership team conducted a [discovery sprint about documentation](https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/products/platform/documentation-site/research/discovery-sprint-12-2019). 

Here is the [product outline](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/platform/documentation-site/product-outline.md) and the [recommendation](https://docs.google.com/presentation/d/14eD-GJ5mBQXh1L9k5Lr83zbVV1iehp0a67qZz0Fn_94/edit#slide=id.p1) that came out of the discovery sprint.

This PR follows our [Design Doc template](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/design-docs/design-doc-template.md) to seek feedback and approval for a solution to search multiple, public documentation sources with a single query from a single entry point.

### DIFF
--- a/docs/docsearch-scraper-config.json
+++ b/docs/docsearch-scraper-config.json
@@ -5,25 +5,19 @@
       "url": "https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/",
       "selectors_key": "folder",
       "page_rank": 0,
-      "tags": [
-        "folder"
-      ]
+      "tags": ["folder"]
     },
     {
       "url": "https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/.+$",
       "selectors_key": "folder",
       "page_rank": 0,
-      "tags": [
-        "folder"
-      ]
+      "tags": ["folder"]
     },
     {
       "url": "https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/.+\\.md$",
       "selectors_key": "file",
       "page_rank": 5,
-      "tags": [
-        "file"
-      ]
+      "tags": ["file"]
     }
   ],
   "selectors": {
@@ -40,14 +34,9 @@
       "text": "article p, article li"
     }
   },
-  "selectors_exclude": [
-    ".up-tree"
-  ],
+  "selectors_exclude": [".up-tree"],
   "custom_settings": {
-    "attributesForFaceting": [
-      "language",
-      "version"
-    ]
+    "attributesForFaceting": ["language", "version"]
   },
   "stop_urls": [
     "https://raw.githubusercontent.com",
--- a/docs/docsearch-scraper-config.json
+++ b/docs/docsearch-scraper-config.json
@@ -1,0 +1,63 @@
+{
+  "index_name": "dev_docs",
+  "start_urls": [
+    {
+      "url": "https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/",
+      "selectors_key": "folder",
+      "page_rank": 0,
+      "tags": [
+        "folder"
+      ]
+    },
+    {
+      "url": "https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/.+$",
+      "selectors_key": "folder",
+      "page_rank": 0,
+      "tags": [
+        "folder"
+      ]
+    },
+    {
+      "url": "https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/.+\\.md$",
+      "selectors_key": "file",
+      "page_rank": 5,
+      "tags": [
+        "file"
+      ]
+    }
+  ],
+  "selectors": {
+    "folder": {
+      "text": ".js-navigation-item td.content"
+    },
+    "file": {
+      "lvl0": "article h1",
+      "lvl1": "article h2",
+      "lvl2": "article h3",
+      "lvl3": "article h4",
+      "lvl4": "article h5",
+      "lvl5": "article h6",
+      "text": "article p, article li"
+    }
+  },
+  "selectors_exclude": [
+    ".up-tree"
+  ],
+  "custom_settings": {
+    "attributesForFaceting": [
+      "language",
+      "version"
+    ]
+  },
+  "stop_urls": [
+    "https://raw.githubusercontent.com",
+    "https://github.com/department-of-veterans-affairs/va.gov-team/commits/*",
+    "https://github.com/department-of-veterans-affairs/va.gov-team/blame",
+    "https://github.com/department-of-veterans-affairs/va.gov-team/pulls",
+    "https://github.com/department-of-veterans-affairs/va.gov-team/pull/*",
+    "https://github.com/department-of-veterans-affairs/va.gov-team/issues/*",
+    "https://github.com/department-of-veterans-affairs/va.gov-team/branches",
+    "https://github.com/department-of-veterans-affairs/va.gov-team/graphs",
+    "https://github.com/department-of-veterans-affairs/va.gov-team/find"
+  ]
+}
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -72,6 +72,16 @@ Subject to review, we will run the crawler in a scheduled job once every 24 hour
 
 It's possible that running the scraper once per day would consume a significant number of operations.
 
+#### Public repos
+
+We will crawl and scrape the following public documentation sources:
+
+- [`va.gov-team` repo](https://github.com/department-of-veterans-affairs/va.gov-team)
+- [`veteran-facing-services-tools` repo](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/)
+- [`vets-contrib` repo](https://github.com/department-of-veterans-affairs/vets-contrib/)
+- [design.va.gov](https://design.va.gov/)
+- [developer.va.gov](https://developer.va.gov/)
+
 #### Private repos
 
 ---
@@ -200,11 +210,11 @@ The recommendation from the discovery sprint is to build a custom documentation 
 
 ### Revision History
 
-| Date         | Revisions Made                                                              | Author        | Reviewed By |
-| ------------ | --------------------------------------------------------------------------- | ------------- | ----------- |
-| Jan 27, 2020 | Initial Draft                                                               | Bill Fienberg |             |
-| Feb 10, 2020 | Update Detailed Design                                                      | Bill Fienberg |             |
-| Feb 11, 2020 | Answer open questions                                                       | Bill Fienberg |             |
-| Feb 12, 2020 | Expand on where/when to run crawler                                         | Bill Fienberg |             |
-| Feb 13, 2020 | Add Usage Limits section, and update Code Location & Work Estimate sections | Bill Fienberg |             |
-| Feb 17, 2020 | Add open question about crawler diffing capabilities                        | Bill Fienberg |             |
+| Date         | Revisions Made                                                                | Author        | Reviewed By |
+| ------------ | ----------------------------------------------------------------------------- | ------------- | ----------- |
+| Jan 27, 2020 | Initial Draft                                                                 | Bill Fienberg |             |
+| Feb 10, 2020 | Update Detailed Design                                                        | Bill Fienberg |             |
+| Feb 11, 2020 | Answer open questions                                                         | Bill Fienberg |             |
+| Feb 12, 2020 | Expand on where/when to run crawler                                           | Bill Fienberg |             |
+| Feb 13, 2020 | Add Usage Limits section, and update Code Location & Work Estimate sections   | Bill Fienberg |             |
+| Feb 17, 2020 | Add list of public docs, and open question about crawler diffing capabilities | Bill Fienberg |             |
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -26,7 +26,7 @@ We are adding a public landing page that will contain a search input where users
 
 ### Detailed Design
 
-For a multi-repo search MVP, we are adding an HTML text input and [Algolia's DocSearch JS snippet](https://github.com/algolia/docsearch) to the [`va.gov-team` GitHub Pages site](https://department-of-veterans-affairs.github.io/va.gov-team/), [Algolia's Search API](https://www.algolia.com/products/search/), and [Algolia's scraper](https://github.com/algolia/docsearch-scraper).
+For a multi-repo search MVP, we are adding an HTML text input and [Algolia's DocSearch JS snippet](https://github.com/algolia/docsearch) to the [`va.gov-team` GitHub Pages site](https://department-of-veterans-affairs.github.io/va.gov-team/). This will use [Algolia's Search API](https://www.algolia.com/products/search/) to query Algolia's Hosted database (link needed), which will be populated by [Algolia's scraper](https://github.com/algolia/docsearch-scraper) using a configuration we supply (link needed).
 
 In general, this kind of system requires the following components:
 
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -84,7 +84,7 @@ The DocSearch JS snippet doesn't seem to use any debouncing. I noted the number 
 
 Since each crawl wipes and repopulates the index, that means each crawl will produce a number of operations approximately equal to the previous number of records. The number of operations may be larger due to new lines/files in the repos, and it may be smaller due to deleted lines/files in the repo.
 
-If the `va.gov-team` repo currently produces ~150K records, then the crawler `va.gov-team` would produce ~150K operations every time it runs. If the `va.gov-team` repo is crawler once per day, then that would produce ~150K operations every day. At that rate, we'd have approximately ~4.5M operations every month from crawling the `va.gov-team` repo alone.
+If the `va.gov-team` repo currently produces ~150K records, then crawling `va.gov-team` would produce ~150K operations every time it runs. If the `va.gov-team` repo is crawler once per day, then that would produce ~150K operations every day. At that rate, we'd have approximately ~4.5M operations every month from crawling the `va.gov-team` repo alone.
 
 ##### Recommended Plan for Our Usage
 
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -50,15 +50,15 @@ By leveraging the following existing technologies, we should only need to add th
 
 ##### The Community Plan
 
-The free tier, the Community plan, includes up to 10K records and 50K operations (read/write/edit/delete). After hitting the limit, we'd have to upgrade to a paid plan.
+The free tier, the Community plan, includes up to 10K records and 50K operations (read/write/edit/delete). If we exceed the limits of the free tier, we'd have to upgrade to a paid tier.
 
 ##### The Starter Plan
 
-The middle tier, the Starter plan, includes up to 50K records and 250k operations. It costs an extra $10/month to get 100K additional operations, and it costs an extra $10/month to get an extra 20K records.
+The middle tier, the Starter plan, includes up to 50K records and 250k operations. If we exceed the limits of the Starter plan, it would cost an extra $10/month for every additional 100K operations, and an extra $10/month for every additional 20K records.
 
 ##### The Pro Plan
 
-And the upper tier, the Pro plan, includes up to 1M records and 5M operations. It costs an extra $5/month to get 100K additional operations, and it costs an extra $5/month to get an extra 20K records.
+The upper tier, the Pro plan, includes up to 1M records and 5M operations. If we exceed the limits of the Pro plan, it would cost an extra $5/month foe every 100K additional operations, and an extra $5/month for every additional 20K records.
 
 ---
 
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -6,7 +6,7 @@
 
 **Status:** **Draft**
 
-**Approvers:** Andrew G., Megan K., Rian F., DevOps
+**Approvers:** Andrew G., Megan K., Rian F., VSP Ops
 
 ## Overview
 
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -84,6 +84,7 @@ While there shouldn't be any PII in our public documentation sources, it is poss
 - What is the Not-Invented-Here risk? How likely is that we're reinventing the wheel?
 - What is the inexperience risk? Have any of the team members involved ever built/deployed/maintained this type of system before?
 - It's possible that work estimates are materially inaccurate.
+- What is the documentation debt risk? At least one person interviewed during the discovery sprint said that we need a solution ASAP. Is there a risk that people are deferring documentation in the present until the custom documentation solution is shipped?
 
 ### Work Estimates
 
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -18,25 +18,17 @@ _The intended audience for this document is software engineers, but it should ma
 
 ### Objective
 
-We want to implement a public landing page where VFS/VSP team members can search multiple documentation source from a single entrypoint.
+We want to implement a public landing page where VFS/VSP team members can search multiple documentation sources from a single entrypoint.
 
 ### Background
 
-_The background section should contain information the reader needs to know to understand the problem being solved. This can be a combination of text and links to other documents._
-
-_Do **NOT** describe the solution here. That goes in High Level Design._
+The documentation we have is spread across many different locations (public repos, private repos, GitHub pages, etc). There currently is not a way for VFS/VSP users to search all of the sources with a single query.
 
 ### High Level Design
 
-_A high-level description of the system. This is the most valuable section of the document and will probably receive the most attention. You should explain, at a high level, how your system will work. Don't get bogged down with details; those belong later in the document._
-
-_A diagram showing how the major components communicate is very useful and a great way to start this section. If this system is intended to be a component in a larger system, a diagram showing how it fits in to the larger system will also be appreciated by your readers._
-
-_Most diagrams will need to be updated over time as the design evolves, so please create your diagrams with a program that is easily (and freely) available and attach the diagram source to the document to make it easy for a future maintainer (who could be you) to update the diagrams along with the document._
+We are adding a public landing page that will contain a form where users can search all of our public documentation in one location.
 
 ## Specifics
-
-_Nothing goes here; all the content belongs in the subsections._
 
 ### Detailed Design
 
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -162,16 +162,13 @@ While there shouldn't be any PII in our public documentation sources, it is poss
 
 ### Work Estimates
 
-1. Provision and configure data store to house records that will be returned from the Search API. - 1 week
-
-- I estimate one week because I suspect we'll either need to get approval to use Algolia's hosted API or get approval to provision a custom data store.
-
-1. Configure crawler to seed the initial data store. - 2 days
-1. Run the crawler to populate the data store. - 2 days
-1. Build a landing page that will contain a search input that submits requests to the Search API. - 2 days
-1. Deploy the landing page to a publicly available location. - 2 days
-
-- I estimate this may take longer if we need to need to coordinate with other teams.
+1. Convert `va.gov-team`'s `docs/index.md` from markdown to HTML - <1 hour
+1. Add a search input to `va.gov-team`'s `docs/index.html` - <5 minutes
+1. Add the DocSearch `<style>` and `<script>` tags to `va.gov-team`'s `docs/index.html` - <5 minutes
+1. Write the config file for the scraper - <1 day
+1. Add CircleCI to `va.gov-team` - <1 day
+1. Add a scheduled job to run the scraper once every 24 hours in `va.gov-team`'s CircleCI pipeline - <1 day
+1. Run the scraper - <1 day
 
 ### Alternatives
 
@@ -199,10 +196,10 @@ The recommendation from the discovery sprint is to build a custom documentation 
 
 ### Revision History
 
-| Date         | Revisions Made                                            | Author        | Reviewed By |
-| ------------ | --------------------------------------------------------- | ------------- | ----------- |
-| Jan 27, 2020 | Initial Draft                                             | Bill Fienberg |             |
-| Feb 10, 2020 | Update Detailed Design                                    | Bill Fienberg |             |
-| Feb 11, 2020 | Answer open questions                                     | Bill Fienberg |             |
-| Feb 12, 2020 | Expand on where/when to run crawler                       | Bill Fienberg |             |
-| Feb 13, 2020 | Add Usage Limits section and update Code Location section | Bill Fienberg |             |
+| Date         | Revisions Made                                                              | Author        | Reviewed By |
+| ------------ | --------------------------------------------------------------------------- | ------------- | ----------- |
+| Jan 27, 2020 | Initial Draft                                                               | Bill Fienberg |             |
+| Feb 10, 2020 | Update Detailed Design                                                      | Bill Fienberg |             |
+| Feb 11, 2020 | Answer open questions                                                       | Bill Fienberg |             |
+| Feb 12, 2020 | Expand on where/when to run crawler                                         | Bill Fienberg |             |
+| Feb 13, 2020 | Add Usage Limits section, and update Code Location & Work Estimate sections | Bill Fienberg |             |
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -95,8 +95,6 @@ _Features you'd like to (or will need to) add but aren't required for the curren
 
 ### Revision History
 
-_The table below should record the major changes to this document. You don't need to add an entry for typo fixes, other small changes or changes before finishing the initial draft._
-
-| Date         | Revisions Made                                                                                                                                   | Author         | Reviewed By |
-| ------------ | ------------------------------------------------------------------------------------------------------------------------------------------------ | -------------- | ----------- |
-| Jan 19, 2016 | Initial Draft based off [Arvados's template](https://dev.arvados.org/projects/arvados/wiki/Design_Doc_Template) which is reminiscent of Google's | Albert J. Wong |
+| Date         | Revisions Made | Author        | Reviewed By |
+| ------------ | -------------- | ------------- | ----------- |
+| Jan 27, 2020 | Initial Draft  | Bill Fienberg |             |
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -136,7 +136,11 @@ We could search for the same term in both the Algolia UI and the landing page to
 
 ### Security Concerns
 
-If the search input dispatched a request directly to the Algolia API, then there'd be minimal risk of the documentation search negatively affecting other systems. If the request was proxied through our API, then we should be able to rely on any existing DOS protections.
+The [DocSearch JS snippet](https://docsearch.algolia.com/docs/dropdown) requires four pieces of data; the `appId` (public), search `apiKey` (public), `indexName`, and `inputSelector`.
+
+The `appId` and search `apiKey` are both public. The search `apiKey` only allows read access. There is a separate admin API key that the crawler will use, and that admin API key will need to be configured in an environment variable in CircleCI.
+
+We don't currently have any custom private attributes on the records in our index. If we did, we could configure [`unretrievableAttributes`](https://www.algolia.com/doc/api-reference/api-parameters/unretrievableAttributes/), which is a list of attributes that cannot be retrieved at query time.
 
 ### Privacy Concerns
 
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -6,29 +6,37 @@
 
 **Status:** **Draft**
 
-**Approvers:** Andrew G., Megan K.
+**Approvers:** Andrew G., Megan K., Rian F., DevOps
 
 ## Overview
 
 ### Objective
 
-We want to implement a landing page where VFS/VSP team members can search multiple documentation sources from a single entrypoint.
+Search multiple documentation sources from a single, public entrypoint.
 
 ### Background
 
-The documentation we have is spread across many different locations (public repos, private repos, GitHub pages, etc). There currently is not a way for VFS/VSP users to search all of the documentation sources with a single query.
+The documentation we have is spread across many different locations (public repos, private repos, GitHub pages, handbooks, etc). There currently is not a way for VFS/VSP users to search all of the documentation sources with a single query.
 
 ### High Level Design
 
-We are adding a landing page that will contain a form where users can search multiple documentation sources in one location.
+We are adding a public landing page that will contain a form where users can search multiple, public documentation sources with a single query.
 
 ## Specifics
 
 ### Detailed Design
 
-_Designs that are too detailed for the above High Level Design section belong here. Anything that will require a day or more of work to implement should be described here._
+We are adding a public landing page (on a custom domain, if possible) that will contain a search input. That search input will query multiple, public documentation sources with one query.
 
-_This is a great place to put APIs, communication protocols, file formats, and the like._
+A successful implementation will have these components.
+
+1. Page with Input Textbox - We need a public page with a search input where users can input their search term(s).
+1. Index/Database - We need a place to store the records that will be returned to the user.
+1. Crawler/Scraper - We need a crawler to frequently scan our documentation sources for additions/deletions/modifications.
+
+We do not currently have an index of our target documentation sources. And we do not currently have infrastructure to crawl our target documentation sources to build said index.
+
+Algolia appears to be the leader when it comes to searching documentation.
 
 _It is important to include assumptions about what external systems will provide. For example if this system has a method that takes a user id as input, will your implementation assume that the user id is valid? Or if a method has a string parameter, does it assume that the parameter has been sanitized against injection attacks? Having such assumptions explicitly spelled out here before you start implementing increases the chances that misunderstandings will be caught by a reviewer before they lead to bugs or vulnerabilities. Please reference the external system's documentation to justify your assumption whenever possible (and if such documentation doesn't exist, ask the external system's author to document the behavior or at least confirm it in an email)._
 
@@ -42,6 +50,10 @@ _The path of the source code in the repository._
 
 _How you will verify the behavior of your system. Once the system is written, this section should be updated to reflect the current state of testing and future aspirations._
 
+#### Testing the search input
+
+#### Testing the crawler
+
 ### Logging
 
 By logging the term(s) that users are querying, we could theoretically identify trends to help increase the quality of our search results.
@@ -52,15 +64,19 @@ _How users can debug interactions with your system. When designing a system it's
 
 ### Caveats
 
-- Access Control: By making the documentation landing page public, it becomes difficult to impossible to allow an unauthenticated user to search documentation from private sources (like a private Github repo). One suggestion is that search results lead to public documentation sources, and we include links to private documentation in the public sources. That way, we can rely on existing access control solutions, while still providing a decent solution. One of the major downsides of that approach is auditing, adding, updating, and maintaining links to/from the public and private sources.
+- Access Control: By making the documentation landing page public, it becomes difficult, if not impossible, to allow an unauthenticated user to search documentation from private sources (like a private Github repo). One suggestion is that search results lead to public documentation sources, and we include links to private documentation in the public sources. That way, we can rely on existing access control solutions, while still providing a decent solution. One of the major downsides of that approach is auditing, adding, updating, and maintaining links to/from the public and private sources.
 
 ### Security Concerns
 
-_This section should describe possible threats (denial of service, malicious requests, etc) and what, if anything, is being done to protect against them. Be sure to list concerns for which you don't have a solution or you believe don't need a solution. Security concerns that we don't need to worry about also belong here (e.g. we don't need to worry about denial of service attacks for this system because it only receives requests from the api server which already has DOS attack protections)._
+The search input would need to be sanitized.
+
+If the search input dispatched a request directly to the Algolia API, then there'd be minimal risk of the documentation search negatively affecting other systems. If the request was proxied through our API, then we should be able to rely on any existing DOS protections.
 
 ### Privacy Concerns
 
-_This section should describe any risks related to user data, PII that are added by this new application. Think about flows of user data through systems, places data is stored and logged, places data is displayed to users. Where is user data stored or logged? How long is it stored?_
+Since the initial implementation will be an unauthenticated landing page, we will have little, if any, data about the user.
+
+While there shouldn't be any PII in our public documentation sources, it is possible it exists, and hasn't been discovered yet. Therefore, it is possible that our search results could lead public users to sources that contain PII.
 
 ### Open Questions and Risks
 
@@ -93,7 +109,7 @@ _Split the work into milestones that can be delivered, put them in the order tha
 
 ### Future Work
 
-_Features you'd like to (or will need to) add but aren't required for the current release. This is a great place to speculate on potential features and performance improvements._
+- Ability to search private documentation sources
 
 ### Revision History
 
--- a/products/platform/documentation-site/research/search-design-doc.md
+++ b/products/platform/documentation-site/research/search-design-doc.md
@@ -2,7 +2,7 @@
 
 **Author(s):** Bill Fienberg <bill.fienberg@oddball.io>
 
-**Last Updated:** 2020-02-10
+**Last Updated:** 2020-02-11
 
 **Status:** **Draft**
 
@@ -26,7 +26,7 @@ We are adding a public landing page that will contain a search input where users
 
 ### Detailed Design
 
-For a multi-repo search MVP, we are adding an HTML text input and [Algolia's DocSearch JS snippet](https://github.com/algolia/docsearch) to the [`va.gov-team` GitHub Pages site](https://department-of-veterans-affairs.github.io/va.gov-team/), [Algolia's Search API](https://www.algolia.com/products/search/), and [Algolia's scraper](https://github.com/algolia/docsearch-scraper). 
+For a multi-repo search MVP, we are adding an HTML text input and [Algolia's DocSearch JS snippet](https://github.com/algolia/docsearch) to the [`va.gov-team` GitHub Pages site](https://department-of-veterans-affairs.github.io/va.gov-team/), [Algolia's Search API](https://www.algolia.com/products/search/), and [Algolia's scraper](https://github.com/algolia/docsearch-scraper).
 
 In general, this kind of system requires the following components:
 
@@ -36,13 +36,21 @@ In general, this kind of system requires the following components:
 - Crawler/Scraper - We need a crawler to scan our documentation sources and push data to our index.
 - Pipeline - We need a way to routinely run the crawler
 
-Each one of the above components could be a separate buy-or-build decision. 
+Each one of the above components could be a separate buy-or-build decision.
 
-By leveraging the following existing technologies, we should only need to add the search input to the landing page, configure the scraper, and configure when to run the scraper. 
+By leveraging the following existing technologies, we should only need to add the search input to the landing page, configure the scraper, and configure when to run the scraper.
 
-- Algolia's Search API product will satisfy the index and API components. 
-- Algolia's open source JavaScript snippet will display search results to the user. 
-- Algolia's open source scraper will crawl and scrape our public documentation sources to push records into our index. 
+- Algolia's Search API product will satisfy the index and API components.
+- Algolia's open source JavaScript snippet will display search results to the user.
+- Algolia's open source scraper will crawl and scrape our public documentation sources to push records into our index.
+
+#### Where/when to run the scraper
+
+For the MVP, we will run the crawler in a scheduled job once every 24 hours in Circle CI.
+
+#### Private repos
+
+---
 
 ### Code Location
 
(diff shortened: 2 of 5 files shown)


products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,114 @@
...
+
+### Open Questions and Risks
+
+- Should we index private documentation sources? If so, what private documentation sources can we expose to public users, if any?
+- Should the landing page be public only, or should it also support authenticated users?
+- How often should we reindex our documentation sources? For perspective, Algolia's open source crawler runs every 24 hours.
+- What should the URL of the documentation landing page be? Since we have `design.va.gov`, does something like `docs.va.gov` make sense?
+- What is the current opportunity cost of building a custom solution? What are we choosing to delay right now by building a custom solution?
+- What is the future opportunity cost of maintaining a custom solution? What will we have to delay in the future by maintaining a custom solution?
+- What is the key person risk? How disruptive would it be if the author(s) of the custom solution were temporarily/permanently unavailable?
```
cvalarida: These are really good questions! Building and maintaining a custom solution makes me nervous for a number of reasons. I don't know how to measure the opportunity costs myself, but my guess is it's not insignificant.

So we can have a point of reference, what do you anticipate the level of effort to be like for building and maintaining a custom solution? Like, what would be involved?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,114 @@
...
+- What is the Not-Invented-Here risk? How likely is that we're reinventing the wheel?
+- What is the inexperience risk? Have any of the team members involved ever built/deployed/maintained this type of system before?
+
+### Work Estimates
+
+_Split the work into milestones that can be delivered, put them in the order that you think they should be done, and estimate roughly how much time you expect it each milestone to take. Ideally each milestone will take one week or less._
```
cvalarida: Related to the above question: What might these look like?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,114 @@
...
+**Status:** **Draft**
+
+**Approvers:** Andrew G., Megan K., Rian F., DevOps
 
 ## Overview
 
-## Detailed Design
+### Objective
+
+Search multiple documentation sources from a single, public entrypoint.
```
cvalarida: What's the objective of this design document? To propose a specific solution (e.g. Algolia) or foster discussion so we can settle on a solution?

cvalarida (commented):
Sooo....I made lots of comments. :sweat_smile: 

The design doc does a good job about outlining the problem space, but after reading it, I'm not sure what the recommendation is. You hint most strongly at using Algolia, but never called it out explicitly.

All things considered, it's a great start! I really, really like the Open Questions and Risks section! :clap:

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,114 @@
...
+
+_How you will verify the behavior of your system. Once the system is written, this section should be updated to reflect the current state of testing and future aspirations._
+
+#### Testing the search input
+
+#### Testing the crawler
+
+### Logging
+
+By logging the term(s) that users are querying, we could theoretically identify trends to help increase the quality of our search results.
```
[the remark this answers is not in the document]
billfienberg: >Most popular searches and results most often returned by the search engine.

https://www.algolia.com/products/analytics/?tab=popular

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,114 @@
...
+
+The search input would need to be sanitized.
+
+If the search input dispatched a request directly to the Algolia API, then there'd be minimal risk of the documentation search negatively affecting other systems. If the request was proxied through our API, then we should be able to rely on any existing DOS protections.
+
+### Privacy Concerns
+
+Since the initial implementation will be an unauthenticated landing page, we will have little, if any, data about the user.
+
+While there shouldn't be any PII in our public documentation sources, it is possible it exists, and hasn't been discovered yet. Therefore, it is possible that our search results could lead public users to sources that contain PII.
```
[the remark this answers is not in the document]
billfienberg: I've added a link to the mitigation steps in https://github.com/department-of-veterans-affairs/va.gov-team/pull/5248/commits/5d669dbd7e63e5bce51a336bc788611dfe0780fd. 

Here are the steps:
https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/removing-sensitive-data-from-a-repository.md#checklist-for-removing-pii-in-md-file-from-a-documentation-repo

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,114 @@
...
+
+### Security Concerns
+
+The search input would need to be sanitized.
+
+If the search input dispatched a request directly to the Algolia API, then there'd be minimal risk of the documentation search negatively affecting other systems. If the request was proxied through our API, then we should be able to rely on any existing DOS protections.
+
+### Privacy Concerns
+
+Since the initial implementation will be an unauthenticated landing page, we will have little, if any, data about the user.
```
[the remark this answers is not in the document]
billfienberg: I suppose not. Line removed in https://github.com/department-of-veterans-affairs/va.gov-team/pull/5248/commits/bd398a59ee0606481984bb2d8f92a0399665561d.

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,114 @@
...
+
+While there shouldn't be any PII in our public documentation sources, it is possible it exists, and hasn't been discovered yet. Therefore, it is possible that our search results could lead public users to sources that contain PII.
+
+### Open Questions and Risks
+
+- Should we index private documentation sources? If so, what private documentation sources can we expose to public users, if any?
+- Should the landing page be public only, or should it also support authenticated users?
+- How often should we reindex our documentation sources? For perspective, Algolia's open source crawler runs every 24 hours.
+- What should the URL of the documentation landing page be? Since we have `design.va.gov`, does something like `docs.va.gov` make sense?
```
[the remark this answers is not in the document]
billfienberg: This site is intended to be very product-team focused. I suspect there would be little, if any, value to a veteran that stumbled across the site we're proposing. 

That being said, I feel like the same argument could be made about `design.va.gov`, and that seemingly wasn't a strong enough argument to prevent `design.va.gov` from going live. 

Additionally, during a conversation with @rianfowler, he pointed out that he would prefer a route (I suspect something like `va.gov/docs`) over a subdomain because subdomains have additional challenges (specifically CORS) that routes don't.

products/platform/documentation-site/research/search-design-doc.md
billfienberg: My gut estimate is that it would take 1-3 sprints to build, deploy, and iterate on to get to a satisfactory state. But I suspect that could easily take much longer if we had to wait for approvals and/or other teams to set up the resources we're requesting.

products/platform/documentation-site/research/search-design-doc.md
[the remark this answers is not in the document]
cvalarida: Yeah, I don't have a strong opinion either way. In any case, as long as we explicitly state what the site is _for_ when you get there, I'm okay with mostly whatever.

For example: https://developer.va.gov
It's a pretty generic subdomain, but once you go there, you can see it's clearly "A Veteran-centered API platform for securely accessing VA data." (Shout out to the awesome lighthouse team working on this! :raised_hands: )

I'm curious if anybody has a name with a bit more baked-in context, though.

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,157 @@
...
+
+### Debugging
+
+The Docker image for the scraper is available on [Docker Hub](https://hub.docker.com/r/algolia/docsearch-scraper). Once we have a config file, we can [run the scraper locally](https://community.algolia.com/docsearch/run-your-own.html#run-the-crawl-from-the-docker-image).
+
+#### Debugging the scraper
+
+#### Debugging the search functionality
+
+We could search for the same term in both the Algolia UI and the landing page to confirm if they both return the same results.
```
cvalarida: What do we do if the search results aren't updated after we create, update, or delete a document (that we expect to affect said results)? What kind of output does the scraper give us? Will we keep the logs somewhere?

products/platform/documentation-site/research/search-design-doc.md
cvalarida: Do you have an idea for what maintenance on this would look like if we went with Algolia vs something else?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,157 @@
...
+- What is the documentation debt risk? At least one person interviewed during the discovery sprint said that we need a solution ASAP. Is there a risk that people are deferring documentation in the present until the custom documentation solution is shipped?
+
+### Work Estimates
+
+1. Provision and configure data store to house records that will be returned from the Search API. - 1 week
+
+- I estimate one week because I suspect we'll either need to get approval to use Algolia's hosted API or get approval to provision a custom data store.
+
+1. Configure crawler to seed the initial data store. - 2 days
+1. Run the crawler to populate the data store. - 2 day
```
cvalarida:
```suggestion
1. Run the crawler to populate the data store. - 2 days
```

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,157 @@
...
+1. Build a landing page that will contain a search input that submits requests to the Search API. - 2 day
```
cvalarida:
```suggestion
1. Build a landing page that will contain a search input that submits requests to the Search API. - 2 days
```

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,157 @@
...
+1. Deploy the landing page to a publicly available location. - 2 day
```
cvalarida:
```suggestion
1. Deploy the landing page to a publicly available location. - 2 days
```

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,157 @@
...
+
+- I estimate this may take longer if we need to need to coordinate with other teams.
+
+### Alternatives
+
+#### Ready-made alternatives
+
+- [Slab](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/platform/documentation-site/research/discovery-sprint-12-2019/technology-discovery.md#slab) supports multi-repo search, but was considered too expensive (about \$24k/year for the current number of users in the #general channel of the DSVA Slack workspace).
```
cvalarida: Considered too expensive by whom? Citation?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,114 @@
...
+1. Index/Database - We need a place to store the records that will be returned to the user.
+1. Crawler/Scraper - We need a crawler to frequently scan our documentation sources for additions/deletions/modifications.
+
+We do not currently have an index of our target documentation sources. And we do not currently have infrastructure to crawl our target documentation sources to build said index.
+
+Algolia appears to be the leader when it comes to searching documentation.
+
+### Code Location
+
+_The path of the source code in the repository._
```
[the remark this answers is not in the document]
meganhkelley: Re: new repo conversations — _does it_ make most sense to put the code in a new repo? What code? If it's the code for the search tool itself (which is the subject of this design doc, so I assume yes), then yes — or rather, in that case I'm inclined to put the code wherever y'all as engineers think is best. 

But we should also keep in mind that this landing page will host content beyond the search functionality (beyond the scope of this design doc) — it will host the entire documentation homepage, including a table of contents, and longer term may we recommend building out "top docs" as part of that website experience as well. The homepage's table of contents, and the contents of those "top docs" should likely be served from `va.gov-team`. 

In which case, should the search tool code live in a new repo, and the homepage itself ultimately pull content from both the search tool repo and the `va.gov-team` repo? Or, some other arrangement?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,114 @@
...
+### Privacy Concerns
+
+Since the initial implementation will be an unauthenticated landing page, we will have little, if any, data about the user.
+
+While there shouldn't be any PII in our public documentation sources, it is possible it exists, and hasn't been discovered yet. Therefore, it is possible that our search results could lead public users to sources that contain PII.
+
+### Open Questions and Risks
+
+- Should we index private documentation sources? If so, what private documentation sources can we expose to public users, if any?
+- Should the landing page be public only, or should it also support authenticated users?
```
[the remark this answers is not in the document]
meganhkelley: Agreed, searching private repos is out of scope for MVP. I think it's fine to leave here (as we discussed @billfienberg, knowing that that _could_ be something we want to implement down the line could have implications for MVP's technical solution), but you could clarify that we've already ruled this out of scope.

products/platform/documentation-site/research/search-design-doc.md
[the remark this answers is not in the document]
meganhkelley: What is the process for getting a `*.va.gov` or `va.gov/*` URL? If that is our recommendation we should include that process here. I am on board with `docs.va.gov` or `va.gov/docs` but am not sold on it being necessary for MVP if the process is extensive.

If we are considering a va.gov URL, we will also need to consult with @mnorthuis as the guardian of va.gov URLs.

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,157 @@
...
+- What should the URL of the documentation landing page be? Since we have `design.va.gov`, does something like `docs.va.gov` make sense?
+- What is the current opportunity cost of building a custom solution? What are we choosing to delay right now by building a custom solution?
+- What is the future opportunity cost of maintaining a custom solution? What will we have to delay in the future by maintaining a custom solution?
+- What is the key person risk? How disruptive would it be if the author(s) of the custom solution were temporarily/permanently unavailable?
+- What is the Not-Invented-Here risk? How likely is that we're reinventing the wheel?
+- What is the inexperience risk? Have any of the team members involved ever built/deployed/maintained this type of system before?
+- It's possible that work estimates are materially inaccurate.
+- What is the documentation debt risk? At least one person interviewed during the discovery sprint said that we need a solution ASAP. Is there a risk that people are deferring documentation in the present until the custom documentation solution is shipped?
```
U-DON: Is there an interim place we can recommend storing documentation? A source we can be confident will be scraped when the doc site gets implemented?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,114 @@
...
+
+Since the initial implementation will be an unauthenticated landing page, we will have little, if any, data about the user.
+
+While there shouldn't be any PII in our public documentation sources, it is possible it exists, and hasn't been discovered yet. Therefore, it is possible that our search results could lead public users to sources that contain PII.
+
+### Open Questions and Risks
+
+- Should we index private documentation sources? If so, what private documentation sources can we expose to public users, if any?
+- Should the landing page be public only, or should it also support authenticated users?
+- How often should we reindex our documentation sources? For perspective, Algolia's open source crawler runs every 24 hours.
```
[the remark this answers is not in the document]
U-DON: Maybe not so pie-in-the-sky if we set GitHub actions to trigger an endpoint on the crawler. 😁

products/platform/documentation-site/research/search-design-doc.md
meganhkelley: It is worth clarifying in the design doc that we are recommending a custom _homepage_ solution — or, specifically for this case, a custom _search_ solution. ("Custom solution" is too vague IMO.)

products/platform/documentation-site/research/search-design-doc.md
meganhkelley: Let's update this to be more reflective of the nuance of our decision — we (as a discovery sprint team — this can be cited to discovery sprint documentation) did not recommend Slab _as an MVP_ for a variety of reasons:
- Price
- User-based access (you could share public links to documentation, but as far as we could tell to navigate the site within Slab would require a Slab account, or an external, public, custom homepage anyway). Relatedly, does not meet users in their existing workflow.
- Not an MVP, requires "all or nothing" implementation
- No guarantee that it would stay accessible on VA network

We don't need to mention all of those in the design doc, but saying that it was exclusively about price is missing the nuance. Also, should we mention that Slab is in fact still on the table as a post-MVP solution? I don't think that's necessary to include but not sure.

Also, I do agree that it's good to bring up off the shelf solutions in this design doc, since they're very relevant to the discussion, but it's worth noting that we're not comparing apples to apples. Slab has a search functionality (Algolia, in fact), but it is also a complete documentation site solution. It's an alternative that would replace custom search, yes, but it's not a direct search-to-search comparison. Does that make sense? If so, that should be clarified in the design doc so that everyone understands what it is.

products/platform/documentation-site/research/search-design-doc.md
meganhkelley: Our interim documentation site solution is the `va.gov-team` repo

products/platform/documentation-site/research/search-design-doc.md
meganhkelley: Also, is key person risk specific to this recommendation? I would think that would be more about our own docs about how to build and maintain any product, and less about whether we're recommending algolia or a different search solution?

products/platform/documentation-site/research/search-design-doc.md
meganhkelley: Another thought on the "documentation debt" point. This question is based on the assumption that there's another solution out there that would be faster. I don't think that's accurate, but I also don't think that's the right question to be asking. If we're worried about timing, we should be saying — do we need to cut features from our MVP in order to ship this sooner? 

The tricky thing with this whole problem area is that the technical solution — a custom homepage — is only half the battle. Our proposed solution for addressing documentation debt (per the discovery sprint recommendation) is more about setting processes and standards, which is an effort that must be completed _regardless_ of the technical solution.

meganhkelley:
Wow, this is great, @billfienberg ! Awesome job pulling this together. And thanks so much for digging in so carefully @cvalarida and others :) 

I've left some comments above, but here are a couple additional notes: 

_High level feedback:_
- Let's be sure to keep this focused on the search functionality that is (a big) part of our docs landing page MVP. Our first step is to launch a lean docs homepage with search and ToC — that will allow us to validate search itself and our approach overall. This design doc shouldn't be trying to solve the entire documentation problem (it's not — you've done a good job of keeping it focused, but it's worth reiterating that). We're just trying to drill down on our technical approach for implementing public, multi-repo search, so that we can start validating our assumptions, gathering data, and iterating. 🎉 

_Things that are missing:_
- @gunsch mentioned that you should account for the process of getting approval to use a new technology — this is the Slack thread he sent that has some context to get you heading in the right direction: 
- There is nothing in here about whether we think we're on Algolia's free plan, or whether we think we need a paid plan, etc. Can you please capture that? And then let's connect @billfienberg about reaching out to Algolia and VSP leadership re: pricing. (Maybe this just needs to go in `Open Questions & Risks`?)

Again, great stuff! Excited to see what other folks think and keep digging in 😎

products/platform/documentation-site/research/search-design-doc.md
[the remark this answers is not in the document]
tskinn: Its definitely technically easier/faster to use a path instead of getting a new subdomain. I personally haven't gone through getting a new subdomain but there is some paperwork and a group within the VA that needs to approve it and then the DNS folks also need to be involved.. We would also need to setup SSL certs and the like.

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,157 @@
...
+## Specifics
+
+### Detailed Design
+
+We are adding a public landing page that will contain the [DocSearch frontend library](https://github.com/algolia/docsearch), and running our own instance of the [`docsearch-scraper`](https://github.com/algolia/docsearch-scraper) to extract content from our documentation sources and push it to an Algolia index.
+
+A successful implementation will have these components.
+
+1. Page with Input Textbox - We need a public page with a search input where users can input their search term(s).
+1. Index/Database - We need a place to store the records that will be returned to the user.
```
tskinn: I thought the Index/Database is hosted by algolia. Am I missing something?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,157 @@
...
+**Status:** **Draft**
+
+**Approvers:** Andrew G., Megan K., Rian F., DevOps
 
 ## Overview
 
-## Detailed Design
+### Objective
+
+Run our own instance of [Algolia's DocSearch](https://community.algolia.com/docsearch/what-is-docsearch.html) on a custom landing page.
```
tskinn: To me, this makes it sound like we are hosting something besides a webpage and from what I can tell this isn't really the case. We can optionally run the scraper ourselves to have more control of scraping but the index/database is hosted by algolia. All we need to do for searching is insert some javascript into the desired webpage and that piece run queries against algolias API. Do I have that right?

products/platform/documentation-site/research/search-design-doc.md
billfienberg: You are correct. If we sign up for Algolia, then Algolia would host the index.

products/platform/documentation-site/research/search-design-doc.md
billfienberg: Algolia allows people to apply to be included in their scraper that runs every 24 hours, but I don't think we meet all the requirements on their [checklist](https://community.algolia.com/docsearch/who-can-apply.html#the-checklist). 

Therefore, I think running the scraper ourselves is required, not optional.

products/platform/documentation-site/research/search-design-doc.md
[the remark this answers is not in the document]
cvalarida: Just from a separation of concerns point-of-view, I think it probably makes sense to have a repo dedicated to the doc search + future features to build on top of that.

products/platform/documentation-site/research/search-design-doc.md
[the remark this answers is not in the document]
cvalarida: Oooh, I like that. Might get weird if we have a source of documentation that _isn't_ GitHub, but I don't think that's the case now. There are also performance considerations, but...implementation details. :smile:

products/platform/documentation-site/research/search-design-doc.md
cvalarida: @meganhkelley Yeah, I was specifically asking about "Algolia vs [insert other fancy technology]" to accomplish the goal, not about the custom docs homepage as a whole. :+1:

What I'm trying to poke at is how much continual tweaking it'll take after it's all set up. If it's really fire-and-forget, awesome! I just tend to assume that's never the case. :sweat_smile:

meganhkelley: +++++ yep, I also assume that no-maintenance-forever is not the case, and this Q is definitely necessary to dig into!!

products/platform/documentation-site/research/search-design-doc.md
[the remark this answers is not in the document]
meganhkelley: We will be wanting to index results from design.va.gov in the MVP, but not sure if that still counts as GitHub (since it’s coming from https://github.com/department-of-veterans-affairs/vets-design-system-documentation and maybe we could use that to trigger the crawler).

products/platform/documentation-site/research/search-design-doc.md
billfienberg: This is a screenshot of the scraper output. 
image

It does seem like it would make sense to keep the logs somewhere.

products/platform/documentation-site/research/search-design-doc.md
billfienberg: It's difficult to speculate how much maintenance it would require, but my gut says that we'd probably want to budget 1-3 months worth of small adjustments, iterations, and fine tuning based on user feedback. I suspect after that, it would rarely need a lot of attention, unless a breaking change occurred (like renaming a repo, migrating a repo, or changes to a CSS selector, etc). 

Phrased another way, for the first few sprints after the MVP ships, perhaps we should budget a small percentage of the sprint capacity (<20%) for maintenance/iteration.

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,114 @@
...
+
+_How users can debug interactions with your system. When designing a system it's important to think about what tools you can provide to make debugging problems easier. Sometimes it's unclear whether the problem is in your system at all, so a mechanism for isolating a particular interaction and examining it to see if your system behaved as expected is very valuable. Once a system is in use, this is a great place to put tips and recipes for debugging. If this section grows too large, the mechanisms can be summarized here and individual tips can be moved to another document._
+
+### Caveats
+
+- Access Control: By making the documentation landing page public, it becomes difficult, if not impossible, to allow an unauthenticated user to search documentation from private sources (like a private Github repo). One suggestion is that search results lead to public documentation sources, and we include links to private documentation in the public sources. That way, we can rely on existing access control solutions, while still providing a decent solution. One of the major downsides of that approach is auditing, adding, updating, and maintaining links to/from the public and private sources.
```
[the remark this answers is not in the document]
rianfowler: I was thinking the same thing.

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,156 @@
...
+- What should the URL of the documentation landing page be? Since we have `design.va.gov`, does something like `docs.va.gov` make sense?
+- What is the current opportunity cost of building a custom solution? What are we choosing to delay right now by building a custom solution?
+- What is the future opportunity cost of maintaining a custom solution? What will we have to delay in the future by maintaining a custom solution?
+- What is the key person risk? How disruptive would it be if the author(s) of the custom solution were temporarily/permanently unavailable?
+- What is the Not-Invented-Here risk? How likely is that we're reinventing the wheel?
+- What is the inexperience risk? Have any of the team members involved ever built/deployed/maintained this type of system before?
+- It's possible that work estimates are materially inaccurate.
+- What is the documentation debt risk? At least one person interviewed during the discovery sprint said that we need a solution ASAP. Is there a risk that people are deferring documentation in the present until the custom documentation solution is shipped?
+
```
rianfowler: - How do we configure what pages get indexed? 
- What is a page? What parts of the page get indexed? 
- How the data on the pages gets transformed into algolia's data structure e.g. how do we configure the titles on the results 
- How are results weighted? How do we control the weights / priorities of results? 

Screen Shot 2020-02-03 at 14 10 46

would like to see a plan that generates the index and demonstrates the results locally- before we commit to building infrastructure to support this. At the very least we should user test this with VSP people to prove that this index is more useful than the one provided by github.

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,156 @@
...
+- Since the documentation landing page will be public only, we will not be able to search private sources of documentation. If searching private documentation becomes a necessary feature in the future, the proposed implementation will likely not be sufficient.
+- How often should we reindex our documentation sources? For perspective, Algolia's open source crawler runs every 24 hours.
+- What should the URL of the documentation landing page be? Since we have `design.va.gov`, does something like `docs.va.gov` make sense?
+- What is the current opportunity cost of building a custom solution? What are we choosing to delay right now by building a custom solution?
+- What is the future opportunity cost of maintaining a custom solution? What will we have to delay in the future by maintaining a custom solution?
+- What is the key person risk? How disruptive would it be if the author(s) of the custom solution were temporarily/permanently unavailable?
+- What is the Not-Invented-Here risk? How likely is that we're reinventing the wheel?
+- What is the inexperience risk? Have any of the team members involved ever built/deployed/maintained this type of system before?
```
rianfowler: I don't know if a search index is something the platform should try to specialize in when there are viable third party options we can use. 

I suspect the bulk of the work is going to be cleaning up the input- indexing poorly structured, poorly edited content may result in a bad user experience. I assume that most of our documentation is in this state. 

I strongly recommend focusing on the part of this with the least amount of certainty (e.g. the indexing, the UX of the results) and to try to find a low effort way to validate or iterate on that. We should avoid committing any infrastructure to this until we have this working the way we expect in a local environment.

products/platform/documentation-site/research/search-design-doc.md
rianfowler: Is there any sort of data munging that has to occur in our set up or do we expect to try to control the index / results solely from the source data?

products/platform/documentation-site/research/search-design-doc.md
billfienberg: That makes sense. Updated in https://github.com/department-of-veterans-affairs/va.gov-team/pull/5248/commits/4207adb63eda693989ab1ce7a45a18f6cf0a57b3

products/platform/documentation-site/research/search-design-doc.md
billfienberg: Here are some links to the crawler config documentation. 
https://community.algolia.com/docsearch/required-configuration.html
https://community.algolia.com/docsearch/config-file.html

To explicitly tell the crawler which routes to start crawling from, we can add the routes to the the [`start_urls` array](https://community.algolia.com/docsearch/config-file.html#start_urls).

To prevent the crawler from crawling a specific route, we can add the route to the [`stop_urls` array](https://community.algolia.com/docsearch/config-file.html#stop_urls-optional). 

To exclude specific parts of a page from the crawler, we can add CSS selectors to the [`selectors_exclude` array](https://community.algolia.com/docsearch/config-file.html#selectors_exclude-optional)

You can read a detailed explanation about how Algolia extracts payloads from selectors [here](https://community.algolia.com/docsearch/how-do-we-build-an-index.html). 

According to https://community.algolia.com/docsearch/required-configuration.html#use-the-right-classes-as-selectors:
> `lvlX` selectors should use the standard title tags like `h1`, `h2`, `h3`, etc. You can also use static class. Set a unique `id` or `name` attribute to these elements as detailed below.
> 
> Every DOM elements matching the `lvlX` selectors must have a unique `id` or `name` attribute. This will help the redirection to directly scroll down to the exact place of the matching elements. These attributes defined the right anchor to use.
> 
> Every textual element (selector text) must be wrapped in a tag `<p>` or `<li>`. This content must be atomic and split into small entities. Be careful to never nest one matching elements into another one, it will create duplicates.

GitHub's markdown seems to properly render the heading and textual elements that the Algolia scraper looks for.
image

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,159 @@
...
-Reviewers: TBD
+**Author(s):** Bill Fienberg <bill.fienberg@oddball.io>
 
-## Objective
+**Last Updated:** 2020-02-04
 
-## Background
+**Status:** **Draft**
+
+**Approvers:** Andrew G., Megan K., Rian F., VSP Ops
```
gunsch: [meta] "VSP Ops" isn't a person, find an individual who you think should approve

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,159 @@
...
 
 ## Overview
 
-## Detailed Design
+### Objective
+
+Run our own instance of [Algolia's DocSearch](https://community.algolia.com/docsearch/what-is-docsearch.html) on a custom landing page.
```
gunsch: "run Algolia" sounds like a proposed solution, not the objective of the design doc?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,159 @@
...
+
+### Open Questions and Risks
+
+- Since the documentation landing page will be public only, we will not be able to search private sources of documentation. If searching private documentation becomes a necessary feature in the future, the proposed implementation will likely not be sufficient.
+- How often should we reindex our documentation sources? For perspective, Algolia's open source crawler runs every 24 hours.
+- What should the URL of the documentation landing page be? Since we have `design.va.gov`, does something like `docs.va.gov` make sense?
+- What is the current opportunity cost of building a custom solution? What are we choosing to delay right now by building a custom solution?
```
gunsch: wait, isn't this design doc about the search MVP? and we're using algolia? what custom solution are we talking about here?

gunsch (commented):
major questions I don't see addressed by this doc?

* what will serve the new homepage?
* where (runtime / environment / frequency / etc.) will the crawler run?
* where (runtime / environment) will the index DB be hosted?

this is a lot closer to building a full custom solution than I was expecting from initial conversations.

products/platform/documentation-site/research/search-design-doc.md
cvalarida: @billfienberg Correct me if I'm wrong, but I think this is here because it'll take a lot of our own infrastructure and wiring to make this particular solution (Algolia) work. The indexing and whatnot isn't the custom bit, but it's not exactly an out-of-the-box solution since [we can't use their scraper](https://github.com/department-of-veterans-affairs/va.gov-team/pull/5248#discussion_r373544321).

products/platform/documentation-site/research/search-design-doc.md
meganhkelley: Which of those requirements do you not think we meet @billfienberg? (cc @cvalarida as I noticed you just linked to this comment)

products/platform/documentation-site/research/search-design-doc.md
billfienberg: @cvalarida, you are correct. We could use Algolia's hosted search engine to avoid maintaining a DB, and maintaining an API to access that DB. We could use Algolia's open source scraper to avoid writing our own scraper. And we could use Algolia's UI library to avoid writing a custom search result renderer. 

But we'd still need to iterate on the configs for the search index, scraper, and UI library, in addition to configuring when to run the scraper (for example, cron job every 24 hours). And I suspect we'd need to review the source and image for the scraper before adding it one of our pipelines. 

TLDR: There are several components to the system, and each one can be an independent buy-or-build decision. 

- Data store
- Way to get data into the data store (scraper)
- Way to get data out of the data store (search API)
- Way to display data to the user

[Algolia's Search API](https://www.algolia.com/products/search/) would satisfy the data store and API components. 

[Algolia's scraper](https://hub.docker.com/r/algolia/docsearch-scraper) would provide a way to get data into the data store. 

And [Algolia's DocSearch UI library](https://github.com/algolia/docsearch) would provide a way to display data to the user.

products/platform/documentation-site/research/search-design-doc.md
billfienberg: Here's the checklist for reference:
https://community.algolia.com/docsearch/who-can-apply.html#the-checklist

I think we fail to satisfy the following requirements:

> You must be the owner of the website, or at least have the power to update its content. You'll have to include a JavaScript snippet to enable DocSearch.

Our docs live in GitHub. We do not own GitHub; Microsoft does. While we have the power to update the contents of our repo, we can't add the JavaScript snippet. 

no-script-tag-in-github-flavored-markdown

> Your website must be a documentation website. We do not index blogs or commercial content.

While we have repos that exist specifically for documentation, I don't know if Algolia would consider them a "documentation website".

gunsch: I believe we can put JS snippets on GH pages though? i.e. on https://department-of-veterans-affairs.github.io/va.gov-team/

does the JS snippet need to be on every page, or just _a_ page?

products/platform/documentation-site/research/search-design-doc.md
billfienberg: I've gone through a couple iterations of the objective. Which one do you think makes the most sense?
- Original: We want to implement a public landing page where VFS/VSP team members can search multiple documentation source from a single entrypoint.
- Generic: Search multiple documentation sources from a single, public entrypoint.
- Specific: Run our own instance of [Algolia's DocSearch](https://community.algolia.com/docsearch/what-is-docsearch.html) on a custom landing page.

Do you consider the Original or Generic option better than the current option (Specific)?

products/platform/documentation-site/research/search-design-doc.md
billfienberg: The JS snippet will need to be included on every page we want to have our search input. So far a landing page, we'd only need the JS snippet on one page. 

https://community.algolia.com/docsearch/dropdown.html

Here's the actual `DocSearch` source: 
https://github.com/algolia/docsearch/blob/master/src/lib/DocSearch.js

products/platform/documentation-site/research/search-design-doc.md
gunsch: I think I'm largely trying to reconcile the stated purpose ("implement Algolia") with the contents (where you talk about needing an indexer and a DB? but those are provided by Algolia, so if the objective is "implement Algolia" then that should be assumed).

Is this doc about "Algolia is already decided on, here's a doc on how we should implement it", or is it "we need a search system, here are some options (and Algolia is a strong contender)"?

products/platform/documentation-site/research/search-design-doc.md
gunsch: Are we evaluating these things independently? i.e. would we consider using a non-Algolia scraper with Algolia's data store? Or Algolia's scraper with our own data store?

products/platform/documentation-site/research/search-design-doc.md
gunsch: @billfienberg I'm confused --- it sounds like you told @meganhkelley  that the checklist requirement we don't meet is "can't add the JS snippet", but if the JS snippet is just about DocSearch and we're putting it on the landing page, then... it sounds like we _do_ meet that checklist requirement?

Can you clarify which checklist requirement we don't meet, and (if it's the JS one) how you're understanding that requirement?

products/platform/documentation-site/research/search-design-doc.md
billfienberg: This doc is about "Algolia is already decided on, here's a doc on how we should implement it."

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,178 @@
...
+
+Once we have records in our Algolia index and the DocSearch frontend library on a page, we should be able to test the search functionality the same way the end user would use it.
+
+### Logging
+
+Algolia includes analytics, such as ["Most popular searches and results most often returned by the search engine."](https://www.algolia.com/products/analytics/?tab=popular)
+
+### Debugging
+
+The Docker image for the scraper is available on [Docker Hub](https://hub.docker.com/r/algolia/docsearch-scraper). Once we have a config file, we can [run the scraper locally](https://community.algolia.com/docsearch/run-your-own.html#run-the-crawl-from-the-docker-image).
```
cvalarida: This will populate the _actual_ database, right? Is there a way to run the scraper without writing to an actual database? Or maybe for this purpose, we have a docker image of the DB that it can write to instead of the live, hosted DB where we have to pay for records stored and operations performed?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,178 @@
...
+
+We could search for the same term in both the Algolia UI and the landing page to confirm if they both return the same results.
+
+### Caveats
+
+- Access Control: By making the documentation landing page public, it becomes difficult, if not impossible, to allow an unauthenticated user to search documentation from private sources (like a private Github repo). One suggestion is that search results lead to public documentation sources, and we include links to private documentation in the public sources. That way, we can rely on existing access control solutions, while still providing a decent solution. One of the major downsides of that approach is auditing, adding, updating, and maintaining links to/from the public and private sources.
+
+### Security Concerns
+
+If the search input dispatched a request directly to the Algolia API, then there'd be minimal risk of the documentation search negatively affecting other systems. If the request was proxied through our API, then we should be able to rely on any existing DOS protections.
```
cvalarida: What's the suggested implementation and how does it affect this?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,178 @@
...
+
+---
+
+### Open Questions and Risks
+
+---
+
+### Work Estimates
+
+1. Provision and configure data store to house records that will be returned from the Search API. - 1 week
```
cvalarida: Is this still applicable?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,178 @@
...
+
+- I estimate one week because I suspect we'll either need to get approval to use Algolia's hosted API or get approval to provision a custom data store.
+
+1. Configure crawler to seed the initial data store. - 2 days
+1. Run the crawler to populate the data store. - 2 days
+1. Build a landing page that will contain a search input that submits requests to the Search API. - 2 days
+1. Deploy the landing page to a publicly available location. - 2 days
+
```
cvalarida: Are these estimates still "accurate?" (That is, do you want to update them given any new information we learned and experiments we performed?)

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,178 @@
...
+- Index/Database - We need a place to store the records that will be returned to the user.
+- Search API - We need a way to get data out of the index.
+- Crawler/Scraper - We need a crawler to scan our documentation sources and push data to our index.
+- Pipeline - We need a way to routinely run the crawler
+
+Each one of the above components could be a separate buy-or-build decision.
+
+By leveraging the following existing technologies, we should only need to add the search input to the landing page, configure the scraper, and configure when to run the scraper.
+
+- Algolia's Search API product will satisfy the index and API components.
```
cvalarida: Are we missing the database portion, or is that what you mean by `index` here? If so, could you call that out more specifically and link to the page on Algolia's site that explains it? (This is the part that we'd have to pay for, right?)

cvalarida (commented):
I think it could be a little more direct in that the approach this is outlining is using Algolia's (paid) hosted option. Also, it might be good to give the records and operations estimates we came up with in here for reference.

products/platform/documentation-site/research/search-design-doc.md
billfienberg: We are not evaluating these things independently. I think there'd be little to no value in using Algolia's scraper or JS library without using their index.

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,207 @@
...
+
+Essentially, every line of text (`<p>`, `<li>`, `<h1-6>`, etc) in a markdown file ends up as a separate record in the Algolia index. We have approximately 3K `.md` files in the `va.gov-team` repo. We ran a test crawler on the `va.gov-team` repo to estimate how many elements would match our selectors, and it found almost 150K matching elements, which equates to almost 150K records.
+
+#### Where/when to run the scraper
+
+Subject to review, we will run the crawler in a scheduled job once every 24 hours in a Circle CI pipeline for the `va.gov-team` repo.
+
+It's possible that running the scraper once per day would consume a significant number of operations.
+
+#### Private repos
```
wyattwalter: I feel like this is the real sticking point for me on any search product. If we index a private repo with a public search function, that information becomes public. Any ideas on how this could work?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,207 @@
...
+
+The documentation we have is spread across many different locations (public repos, private repos, GitHub pages, handbooks, etc). There currently isn't a way for VFS/VSP users to search multiple documentation sources with a single query.
+
+### High Level Design
+
+## Specifics
+
+### Detailed Design
+
+For a multi-repo search MVP, we are adding an HTML text input and [Algolia's DocSearch JS snippet](https://github.com/algolia/docsearch) to the [`va.gov-team` GitHub Pages site](https://department-of-veterans-affairs.github.io/va.gov-team/). This will use [Algolia's Search API](https://www.algolia.com/products/search/) to query Algolia's hosted database (https://www.algolia.com/pricing/), which will be populated by [Algolia's scraper](https://github.com/algolia/docsearch-scraper) using a [configuration we supply](docs/docsearch-scraper-config.json).
```
wyattwalter: What repositories would we end up searching? The MVP lists this repo, but what others are you targeting?

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -1,15 +1,178 @@
...
+
+For the MVP, we will run the crawler in a scheduled job once every 24 hours in Circle CI.
+
+#### Private repos
+
+---
+
+### Code Location
+
+The code will live in a new repo (tentatively titled something like `docs-landing-page`).
```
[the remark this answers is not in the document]
billfienberg: This is no longer true. I updated this PR to include the WIP `config.json` in https://github.com/department-of-veterans-affairs/va.gov-team/pull/5248/commits/9323a758758249a746ebfb8206a65921d56f2340.

products/platform/documentation-site/research/search-design-doc.md
billfienberg: You are correct. The Docker image for the scraper will write to the actual DB. I haven't seen anything in their scraper docs to run the scraper without writing to the DB.

products/platform/documentation-site/research/search-design-doc.md
billfienberg: Since the request is dispatched directly to Algolia, I updated the security concerns, and removed this chunk in https://github.com/department-of-veterans-affairs/va.gov-team/pull/5248/commits/dce864810f76a161676225fca07ca3fdaa64671d.

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -12,32 +12,75 @@
...
+The free tier, the Community plan, includes up to 10K records and 50K operations (read/write/edit/delete). If we exceed the limits of the free tier, we'd have to upgrade to a paid tier.
+
+##### The Starter Plan
+
+The middle tier, the Starter plan, includes up to 50K records and 250k operations. If we exceed the limits of the Starter plan, it would cost an extra $10/month for every additional 100K operations, and an extra $10/month for every additional 20K records.
+
+##### The Pro Plan
 
-A successful implementation will have these components.
+The upper tier, the Pro plan, includes up to 1M records and 5M operations. If we exceed the limits of the Pro plan, it would cost an extra $5/month foe every 100K additional operations, and an extra $5/month for every additional 20K records.
```
gunsch:
```suggestion
The upper tier, the Pro plan, includes up to 1M records and 5M operations. If we exceed the limits of the Pro plan, it would cost an extra $5/month for every 100K additional operations, and an extra $5/month for every additional 20K records.
```

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -12,32 +12,75 @@
...
+- Writing the config file for the scraper
+- Adding CircleCI to `va.gov-team`
+- Adding a scheduled job to `va.gov-team`'s CircleCI pipeline to run the scraper once every 24 hours
+- Manually triggering the initial scraper run to populate our Algolia index
+
+#### Usage limits
+
+[Algolia's pricing page](https://www.algolia.com/pricing/) shows three different tiers, in addition to an enterprise plan. The primary differentiator between the tiers is the number of records and the number of operations (read/write/edit/delete) included with each plan.
+
+##### The Community Plan
```
gunsch: which of these plans is most relevant to your design? which one do you think we'll need to procure?

products/platform/documentation-site/research/search-design-doc.md
gunsch: I think the answer here is "we're not going to index the private repos", but that should be explicitly documented in this design

docs/docsearch-scraper-config.json
```diff
@@ -0,0 +1,63 @@
+{
+  "index_name": "dev_docs",
+  "start_urls": [
+    {
+      "url": "https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/",
```
cvalarida: Since there isn't a good way to leave comments in a JSON file, the explanation:

This line tells the scraper where to actually _start._ Note how there isn't a regex that captures a bunch of possibilities like the other two `start_urls`.

docs/docsearch-scraper-config.json
```diff
@@ -0,0 +1,63 @@
...
+        "folder"
+      ]
+    },
+    {
+      "url": "https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/.+$",
+      "selectors_key": "folder",
+      "page_rank": 0,
+      "tags": [
+        "folder"
+      ]
```
cvalarida: This tells the scraper, "Whenever the URL matches this line, we know it's a folder view in GitHub, so use the `folder` `selectors_key`."

We're also setting `page_rank` to 0 in an attempt to push these search results to the bottom. The trouble is that we can't separate out the search and indexing functions of the scraper, so we're adding all the GH tree view links to the search results. :disappointed: To account for that, we're pushing those results to the bottom.

docs/docsearch-scraper-config.json
```diff
@@ -0,0 +1,63 @@
...
+    },
+    {
+      "url": "https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/.+\\.md$",
+      "selectors_key": "file",
+      "page_rank": 5,
+      "tags": [
+        "file"
+      ]
```
cvalarida: And finally, this tells the scraper that these files—markdown files in the `va.gov-team` repo—should use the `selectors_key` of `file`.

With all these `start_urls`, to add more repos, we can [add a variable](https://docsearch.algolia.com/docs/config-file/#using-regular-expressions) to the `url` for the repo name.

Any URL that doesn't fit these `start_urls` will have no `selectors_key`. Because `selectors` below doesn't have a `"default"` property, they won't be crawled or indexed.

docs/docsearch-scraper-config.json
```diff
@@ -0,0 +1,63 @@
...
+    }
+  ],
+  "selectors": {
+    "folder": {
+      "text": ".js-navigation-item td.content"
+    },
```
cvalarida: For the folder view pages, this tells the scraper to look only at the links in this table:
image

docs/docsearch-scraper-config.json
```diff
@@ -0,0 +1,63 @@
...
+    "file": {
+      "lvl0": "article h1",
+      "lvl1": "article h2",
+      "lvl2": "article h3",
+      "lvl3": "article h4",
+      "lvl4": "article h5",
+      "lvl5": "article h6",
+      "text": "article p, article li"
+    }
```
cvalarida: This tells the scraper to look in the `<article>` element for the headings for the "levels," and paragraphs + list items for the actual content.

cvalarida (commented):
Added some comments to explain the `config.json` for posterity.

products/platform/documentation-site/research/search-design-doc.md
billfienberg: We will crawl and scrape the following public documentation sources:

- [`va.gov-team` repo](https://github.com/department-of-veterans-affairs/va.gov-team)
- [`veteran-facing-services-tools` repo](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/)
- [`vets-contrib` repo](https://github.com/department-of-veterans-affairs/vets-contrib/)
- [design.va.gov](https://design.va.gov/)
- [developer.va.gov](https://developer.va.gov/)

The list was added to the design doc in https://github.com/department-of-veterans-affairs/va.gov-team/pull/5248/commits/8ba0b54f413ef814861cfbc06bb0f1b1f7b36efc

products/platform/documentation-site/research/search-design-doc.md
billfienberg: Short answer: the Pro plan

I've updated this section, and included my calculations, in https://github.com/department-of-veterans-affairs/va.gov-team/pull/5248/commits/18b1e99e326c38006711b90b8754dfc49658f103.

products/platform/documentation-site/research/search-design-doc.md
billfienberg: Explicitly documented in c57dc1119981f03d57062e6c027f9a7abece2872

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -95,38 +138,39 @@ We could search for the same term in both the Algolia UI and the landing page to
 
 ### Security Concerns
 
-If the search input dispatched a request directly to the Algolia API, then there'd be minimal risk of the documentation search negatively affecting other systems. If the request was proxied through our API, then we should be able to rely on any existing DOS protections.
+The [DocSearch JS snippet](https://docsearch.algolia.com/docs/dropdown) requires four pieces of data; the `appId` (public), search `apiKey` (public), `indexName`, and `inputSelector`.
+
+The `appId` and search `apiKey` are both public. The search `apiKey` only allows read access. There is a separate admin API key that the crawler will use, and that admin API key will need to be configured in an environment variable in CircleCI.
```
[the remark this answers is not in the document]
billfienberg: I added a link to [Using Environment Variables in CircleCI](https://circleci.com/docs/2.0/env-vars/) in 1fa731aa04939f4fb4af75f68220b8fbfe431ede.

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -12,32 +12,75 @@
...
-1. Index/Database - We need a place to store the records that will be returned to the user.
-1. Crawler/Scraper - We need a crawler to scan our documentation sources and push data to our index.
-1. Cron job - We need a cron job to run the crawler at a reasonable interval (e.g. every 24 hours)
+---
+
+Essentially, every line of text (`<p>`, `<li>`, `<h1-6>`, etc) in a markdown file ends up as a separate record in the Algolia index. We have approximately 3K `.md` files in the `va.gov-team` repo. We ran a test crawler on the `va.gov-team` repo to estimate how many elements would match our selectors, and it found almost 150K matching elements, which equates to almost 150K records.
+
+#### Where/when to run the scraper
+
+Subject to review, we will run the crawler in a scheduled job once every 24 hours in a Circle CI pipeline for the `va.gov-team` repo.
```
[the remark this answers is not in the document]
billfienberg: Chris suggested adding a "subject to review"  qualifier before we had an estimate of how many records/operations we'd be consuming. 

Now that we have those estimates, I have removed the "subject to review" qualifier.

products/platform/documentation-site/research/search-design-doc.md
```diff
@@ -12,32 +12,75 @@
...
+
+It's possible that running the scraper once per day would consume a significant number of operations.
```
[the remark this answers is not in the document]
billfienberg: The danger is cost. And "significant" depends on what tier we use. 

I've removed this vague line, and updated the **Usage Limits** section in 18b1e99e326c38006711b90b8754dfc49658f103.

billfienberg:
Closing this PR while we wait to see if we get approval to sign up for Algolia. 

If we get approval, this PR will be reopened.

PR #5750, opened 2020-02-11
review instance login sequence diagram

omgitsbillryan:


platform/engineering/backend/review-instances.md
```diff
@@ -117,10 +117,10 @@ Get a bash shell in one of the containers:
 These instances are destroyed when the non-master branch(es) they are related to are deleted or the instance is older than 7 days.
 
 
-### User Authentication
+## User Authentication
 
-ID.me SAML configuration requires explicitly defining a callback URL via a manual process. This callback URL receives authentication information after a successful authentication process. The review instance requires a special nginx configuration that intercepts the callback to the dev-api.vets.gov server, and forwards the authentication information to the appropriate review instance (mapped by the `RelayState` parameter, which is provided to the review instance vets-api config with the `REVIEW_INSTANCE_SLUG` environment variable).
+![Authentication Flow](products/identity-personalization/login/reference-documents/auth/review_instance_auth_flow.png)
```
wyattwalter: This image reference seems to be broken when I do a "view file" from the PR menu. I think maybe you can just reference it relative to this directory (`review_instance_login_sequence.png`)

omgitsbillryan: doh, good catch

wyattwalter:
Text looks great to me, diagram is a great update. Just need to make sure the image reference works.

platform/engineering/backend/review-instances.md
omgitsbillryan: Only caveat is that it's not in the same dir - but will fix nevertheless.

PR #5886, opened 2020-02-14
Micro Front End Design Document

cvalarida:
Like it says on the tin and in the document itself, this is a design document for a proposed micro front end architecture.

To see the rendered document, go [here](https://github.com/department-of-veterans-affairs/va.gov-team/blob/be7adc60a09025d6e1981bc814c70f23f76cea01/platform/engineering/design-docs/micro-frontend-architecture.md).

### DIFF
--- a/platform/engineering/design-docs/micro-frontend-architecture.md
+++ b/platform/engineering/design-docs/micro-frontend-architecture.md
@@ -13,7 +13,7 @@ To determine an engineering approach that enables application teams to operate
 independently.
 
 This document is intended for front end and DevOps engineers on the
-Veteran-facing Services Platform (VSP) and Veteran-Facing Services (VSA) teams.
+Veteran-facing Services Platform (VSP) and Veteran-Facing Services (VFS) teams.
 
 
 ### Background
--- a/platform/engineering/design-docs/micro-frontend-architecture.md
+++ b/platform/engineering/design-docs/micro-frontend-architecture.md
@@ -72,7 +72,7 @@ release.
 
 ###### Test failures in `master`? How?
 This happens most frequently when two individual branches work on similar parts
-of the code. The tests on each branch fail, but the combined changes cause the
+of the code. The tests on each branch pass, but the combined changes cause the
 result—`master`—to fail.
 
 **TODO:** Use an actual example of where this happened.
--- a/platform/engineering/design-docs/micro-frontend-architecture.md
+++ b/platform/engineering/design-docs/micro-frontend-architecture.md
@@ -22,36 +22,30 @@ Veteran-facing Services Platform (VSP) and Veteran-Facing Services (VFS) teams.
   - **Builds are slow** and **require a lot of memory**
 - Team autonomy
   - **CI failures** affect all teams
-  - **Individual applications can't roll back**
+  - **Individual applications can't roll back** without rolling back all
+    applications
   - **Developing applications requires CMS content**
   - Application dependencies are tied together making **library upgrades
-    difficult and error prone**
+    difficult,** time-consuming, and risky
 
 ### High Level Design
 
 **Each application** will be version controlled in its own **separate
 repository** with its own **CI pipeline** and **deployment schedule**.
 
+<!-- TODO: Move this to the deploy section of the detailed design -->
 ![Separate Builds - Separate content bucket](https://user-images.githubusercontent.com/12970166/74556016-d1bbb700-4f11-11ea-9a6f-d562e9048a0a.png)
 [Link to original diagram](https://www.lucidchart.com/invitations/accept/5ca3645c-fcf8-4cd1-bad2-d8ff17c9f1ba)
 
-
-To support this, the [platform
-code](https://github.com/department-of-veterans-affairs/vets-website/tree/c8e47d202e1457807de6c2a179f98767cc0ae9cc/src/platform)
-will be broken down and made into discreet **NPM modules** which can be imported
-and used in the applications.
-
-<!--
-  Should I add a transition plan here, in the Specifics section, as a new
-  section after that, or not at all?
--->
+To support this, **the [platform
+code](https://github.com/department-of-veterans-affairs/vets-website/tree/c8e47d202e1457807de6c2a179f98767cc0ae9cc/src/platform)**
+will be broken down and **made into NPM modules** which can be imported and used
+in the applications. VSP will **provide guidance** on how to **create a new
+application** from scratch and **migrate an existing application** to a new repo.
 
 
 ## Specifics
 
-**Coming soon to a design doc near you!**
-
-<!--
 
 ### Detailed Design
 _Designs that are too detailed for the above High Level Design section belong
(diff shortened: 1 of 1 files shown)


platform/engineering/design-docs/micro-frontend-architecture.md
```diff
@@ -0,0 +1,255 @@
...
+<!-- What other problems are we running into with the current architecture? -->
+<!-- Code ownership? -->
+<!-- Dependency management? -->
+<!-- Deploy schedule? -->
+
+
+### High Level Design
+
+**Each application** will be version controlled in its own **separate
+repository** with its own **CI pipeline** and **deployment schedule**.
```
U-DON: I wanted to dig up an old documented decision on why we went with a monorepo, but couldn't find it...

Does using something like `lerna` within a monorepo with each app being its own package work within your vision?

platform/engineering/design-docs/micro-frontend-architecture.md
```diff
@@ -0,0 +1,255 @@
+# Micro Front end Architecture
+
+**Author(s):** Chris Valarida (chris.valarida@adhocteam.us)  
+**Last Updated:** February XX, 2020  
+**Status:** **Draft** | In Review | Approved  
+**Approvers:** Andrew Gunsch, Rian Fowler, Wyatt Walter
+
+<!-- Should the approvers include anybody from DEPO? -->
```
gunsch: Dror has indicated he'd like to be a reviewer on all design docs once he gets here, at least initially while he ramps up and gets a handle on our processes

platform/engineering/design-docs/micro-frontend-architecture.md
```diff
@@ -0,0 +1,255 @@
...
+
+
+## Overview
+
+### Objective
+To establish a micro front end architecture which enables application teams to
```
gunsch: careful here: this objective is a little bit "solution in search of a problem". what are the problems we're solving with the MFE arch? why are we doing it?

(you do address this below in "Problems", but probably worth reframing the objective around that)

platform/engineering/design-docs/micro-frontend-architecture.md
```diff
@@ -0,0 +1,255 @@
...
+
+
+#### The Problems
+##### Failures in one Service affect all other Services
+When one application team introduces test failures into `master`, all PRs from
+all teams that include the offending commit in `master` will fail in Continuous
+Integration (CI) because all tests are run, not just the tests for a single
+application.
+
+**Example:** Team A introduces a test failure in `master`. Team B has an urgent
```
gunsch: how often does this happen? doesn't CI generally prevent failing tests from getting checked in?

edit: I see your comment below about two conflicting changes. unless that's happening a lot, that seems like a pretty rare case to be the justification for this big of an overhaul

platform/engineering/design-docs/micro-frontend-architecture.md
```diff
@@ -0,0 +1,255 @@
...
+bug fix that needs deployed as soon as possible, so they make a branch of
+`master` and quickly fix it. Team B's branch can't be merged until Team A's
+unrelated test failure is fixed.
+
+Likewise, when there's a content issue, such as an accessibility violation, all
```
gunsch: same question, how often does this happen? what leads to it occurring?

platform/engineering/design-docs/micro-frontend-architecture.md
gunsch: @U-DON I'd definitely be interested to hear more on the monorepo vs. separate repos tradeoffs. I think Dror will as well once he arrives (already had a chat with him about it).

@cvalarida is there a way we can get some/many of the MFE benefits _without_ separating out repos? that seems like a line of research at least worth evaluating the pros/cons of

platform/engineering/design-docs/micro-frontend-architecture.md
```diff
@@ -0,0 +1,255 @@
...
+been resolved by a content editor.
+
+**Important exception:** Test failures in the latest `master` branch will not prevent
+content-only deployments since those use the latest stable `vets-website`
+release.
+
+
+###### Test failures in `master`? How?
+This happens most frequently when two individual branches work on similar parts
+of the code. The tests on each branch fail, but the combined changes cause the
```
gunsch:
```suggestion
of the code. The tests on each branch pass, but the combined changes cause the
```

platform/engineering/design-docs/micro-frontend-architecture.md
```diff
@@ -0,0 +1,255 @@
...
+result—`master`—to fail.
+
+**TODO:** Use an actual example of where this happened.
+
+
+##### Developing applications requires CMS content
```
gunsch: how does MFE address this?

platform/engineering/design-docs/micro-frontend-architecture.md
cvalarida: Yup! This is definitely something I thought about. I think something like `lerna` is _interesting_, but I don't know what advantages it would have. My understanding is fairly limited. From what I can tell, the advantages of a monorepo are minimal, and it's yet more overhead engineers need to understand before they can be productive.

Mostly I was concerned the CI. Specifically, if I'm making a PR to App 1, how do we tell the CI to only run unit tests on App 1? (Because the rest of the unit tests don't apply.) It's probably possible, but I have no idea how to do it.

Additionally, I'm concerned about incidental coupling between applications. Specifically, it'd be easy for engineers working on App 1 to import something from App 2 without the App 2 team knowing about it. When the App 2 team makes a change to that thing, it'll affect App 1. If the repos were split up, this would still be possible, I think, but it would be harder. Like, you'd have to deliberately list App 2 as a dependency in `package.json` rather than just tossing an `import` statement in a JS file.

That said, I'm not very opinionated on this. I like the clear separation of concerns we get when splitting out the applications into their own repositories, but I'm definitely not against the monorepo approach. I'm super interested in hearing what other folks have to say.

platform/engineering/design-docs/micro-frontend-architecture.md
cvalarida: In the last few months, this has happened maybe half a dozen times at least. It's not super rare.

platform/engineering/design-docs/micro-frontend-architecture.md
cvalarida: Back before broken links in a build triggered a fallback to the cached content, it was as much as three or four times times a week. After we made that change, it's been pretty rare. I'd guess maybe three or four times in the past four months or so?

I don't know what all the avenues are that lead to this happening, but last time, it was because a content editor entered invalid HTML into a WYSIWYG component. I think it was forgetting to close a `<ul>`.

As we get more and more content editors on board, the chances of this happening only increase. I think the CMS team mentioned a safety measure they're going to put in place, for what it's worth.

platform/engineering/design-docs/micro-frontend-architecture.md
cvalarida: :man_facepalming: Good catch.

platform/engineering/design-docs/micro-frontend-architecture.md
cvalarida: Each application will have its own build process which creates its HTML, JavaScript, and CSS files. This would be independent from the content build process.

:thinking: Guess I didn't call that out explicitly. Just implied it with "its own CI pipeline." :+1:

platform/engineering/design-docs/micro-frontend-architecture.md
gunsch: if it's from bad CMS content, this seems like the kind of thing that should be improved by CMS fixes to their own CI (or FE build), not rearchitecting our entire repo + deployment process :)

if the other things are "once a month", same comment. they're nice things to fix, but they don't justify this level of overhaul.

platform/engineering/design-docs/micro-frontend-architecture.md
gunsch: as a reference point, part of how Google's monorepo succeeds is a CI system that tries to determine "affected tests" for a given change, and only run those. it generally depends on [Bazel](https://bazel.build/) to do that, by using the dependency tree of build targets.

that said, some of the concepts are still usable. the naive approach I'd start with: if code and tests are located in structurally consistent ways, we could probably provide utilities to run only tests within certain directories (test filters?), and then whatever script the CI executes decides based on which files are modified which tests need to get run. changes to common code probably still run all tests. is that something that'd be workable in our repo?

U-DON: ^ This is actually an approach I took to running the unit tests for the [early version of a CircleCI workflow](https://github.com/department-of-veterans-affairs/vets-website/blob/a969e53de3fa7066119d9fc9dc4a6c9914642bb3/script/run-isolated-tests.sh) I adapted from our current Jenkins pipeline. It's a check that's running on PRs now (non-blocking), so we can observe its efficacy.

platform/engineering/design-docs/micro-frontend-architecture.md
rianfowler: is this scoping the problem away from the content builds?

platform/engineering/design-docs/micro-frontend-architecture.md
rianfowler: - I don't think Lerna really meets our use case. Don't think of it as a "multiple package in one repo" tool but as a "tool to manage multiple npm modules published from a single repo." It runs npm scripts from the root directory- running package specific scripts can be a challenge with the way it manages dependencies. 
  - We should evaluate it- I'm not an expert. But I had issues with things like running package specific linting rules from the root npm scripts. I don't think its behavior is intuitive and could be confusing to VFS teams. 
- I would like to encourage better boundaries and decoupling of the website pieces. Some of this has led to issues like app specific concerns in the global build script, forms system, and helpers.
  - separating the forms system from the website repo was problematic because of how entangled it is with apps and with how poorly it separates concerns 
- Versioning apps can encourage behaviors that reduce friction during ownership changes (app can change ownership every 1-2 year): 
  - documenting changes in release notes 
  - thinking of changes in terms of explicit releases

rianfowler (commented):
good start- thanks for putting this together

platform/engineering/design-docs/micro-frontend-architecture.md
cvalarida: Agreed on all points!

platform/engineering/design-docs/micro-frontend-architecture.md
cvalarida: Cases worth adding (note to self):
- When a team writes a test that will inherently change over time
  - Such as the one [this PR](https://github.com/department-of-veterans-affairs/vets-website/pull/11793) addressed
- When a team writes a test that modifies a global property without resetting it
  - The failure looks something like [this](http://jenkins.vfs.va.gov/blue/organizations/jenkins/testing%2Fvets-website/detail/master/6136/tests)
  - It may not be obvious, and it may only break tests given a certain `CHOMA_SEED`

platform/engineering/design-docs/micro-frontend-architecture.md
```diff
@@ -0,0 +1,255 @@
...
+- Validating the content for broken links, accessibility violations, etc.
+- Building all JavaScript bundles
+  - Both the common bundle and individual application bundles
+- Building all CSS bundles
+  - Both common and application-specific
+- Running all >5k unit tests for all applications
+- Running all end-to-end tests for all applications
+
+
+##### Deploy
```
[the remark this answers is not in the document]
cvalarida: Ah, cool. I thought I read that somewhere, but didn't realize we had actual documentation for it. Thanks! :raised_hands:

platform/engineering/design-docs/micro-frontend-architecture.md
cvalarida: I'm not sure I understand your question, @rianfowler.

cvalarida:
I just pushed up some changes I had been working on just to get them out for posterity, but I'm closing this one in favor of several smaller, more narrowly-focused design docs that are intended to get us in basically the same place. The scope of this one is too broad.

I think the design docs I have will be something like follows:
1. Split out the content build into its own job
2. Split out the platform code from the application code
3. Split up the applications into separate build and test jobs
4. Split up the application deploys into separate jobs

Not sure about 3 & 4 yet or the actual order of operations here, but that's my vague-ish goal.

PR #5935, opened 2020-02-17
Specify that teams must give a peer review first

bkjohnson:
We keep encountering scenarios where the frontend-review-group is asked to provide a review before a peer review has even been given.  When this happens it would be helpful if we could link to documentation that clearly defined what to do in order for the review group to provide a review.


cvalarida:
cc @gunsch

gunsch (approved):

PR #6657, opened 2020-03-07
Updates to change some references of vets.gov-team to va.gov-team.

karlbrown-va:
* Updates to change some references of vets.gov-team to va.gov-team. 
The general approach followed was to look at the context - if it was documentation
referring to the migration, those instanced were not modified.
* Fixed a few broken links as well. There are a lot more, that can be
dealt with over time.


products/structured-data/integration-efolder.md
```diff
@@ -20,6 +20,7 @@ The goal of the eFolder API is to allow unstructured documents to be directly up
 * related docs on Central Mail:
     https://github.com/department-of-veterans-affairs/vets-api-clients/blob/master/services/vba_documents/UnderstandingDocumentProcessing.md
 * https://github.com/department-of-veterans-affairs/vets.gov-team/tree/master/Products/Identity/eFolder
```
jenstrickland: Just checking, in case this vets.gov-team link wasn't intended to keep.

jenstrickland (approved):

products/structured-data/integration-efolder.md
karlbrown-va: I think I left that one in; I actually don't have access to vets.gov-team yet - if you can see it and see that the contents are the same you could remove it.

PR #6138, opened 2020-02-25
Flag for manual review design doc

bkjohnson:
This is a draft design doc for what we've been calling the automated "flag for manual review" process.  It is not _strictly_ a design doc since I ended up creating the solution ahead of time as I was experimenting with Github's API and Circle's CI config, but I tried to capture some design decisions and open questions.

All feedback welcome.

[Here is an earlier issue](https://github.com/department-of-veterans-affairs/va.gov-team/issues/2521) which describes some of the goals. I will try to work that into the design doc somehow.

PR #6915, opened 2020-03-16
Separate content build

### LINKED ISSUE #5822: Write a design document for content build separation
## Description
Write a design document for the separation of the content build from the application build. Explanation of why this is important can be found in the document itself.

cvalarida:
This design document is the first in what I expect to be a series of documents for re-architecting the front end build for VA.gov. This one specifically is for separating out the CMS content build from the applications.

View the latest rendered markdown [here](https://github.com/department-of-veterans-affairs/va.gov-team/blob/separate-content-build-cv/platform/engineering/design-docs/2020-04-03-separate-content-build.md).

Resolves https://github.com/department-of-veterans-affairs/va.gov-team/issues/5822


platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Objective
+The objective here is to define an approach to isolate the concerns of the
+content build from the front end application build and vice versa.
+
+This design document is intended for front end engineers and DevOps engineers on
+the Veteran-facing Services Platform (VSP) and other Veteran-Facing Services
+(VFS) teams. Most directly affected is the Content Management System (CMS) team.
+
+### Background
```
rianfowler: There's good information in here but I think we need to work on the outline and organization to tie it together more. Some ideas: 

- Start by describing things in terms of business goal. This defines the core pieces you'll refer to in your design document. 
  - VA.gov is composed of static content (.html), frontend applications (.js), and styling (.css). All of these are deployed to and hosted from S3. 
  - static content is deployed to S3 as content updates are published in the Drupal CMS
  - frontend applications and styling are deployed daily as part of the vets-website CI pipeline 
- Describe the major processes in terms of workflows (triggers, inputs, processes, outputs, exceptions)
  - automatic triggers: 
    - static content: CMS publishing event?
    - frontend applications and styling: scheduled weekdaily jenkins job
  - manual triggers 
  - metalsmith 
    - I wouldn't describe from the implementation perspective. Metalsmith is an orchestrator- even though the stuff is in one script, there are a couple of processes contained. the static content build, the application + styling builds. 
    - Part of the idea is to describe a cleaner mental model for these things so when you describe the design, it's more obvious why we're going this route. Like "here's this clean concept" and "here's this legacy messy implementation" and "here's the cleaner implementation that matches the clean concept" 
  - validations

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,282 @@
...
+      - Falls back to the cached content in S3 from the last successful deploy
+        (full or partial)
+        <!-- TODO: Verify that either deploy will cache the pages.json and files -->
+    - Runs the accessibility tests and link checker
+
+See [here](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started/workflow/deploy/)
+for more details.
+
+#### Pain points
+- Engineers developing applications need CMS content to build their applications
```
rianfowler: In general, pain points can be described in terms of dependencies, blockers, friction, and volatility. This is what I'm scanning for when I read this document- abstractions of problems in workflows. 

- Validation failures unrelated to a code change can **block** automatic pull request approvals
  - more detail on why this is a problem. in this case I think it's an isolation problem- publication is not tied to any validation so we perform that validation in an unrelated workflow (the website CI)
- Building content adds **friction** to running frontend applications locally
  - Frontend app development relies on the output of the CMS build process. No steps are coordinated between the static content and frontend application build processes. 
  - Developers are required to run this build because the implementation of these two processes is entangled

rianfowler (commented):
good start- added a couple of comments and I'm going to keep thinking about it.

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+    wasn't intended to be client-facing
+
+#### Single S3 bucket
+Instead of deploying content to a separate S3 bucket from the applications, we
+could continue to deploy them both to the same bucket. We decided to split up
+the deployment to:
+- Promote a higher separation of concerns
+  - **Question:** Is this even a thing we should be worried about?
+- Enable simpler rollbacks
+  - **Question:** Is this valid?
```
Clark-W-Griswold: Something to consider here is the 100 bucket limit per account. If we grow greater than 100 teams, this will be a problem:
https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

Single bucket multiple folders is what I would suggest. s3 has a flat structure and it treats the folder structure as keys to the object. There is only a limit on the object size and not the bucket size:
https://docs.aws.amazon.com/AmazonS3/latest/user-guide/using-folders.html

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+    - Input: CMS content
+    - Output: HTML files, assets from the CMS (PDFs, images and the like)
+1. Application build
+    - Input: `vets-website` code
+    - Output: JavaScript and CSS bundles
+
+The output of both these builds will be deployed to separate buckets. The
+reverse proxies will route traffic intelligently to the appropriate bucket given
+the request.
+
```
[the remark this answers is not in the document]
cvalarida: This is probably a good idea, but out of scope for this document.

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+- The content build will live in a separate repository
+  - **Question:** What should we call this?
+  - **Question:** Can the application build be run in Circle?
+  - **Question:** Where can we run the content build?
+- Content will be deployed to a separate S3 bucket from the application code
+- The S3 buckets will have versioned directories in them with the contents of
+  each deployment
+  - Old directories will be removed after 90 days
+    - **Question:** What makes sense? This is a totally arbitrary number
+
```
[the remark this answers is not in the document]
cvalarida: Great question! I think we'll start with one version for everything as the MVP, and we can add more advanced routing later through the forward proxy. I could call that out in the Future Work section. :+1:

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+See [here](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started/workflow/deploy/)
+for more details.
+
+#### Pain points
+- Engineers encounter **confusing build errors** that are due to the mismatch in
+  outdated locally cached content and updated Liquid templates
+- Fetching content from the CMS and building HTML files when developing
+  applications **slows down the local builds** for engineers unnecessarily
+- When their applications have nothing to with the CMS
```
gunsch: is this an unfinished sentence, or a misplaced bullet point?

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+[Metalsmith](https://github.com/segmentio/metalsmith) is a static content build
+pipeline orchestrator. `vets-website` uses it to manage the steps in the content
+build. The highlights of what this script does include:
+- Fetching static content from Drupal
+- Building static HTML pages from the content
+- Triggering Webpack to build JS and CSS bundles
+- Creating the HTML pages for the React Single-Page Applications (SPAs)
+- Checking all HTML pages for broken links
+
+The script can be found [here](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/stages/build/index.js).
```
gunsch: readability pro-tip, never link the word "here"

```suggestion
See [the build script](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/stages/build/index.js) for more details
```

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+  - **Question:** Where can we run the content build?
+- Content will be deployed to a separate S3 bucket from the application code
+- The S3 buckets will have versioned directories in them with the contents of
+  each deployment
+  - Old directories will be removed after 90 days
+    - **Question:** What makes sense? This is a totally arbitrary number
+
+#### Routing
+- The nginx will route their traffic based on some configuration file
+  - **Question:** What configuration file? How does it work?
```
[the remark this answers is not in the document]
cvalarida: I think I'd prefer to keep configuration out of the browser. It's just a hunch, but I feel like that'd come back to bite us in some unexpected way down the road. Seems weird to be like, "Here's the version, you figure out what to go get," instead of just serving the right thing, y'know?

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+2. [Full deploy](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started/workflow/deploy/#full-deploy-of-vagov-client-app)
+    - For `vets-website` code changes
+    - Creates a new release and deploys it
+    - Fetches the latest Drupal content
+      - Falls back to the cached content in S3 from the last successful deploy
+        (full or partial)
+        <!-- TODO: Verify that either deploy will cache the pages.json and files -->
+    - Runs the accessibility tests and link checker
+
+See [here](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started/workflow/deploy/)
```
gunsch:
```suggestion
See [the deployment documentation](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started/workflow/deploy/)
```

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+#### Deployment
+There are currently two kinds of deployments:
+1. [Partial deploy](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started/workflow/deploy/#partial-deploy--static-page-changes-only)
+    - For static pages only
+    - Uses the latest `vets-website` release and static assets
+      - Does not trigger Webpack to build the JS and CSS bundles
+      - Instead,
+        [fetches](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/stages/build/plugins/configure-assets.js#L22-L23)
+        the assets [from S3](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/stages/build/plugins/download-assets.js#L111)
+        - Process is triggered [here](https://github.com/department-of-veterans-affairs/vets-website/blob/master/Jenkinsfile.content#L32)
```
gunsch:
```suggestion
        - Process is triggered [from this Jenkins config](https://github.com/department-of-veterans-affairs/vets-website/blob/master/Jenkinsfile.content#L32)
```

gunsch: tiny nit: each of these links is going to get out of date fast since you're linking to `master` with line numbers

platform/engineering/design-docs/separate-content-build.md
cvalarida: Ooh, interesting. :thinking:

Okay, so you're proposing achieving the same _kind_ of isolation of the applications and content by using separate directories in a single bucket?

I think the reason I came up with the multiple bucket idea was because I couldn't figure out how independent rollback would work, but with versioned folders like you mentioned in our chat the other day, I think that pretty much solves that problem.

I'd be down for this single-bucket approach myself, but I'd like to hear opinions from @nathanhruby or @wyattwalter.

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+
+The **[CI pipeline](#ci-pipeline)** runs this build script (among other things)
+whenever a commit is pushed to a branch in GitHub. On a branch, this build must
+be successful **before the branch can be merged.**
+
+The output of this build are **[deployed](#deployment) to and hosted from an AWS
+S3 bucket**. Applications are deployed on a weekdaily schedule. New content is
+deployed by a manual trigger from the CMS and by the weekdaily application
+deployment.
+
```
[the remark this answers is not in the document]
cvalarida: I can call it out here too, yeah. :+1: 

I'll even add that link to the document! :smile: Thanks!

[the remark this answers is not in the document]
cvalarida: @nathanhruby Can you verify that this is what we do now? I'm reasonably confident, but know so little about he details of this process. :grimacing:

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+#### Application
+
+##### Build
+We have automated unit and end-to-end tests that are run on every CI build.
+
+##### Deploy
+**Question:** How _do_ we tell whether or not the deploy was successful and the
+expected version of the application code is live in production?
+
```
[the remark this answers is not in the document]
cvalarida: I had the idea to add the app version as a Sentry tag quite some time ago too. I like that. My only concern (which I mentioned in another comment) is around using that to grab the right asset on the front end. Just makes me a bit nervous.

platform/engineering/design-docs/separate-content-build.md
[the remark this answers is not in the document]
gunsch: I suspect that "CloudFront in front of S3" is a pretty separate FE conversation / architecture change we consider have, but we already depend pretty directly on S3 as-is, so that's at least not a new pattern. (IIRC there are govt regulation concerns with CloudFront that don't work for us)

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+  - We have a safeguard that mostly works; if we run into a failure using the
+    latest CMS content, we try again with the cached content in S3 from the last
+    successful build
+    - This doesn't work when the GraphQL query has been updated; that
+      invalidates the cache
+    - Errors in liquid templates would potentially break the build and falling
+      back to the cache won't help
+
+### High Level Design
+The current front end build will be split up into two distinct builds:
```
gunsch: To confirm: this is talking just about splitting up the _build_ step, not changing anything w/r/t deployment?

Could you explain in this doc how these two build types map to the partial vs. full deploys?

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+1. Content build
+    - Input: CMS content
+    - Output: HTML files, assets from the CMS (PDFs, images and the like)
+1. Application build
```
gunsch: I feel like there's a place for some diagrams here about how these two builds interact

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+
+**Another important note:** There is no way to coordinate these two deploys to
+make an application live for the first time. The process will be to manually:
+1. Verify the application assets are live in production
+1. Publish the landing page
+1. Deploy a new content build
+
+## Specifics
+
+### Detailed Design
```
gunsch: I don't see anything about how these builds are triggered. when are these builds run and deployed? who can run/deploy them?

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+
+#### Build and deploy
+- The content build will live in a separate repository
+  - **Question:** What should we call this?
+  - **Question:** Can the application build be run in Circle?
+  - **Question:** Where can we run the content build?
```
gunsch: are you intending to answer these in this doc? :)

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+- Content will be deployed to a separate S3 bucket from the application code
+- The S3 buckets will have versioned directories in them with the contents of
+  each deployment
+  - Old directories will be removed after 90 days
```
gunsch: how? what removes them?

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Detailed Design
+
+#### Build and deploy
+- The content build will live in a separate repository
+  - **Question:** What should we call this?
+  - **Question:** Can the application build be run in Circle?
+  - **Question:** Where can we run the content build?
+- Content will be deployed to a separate S3 bucket from the application code
+- The S3 buckets will have versioned directories in them with the contents of
```
gunsch: can you more concretely specify what this directory structure will look like?

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+## Specifics
+
+### Detailed Design
+
+#### Build and deploy
+- The content build will live in a separate repository
+  - **Question:** What should we call this?
+  - **Question:** Can the application build be run in Circle?
+  - **Question:** Where can we run the content build?
+- Content will be deployed to a separate S3 bucket from the application code
```
gunsch: what bucket?
do we need separate buckets for staging/prod?

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+  each deployment
+  - Old directories will be removed after 90 days
+    - **Question:** What makes sense? This is a totally arbitrary number
+
+#### Routing
+- The nginx will route their traffic based on some configuration file
+  - **Question:** What configuration file? How does it work?
+
+#### Testing changes to applications
+- The `webpack-dev-server` will create application landing pages for local
```
gunsch: this seems like something that needs a lot more details

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+of `build/` to `build-content/` for clarity.) Once the implementation is
+complete, this code will **move to a new repository.**
+
+The **application build** will consist of **webpack configuration files** found in
+the **root of the project.** If we need any script files to aid in the build
+process, they will be located in `vets-website/scripts/build/`.
+
+### Testing Plan
+
+#### Conslidated build script
```
gunsch:
```suggestion
#### Consolidated build script
```

gunsch (commented):
first round of feedback!

platform/engineering/design-docs/separate-content-build.md
cvalarida: Ah, dang it. Yeah, I rephrased this section and that point got orphaned. :face_with_head_bandage:

platform/engineering/design-docs/separate-content-build.md
cvalarida: Right, this specific line is talking about the build step. The document as a whole I think has to address the deployments, but that's not this part.

I think I can add some points to the section below to map the new builds to the current deploys. :+1:

platform/engineering/design-docs/separate-content-build.md
cvalarida: I wasn't sure how to represent this visually in a way that was more helpful than bullet points. Any suggestions?

platform/engineering/design-docs/separate-content-build.md
cvalarida: Ah, good point. Short story:
- Application build is triggered on all commits pushed to GitHub (same as it is now)
- Application code is deployed on a weekdaily schedule
  - Probably in Jenkins at first until we can move it to Circle or something
- Content build and deploy (together) are triggered from Drupal
  - Same as it is now—I just don't know the specifics of how that all is wired up
  - This can be triggered by users with a certain permission in Drupal (again, not sure the specifics)

I'll sync with @ElijahLynn to get the specifics on the Drupal trigger and add those points in.

platform/engineering/design-docs/separate-content-build.md
cvalarida: Yup! As soon as I have answers! :sweat_smile: 

For now, these are questions I either don't know how to answer myself or would like input from you and the other reviewers. The final draft will have these all answered or removed.

platform/engineering/design-docs/separate-content-build.md
cvalarida: No idea! I know just enough to be dangerous here. After @Clark-W-Griswold's review, I'm reconsidering this multi-bucket approach, so this may get removed.

I'd appreciate any insight from @nathanhruby or @wyattwalter.

platform/engineering/design-docs/separate-content-build.md
cvalarida: Probably a script somewhere. _Waves hands vaguely._

Is this a thing we even need to worry about, @wyattwalter?

platform/engineering/design-docs/separate-content-build.md
cvalarida: Short story: Currently, these landing pages are generated in the Metalsmith build. When we move the Metalsmith build out of `vets-websites`, it'll introduce a couple needs:
- Engineers still need those pages to serve up the applications they're building
- E2E tests still need to be run against "live" pages, not just pointed at JS bundles

To solve these problems, my idea was to make the dev server spin up temporary landing pages. These temporary pages would not be in production, but are meant to mirror it.

:thinking: There are probably other concerns this new process would introduce, but I'm not sure what all they are yet.

Is this the kind of detail you think should be added here, or are you thinking about something else entirely?

gunsch: > There are probably other concerns this new process would introduce, but I'm not sure what all they are yet.

That's my concern --- and that's what this design doc should address! So yes.

Go back to "you should be able to hand this doc off to a junior engineer and expect them to be able to implement it." Major questions / concerns should be cleared out already, and ideally there wouldn't be any unanticipated roadblocks left.

platform/engineering/design-docs/separate-content-build.md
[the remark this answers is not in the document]
cvalarida: Reading that versioning document more closely, I think it's _not_ what we're doing. I think we have folders inside S3 that we use as versions...I think.

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,374 @@
...
+### Detailed Design
+
+#### Build and deploy
+- The content build will live in a separate repository
+  - **Question:** What should we call this?
+  - **Question:** Can the application build be run in Circle?
+  - **Question:** Where can we run the content build?
+- Content will be deployed to a separate S3 bucket from the application code
+- The S3 buckets will have [versioned
+  directories](https://docs.aws.amazon.com/AmazonS3/latest/dev/Versioning.html)
```
cvalarida: Do we want versioned buckets like this, or use directories within the bucket named by like commit hash or timestamp or something?

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Detailed Design
+
+#### Build and deploy
+- The content build will live in a separate repository
+  - **Question:** What should we call this?
+  - **Question:** Can the application build be run in Circle?
+  - **Question:** Where can we run the content build?
+- Content will be deployed to a separate S3 bucket from the application code
+- The S3 buckets will be [versioned](https://docs.aws.amazon.com/AmazonS3/latest/dev/Versioning.html)
```
cvalarida: Do we want versioned buckets like this, or use directories within the bucket named by like commit hash or timestamp or something?

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,381 @@
...
+  generate temporary landing pages for React applications
+  - This will happen for all build types
+    - Engineers will be able to access the site on a `localhost` build
+    - CI can run E2E tests against production code without the output of the
+      content build
+- The E2E test script in CI will start a process to serve these pages so the
+  tests can run a browser and access the applications
+- **Important note:** These HTML files will not be served in production; the
+  landing pages for the applications come from the content build
+  - **Question:** How do we make this happen?
```
cvalarida: I'm thinking something vaguely like putting the HTML files in a `build/` or `landing-pages/` or something, and the "real" Webpack assets (JS, CSS, etc.) in like a `dist/` or something.

But then how do we go from there to either
1. Not pushing the temporary landing pages to the S3 bucket, OR
1. Not routing to them when the request for the application page is made

platform/engineering/design-docs/separate-content-build.md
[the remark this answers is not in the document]
rianfowler: FWIW I think we need to have JS/CSS in the critical rendering path referenced in the initial document response. `fetch()`ed JS has very low download priority on browsers- lower than initial doc CSS and images.

platform/engineering/design-docs/separate-content-build.md
rianfowler: Would this require changing the URL in the `<script/>` tag on every release?

platform/engineering/design-docs/separate-content-build.md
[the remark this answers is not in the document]
nathanhruby: What we do currently is have the build generate artifacts in the form of tarballs of what the bucket should be, and store those in a dedicated s3 bucket (and/or github as part of a release).  

The deploy then grabs the tarball for the git commit you are deploying and does a sync of data to the correct buckets (we deploy to a site bucket as well as an assets bucket).

Thus a redeploy of any git sha (including previous) is how one would do a "rollback"

platform/engineering/design-docs/separate-content-build.md
nathanhruby: the 100 bucket limit can be raised to 1000 whihc is not a limit I think we'd hit.  

Note that for static content sites the bucket must be named after the site being served, so with a single bucket you are also going to force a single hostname (as is the case now).

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,370 @@
...
+assets in the `/applications/` directory.
+
+![Coordinating deployments to S3](images/coordinating-deployments-to-s3.png)
+
+To ensure that one build won't override files in the other, we'll have to
+1. Update
+   [`vets-website-deploy.sh`](https://github.com/department-of-veterans-affairs/devops/blob/2c63e6d62fa5ad2b6b87478e30d0a7bc2820c252/ansible/deployment/roles/deploy-vets-website/files/vets-website-deploy.sh)
+   to copy only to the `/applications/` directory
+1. Create a new deploy script for the content deployment which copies everything
+   over _except_ the `/applications/` directory
```
cvalarida: This seems not quite right, but....I don't remember the specifics. :bat-signal: @nathanhruby

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,370 @@
...
+
+### Detailed Design
+
+#### Content build in a separate repository
+The content build will live in a separate repository called `content-build`.
+This repository will contain only code pertaining to the content build.
+
+![Separated content code base](images/phase-2-separated-content-build.png)
+
+#### Coordinating deployments to S3
```
cvalarida: This is maybe not the best description for what I'm talking about. What other words can I use instead?

platform/engineering/design-docs/separate-content-build.md
cvalarida: Also, would this be a good place to add the bit about updating the "poor man's CDN" (as you so eloquently described it the other day :smile: )? Looking at the script I linked to, it kind of reads like that's the one that copies it to the assets bucket, but....I'm not 100% sure and would appreciate any clarification you can lend.

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -152,56 +152,65 @@ The current front end build will be split up into two distinct builds:
...
-- The nginx will route their traffic based on some configuration file
-  - **Question:** What configuration file? How does it work?
+#### Content build in a separate repository
+The content build will live in a separate repository called `content-build`.
+This repository will contain only code pertaining to the content build.
+
+![Separated content code base](images/phase-2-separated-content-build.png)
+
+#### Coordinating deployments to S3
+We use S3 to serve a static site. To do this, the bucket consists of a file tree
```
gunsch: For my own knowledge: what handles this translation of "requested URL" to "bucket to fetch file from"?

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -152,56 +152,65 @@ The current front end build will be split up into two distinct builds:
...
 make an application live for the first time. The process will be to manually:
 1. Verify the application assets are live in production
-1. Publish the landing page
+1. Make the landing page live in production in the content build
+    - Depending on the answer to [this
+      question](#react-application-landing-pages), this may mean publishing the
+      page in Drupal or switching the flag in the JSON file
 1. Deploy a new content build
 
 ## Specifics
```
gunsch: I feel like there's an open question here around... how do you know which content + application builds are compatible? If content + application can be pushed separately from another, what kind of drift is possible, and how do we mitigate that?

Related q: if in an emergency, you have to roll back to a "known good release", how do you know which one that is, if you now have two separate released units that work together in tandem to serve the site?

platform/engineering/design-docs/separate-content-build.md
cvalarida: I believe this is the forward proxy? cc @nathanhruby

platform/engineering/design-docs/separate-content-build.md
cvalarida: Yeah, good question. So the only relationships between the two builds that I can think of right now are:
- The application landing pages being present altogether
- The content on those landing pages
  - `<script>` tags
  - The DOM nodes that the React applications look for to mount on
  - This also is applicable to pages with "widgets;" they'll also need the `<script>` with the right `src` and DOM node to mount on

## Types of drift

### Missing application pages
If we had to revert a content release for whatever reason, we _may_ lose an application page.

### Contents of an application landing page change
The contents of the pages shouldn't change between releases. (I have a `todo` item to add a section on changing the cache-busting mechanism from adding hashes to bundle names to using a cache control header.)

### Rolling back a newly-released application's code, but not landing page
If we release a new application and had to roll the application deployment back, we could end up with a landing page that points to a JS bundle that doesn't exist.

### Landing page deployed before code
The content release which includes a newly prod-facing application landing page is deployed before the code is deployed to production.

I addressed this in the document; I think it'll have to be up to the teams releasing a new application to ensure that the code is live before they make the page live. I'd love to hear alternatives to this, though!

## Possible solutions
- Checking applications manually
  - Maybe not as bad an option as it sounds
  - We should know when an application page has been added and the content release has been reverted (instead of a new release deployed to fix whatever problem occurred)
  - We should also know when the application page or template changed
- Setting up an out-of-band deploy-validation test
  - What I'm envisioning now is some kind of automated browser test that loads up each page we expect to have `vets-website` code served at
  - An idea I had to serve as something of a bridge between the two systems is a kind of "application registry" repo
    - It would basically just be a single JSON file with information about each application such as the URL it should live at and whether it should be available in production (we can add whatever else we need later)
    - We could use this to tell that deployment tester where to look for each application
  - Caveat: This only tests _after_ the deploy, so it won't _prevent_ a "bad" deployment
- What else?

## Post script
This kind of drift is technically possible now (though we'd have to like change the application landing page template to not include that DOM node or something to make it happen). We currently have no way of automatically checking, so anything we do here is an enhancement.

cvalarida: That was a lot of words! How much of this should I add to the document?

platform/engineering/design-docs/separate-content-build.md
cvalarida: Also, to be clear, after talking with Nathan, I changed the approach from multiple buckets to a single bucket, so this is the same as it is now.

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,381 @@
...
+
+#### Cache busting application code
+To ensure the latest code is downloaded by the browser, we have to bust the
+browser's cache. **Currently**, we're adding hashes to the asset file names and
+changing the `<script>` tag's `src` property to point to the file name with
+hash.
+
+Once we split out the content build, however, it's infeasible to continue this
+approach. Instead, we'll rely on the (`Cache-Control`
+header)[https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control].
```
cvalarida: Anything to add to this, @rianfowler?

platform/engineering/design-docs/separate-content-build.md
cvalarida: Added a section on this below. It's pretty sparse, but it's what I know.

platform/engineering/design-docs/separate-content-build.md
gunsch: most/all of it? answers to questions like this on a design doc should generally be "I've updated the doc with the information you requested" :)

rianfowler: >how do you know which content + application builds are compatible?

We don't know this. We won't know this. The CMS and application builds should be completely separated. Apps teams should explicitly manage the elements in the CMS that their react applications hook to. If the elements aren't there or don't work, then it has to be fixed in the CMS. 

If we want monitoring in place to make sure that someone didn't break an app by changing something in the CMS, then that should be handled as a monitoring concern. to me this is a CMS issue- if an element in the content is driving an application, then there should be some sort of locking or alerting when that changes.

rianfowler: >if in an emergency, you have to roll back to a "known good release", how do you know which one that is, if you now have two separate released units that work together in tandem to serve the site?

We can only roll back to a known good application release. My understanding is that the CMS is going to be deploying content up to every 5 minutes and could have hundreds of editors working in it at a time. We need to isolate the application deploy from this as much as possible.

cvalarida: I've added this as a risk with notes about how what we're doing now.

gunsch: > We don't know this. We won't know this. The CMS and application builds should be completely separated. Apps teams should explicitly manage the elements in the CMS that their react applications hook to. If the elements aren't there or don't work, then it has to be fixed in the CMS.

One thing I've seen done for this is a policy around waiting for _two_ prod releases before referencing something new.

i.e. if an application team checks in a new React element Monday am, it goes to prod Monday pm, and then they reference it in CMS Monday evening... that puts them in danger of an application-build LKGR rollback breaking their application.

but if we have a policy that says: check in new React element Monday am, wait for TWO releases to go out before referencing it in CMS, then they should still be safe from a LKGR rollback change.

rianfowler: >check in new React element Monday am, wait for TWO releases to go out before referencing it in CMS, then they should still be safe from a LKGR rollback change.

@gunsch this sounds good. Flipper doesn't support static content changes so maybe whenever we figure that out, it can smooth out this process.

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,477 @@
...
+    - This maps pretty cleanly to the current partial deploy
+1. Application build
+    - Input: `vets-website` code
+    - Output: JavaScript and CSS bundles
+    - This essentially maps to the current full deploy minus the content
+
+The output of both these builds will be deployed to a single S3 bucket with
+safeguards in place to ensure they don't override each other.
+
+**Another important note:** There is no way to coordinate these two deploys to
```
rianfowler: I think we should work this paragraph some more. 

- use it to summarize the tradeoffs 
- tradeoffs: 
  - the two processes are no longer block each other when validations fail and no longer share data 
  - cache busting currently relies on using new file names in static content on every deploy. splitting the processes will require using a different cache busting technique like `cache-control` headers 
  - new application releases will require manual coordination between CMS and website deployment processes

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,477 @@
...
+  - **Question:** How will we validate links to `vets-website` assets?
+  - **Question:** Where will this job be run? Jenkins? Nomad? Circle?
+- The application build will be responsible for creating the `settings.js`
+  - It's currently a step in the Metalsmith script
+- The landing pages to the applications will be created in the CMS build
+- The CMS repo will no longer import `vets-website`, but rather the new content
+  build repo
+  
+#### React application landing pages
+**Important note:** JavaScript application landing pages will be created during
```
rianfowler: Rework this to emphasize that some static pages are not sourced from the CMS and are manually generated by the current build process. These pages will need to be added to the CMS and this section is a proposal to continue generating them with the build script until they are migrated into the CMS.

platform/engineering/design-docs/separate-content-build.md
rianfowler: I see you have a cache busting section that can be referenced. Maybe just move this into the same level section as "new application release" and reference both sections here.

platform/engineering/design-docs/separate-content-build.md
rianfowler: We need to summarize the trade off of using the `cache-control` header:

browsers will always make a request before using their cache for the main document
  - higher traffic for S3
  - lower performance for users
- new deployments will be available to browsers only after the the cache expiration time

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,477 @@
+# Separate Content Build
```
gunsch: filename: can you rename this to something with a YYYY-MM-DD date?

i.e. 2020-03-31-separate-content-build.md

platform/engineering/design-docs/separate-content-build.md
cvalarida: Has this been addressed by the blurb before the Build Script section?

platform/engineering/design-docs/separate-content-build.md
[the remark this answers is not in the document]
cvalarida: We can change this up later if we feel the need, but for now, I'm checking this off as out-of-scope for this document.

I look forward to your design document on the deploy process you're thinking up, @Clark-W-Griswold :grin:

platform/engineering/design-docs/separate-content-build.md
cvalarida: I added the points I mentioned. Does that resolve this?

platform/engineering/design-docs/separate-content-build.md
cvalarida: Does it make sense to outline that policy in this document?

platform/engineering/design-docs/separate-content-build.md
cvalarida: Just to clean up the loose ends, I'm marking this as resolved for now. Feel free to reopen if needed, though.

platform/engineering/design-docs/separate-content-build.md
cvalarida: > new deployments will be available to browsers only after the the cache expiration time

Is that...true? With `no-cache`, the browser will always make the request (and get back a `304 Not Modified` if it hasn't been modified since cached, I believe), so browsers should always get the latest deployments. Unless I'm misunderstanding something, anyhow.

I'll add the other trade-offs, though. :+1:

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,477 @@
...
+code is in production before deploying the landing page for it.
+
+
+### Work Estimates
+The following estimates vary greatly depending on who's doing the work.
+
+1. Write a standalone script that can be pointed to the build output directory
+   to check for broken links
+    - **Estimate:** 2 hours - 2 days
+1. Trigger this job in Jenkins (Nomad? Circle?) after a `master` build succeeds
```
rianfowler, line 398: Have we talked through how this would work with devops? My rough understanding is that we would hit the [release API](https://developer.github.com/v3/repos/releases/) to trigger a job in Nomad but the outstanding question was how do we pass the build JS / CSS from Circle to Nomad. 

I don't think it's as simple as compressing them into an archive and putting them somewhere- we might have to do something with docker.

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,477 @@
...
+    - Use
+      [`html-webpack-plugin`](https://github.com/jantimon/html-webpack-plugin)
+      to generate temporary application landing pages
+      - Create a template which we can use with the above plugin
+      - This should only be used in `localhost` builds until `vets-website` is
+        no longer responsible for deploying the content
+    - Write a script to run content and application builds while they're both in
+      `vets-website`
+    - **Estimate:** 1 - 3 days
+1. Modify the `createReactPages` step to use [the new JSON
```
rianfowler, line 413: does the CMS team need to sign off on this approach?

cvalarida: This part is essentially the same as it has been. The part where we'll want alignment is the bit where we add the React landing pages as pages in the CMS. I evidently haven't outlined that bit very well, so once I do, I'll ask somebody from that team to check it out. :+1:

platform/engineering/design-docs/separate-content-build.md
cvalarida: I'm not really following on this one. What "new application release" section?

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,477 @@
...
+      - Does not trigger Webpack to build the JS and CSS bundles
+      - Instead,
+        [fetches](https://github.com/department-of-veterans-affairs/vets-website/blob/006185b7c40fe85a8b219ae99f3aefb3d8bf0e09/src/site/stages/build/plugins/configure-assets.js#L22-L23)
+        the assets [from S3](https://github.com/department-of-veterans-affairs/vets-website/blob/006185b7c40fe85a8b219ae99f3aefb3d8bf0e09/src/site/stages/build/plugins/download-assets.js#L111)
+        - Process is triggered [from this Jenkins config](https://github.com/department-of-veterans-affairs/vets-website/blob/006185b7c40fe85a8b219ae99f3aefb3d8bf0e09/Jenkinsfile.content#L32)
+    - Does **NOT** run any content validation such as accessibility tests or
+      broken link checker
+2. [Full deploy](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started/workflow/deploy/#full-deploy-of-vagov-client-app)
+    - For `vets-website` code changes
+    - Creates a new release and deploys it
```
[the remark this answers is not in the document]
cvalarida: Turns out I maybe don't know what I'm talking about here. The Jenkins code is gnarly and I'm having trouble figuring out what this means. Somewhere along the way, a GitHub release is triggered and I was pretty sure Jenkins grabs that release (compressed `vets-website` at the latest `master` commit...or....I don't know, I guess) and does its thing.

Mayhap the likes of @nathanhruby or @wyattwalter can clarify what exactly is going on? I think I'm more confused now than when I started trying to figure this out. :disappointed:

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -0,0 +1,477 @@
...
+The **[CI pipeline](#ci-pipeline)** runs this build script (among other things)
+whenever a commit is pushed to a branch in GitHub. On a branch, this build must
+be successful **before the branch can be merged.**
+
+The output of this build are **[deployed](#deployment) to and hosted from an AWS
+S3 bucket**. Applications are deployed on a weekdaily schedule. New content is
+deployed by a manual trigger from the CMS and by the weekdaily application
+deployment.
+
+The entangling of the content and application builds have [led to
```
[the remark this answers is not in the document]
cvalarida: I changed a previous sentence from
> These builds are currently **orchestrated by a single build script.**

to

> These two distinct tasks are currently **entangled in a single build script.**

Does that resolve this?

platform/engineering/design-docs/separate-content-build.md
cvalarida: I added more about this in the `React application landing pages` section.

platform/engineering/design-docs/separate-content-build.md
```diff
@@ -62,9 +62,9 @@
 The objective here is to define an approach to isolate the concerns of the
 content build from the front end application build and vice versa.
 
-This design document is intended for front end engineers and DevOps engineers on
-the Veteran-facing Services Platform (VSP) and other Veteran-Facing Services
-(VFS) teams. Most directly affected is the Content Management System (CMS) team.
+This design document is intended for front end and DevOps engineers on the
+Veteran-facing Services Platform (VSP) and Veteran-Facing Services (VFS) teams.
+The most directly affected group is the Content Management System (CMS) team.
```
gunsch: This statement seems at odds with not having a CMS team representative as approver, which I understood was because this has _minimal_ impact on CMS team. Can you clarify?

platform/engineering/design-docs/2020-04-03-separate-content-build.md
```diff
@@ -0,0 +1,486 @@
+# Separate Content Build
+
+**Author(s):** Christopher Valarida  
+**Last Updated:** March 16 2020  
+**Status:** Draft | **In Review** | Approved  
+**Approvers:** Dror Matalon, Andrew Gunsch, Rian Fowler
```
gunsch, line 6:
```suggestion
**Approvers:** Dror Matalon, Andrew Gunsch [x], Rian Fowler  
```

I'm OK approving here, I think at a high level most of the major questions are answered and this direction seems solid. Obviously finish getting Rian's approval still, and maybe also CMS team? (see comment below).

Ping Dror and ask if he wants to review this or not; he mentioned to me today he's trying to be less in the weeds on everything now that he's back, so he may opt to pass on this.

platform/engineering/design-docs/separate-content-build.md
gunsch: yeah, I think so --- at least as a proposal. no need to articulate the full explicit policy here or pretend that this design doc is a policy doc, but at least saying "we acknowledge this risk, and this is how we intend to address it" would be worth including.

platform/engineering/design-docs/separate-content-build.md
cvalarida: Yeah, I think this is maybe a vestige of a past iteration. I'd be happy to put a reviewer on here from the CMS team, but the impact should be fairly minimal. :thinking: Maybe "most affected" meant "this change will require some work from them, which is more than any other VFS team." Not sure what I was talking about. :confused: 

At any rate, [I've talked with a few folks](https://dsva.slack.com/archives/CT4GZBM8F/p1585783218385800) from their team about this briefly, and invited them to review and comment on this PR. I outlined what I expect the impact to them will be.

platform/engineering/design-docs/separate-content-build.md
cvalarida: I'm breaking this out into a separate design document. For now, we won't be changing this.

drorva:
Overall this is such a good plan that I'm surprised that the two build were integrated in the first place :-).
# CMS Build Speed
One of my main concerns is the time for content build and deploy the CMS. The cycle of:
* make changes to the CMS
* deploy to see these to dev or stagin
* see what needs to be changed
Can't be very long or else it'll be a huge pain.
What is the current time?
It look like we're doing a full deploy of all pages every time. This will just make the deploy bigger and slower. If this is indeed a problem, we can look at just uploading new files. This should be pretty easy using rsync as part of the pipeline or just use https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html. Probably use 

Also might consider periodic deploys on file changes (once or twice an hour).

# Cache busting
This sounds like an unnecessary hack. Why not use the standard browser caching tools. If we know we only deploy to production once a day, we set the expiration accordingly. Or is this just for testing? If it's just for testing, I'm fine with it.

cvalarida:
Thanks, @drorva! :smile: 

## CMS Build Speed
>Can't be very long or else it'll be a huge pain.

You're absolutely right. This is one of the problems the CMS Super Team™ was put together to find a solution to. As part of that effort, we _did_ come up with a few solutions, but those are separate from the plan this document outlines.

The change here is to remove `vets-website` from the content build and deploy (and vice versa). While the speed concern wasn't the impetus for this change, separating the builds _will_ shave off a few seconds during the content deploy since it won't have to fetch the `vets-website` assets. Further improvements are being made by the CMS team right now to speed the process up, I believe.

My goal with this phase is more to separate the responsibilities of the two builds than it is to change any process for now. I expect that will come later.

## Cache Busting
>Why not use the standard browser caching tools.

Which browser caching tools do you mean? The approach I'm planning on now is using the `Cache-Control: no-cache` header on the application assets.

We _typically_ deploy once per weekday, but do push out critical bug fixes outside of those regularly-scheduled deploys as needed. If we set the cache to 24 hours, I believe those off-schedule bug fixes wouldn't make it to the browser if the user already loaded and cached the buggy application's assets.

My plan for now is to experiment with the `Cache-Control` header and use SpeedCurve to monitor the render time in React applications to see if it increases significantly enough for us to figure out another approach.

That said, I'll need to verify with @jhonnyoddball that we can see that particular metric in SpeedCurve and set up a performance budget for it if we don't already have one in place.

If you have other ideas for this, please let me know! I do love me a good experiment to determine the "right" approach. :smile:

drorva:
Makes sense that build speed is out of scope for this. 

I was reacting to "cache busting currently relies on using new file names in static content on every deploy" but it seems like that was the old approach. 
I wouldn't use "Cache-Control: no-cache" because that means that on every page we fetch everything from the server which will slow performance. Using a short appropriate time, an hour maybe, would be much better.

gunsch:
> I wouldn't use "Cache-Control: no-cache" because that means that on every page we fetch everything from the server which will slow performance. Using a short appropriate time, an hour maybe, would be much better.

@drorva I'm the one who recommended starting with `no-cache`. This does still involve browsers checking with the server, but it doesn't involve re-downloading content --- E-Tags are still used, S3 will still report a 304, and the browser won't re-download the file.

I'm explicitly trying to avoid a situation in which we need to roll out an emergency fix/update to things, and have to tell the stakeholder "it's rolled out, but now we have to wait an hour for browser caches to clear, during which things are still broken." Plus it avoids any failure modes about caching between different resources (JS/CSS) getting out of sync somehow, etc.

If this ends up being a material performance regression, I'm happy to re-evaluate. I suspect this will be trivially noticeable at most.

drorva:
On second thought, it probably won't fetch the data again, it'll send the request and get a 304 Not Modified. So we're not downloading on every page, we're requesting on every page. Frankly, my knowledge of best practices in this arena is dated, and caching is quite complicated. So I'll let you folks do the research and come up with the best approach :-).

drorva:
btw, have you heard
`There are only two hard things in Computer Science: cache invalidation and naming things.`
Which is clearly wrong, since he missed race conditions :-)

platform/engineering/design-docs/separate-content-build.md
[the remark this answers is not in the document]
rianfowler: Not really. Where does the design document point out that the output of webpack is used as an input of the static content build and explain why that is a problem?

platform/engineering/design-docs/2020-04-03-separate-content-build.md
```diff
@@ -0,0 +1,486 @@
...
+the current build script's output. To test that this is working properly we can:
+
+1. Run the script pre-separation (checkout `master` for this)
+1. Rename the `vets-website/build/localhost/` directory
+1. Run the post-separation build script
+1. Check that the all same files are there
+    - We can use `find` and `cmp` for this as needed
+    
+### Logging
+**Question:** What _should_ it log? Anything? Is this a Jenkins / Nomad /
```
rianfowler, line 311: Maybe work with @bkjohnson  on this. We could document the specific way we're going to break these tasks up in CI using Circle's nomenclature like 

- the build process will be broken into multiple workflows 
- the workflows will notify success on github 

etc etc 

Screen Shot 2020-04-07 at 11 01 25

The nomad piece is still a question. We know that we will need to build some sort of artifact in Circle that can be passed to Nomad process that's triggered via the github deployment API.

platform/engineering/design-docs/2020-04-03-separate-content-build.md
```diff
@@ -0,0 +1,486 @@
...
+CircleCI thing?
+
+### Debugging
```
rianfowler, line 314: We should mention the logs that will be available to troubleshoot branch and deployment builds. The primary UI is going to be in Circle- we're going to do as much of the process as we can there.

cvalarida:
Heh, this design document had to consider cache busting _and_ race conditions (kind of), but not really naming things. Best two out of three, eh? :sweat_smile: 

Is there anything else that we need to address in here before getting your stamp of approval, @drorva?

platform/engineering/design-docs/2020-04-03-separate-content-build.md
```diff
@@ -0,0 +1,486 @@
...
+
+### Caveats
+- The content validation doesn't happen in the build job
+  - The application CI runs tests, but the content CI does not
+  - This is because it takes a long time and content editors are used to seeing
+    their changes in production quickly
+  - Validation instead happens in a separate job
+
+### Security Concerns
+There are no new security concerns with a separated build process.
```
rianfowler, line 347: some ideas for this section: 

- We haven't resolved an [open question](https://github.com/department-of-veterans-affairs/va.gov-team/pull/6138#pullrequestreview-369232868) about bot tokens in circle yet. 
- Circle runs outside of the VA network
- the website build relies on public npm (existing concern)

rianfowler (commented):
I think we're getting close to this being done.

platform/engineering/design-docs/2020-04-03-separate-content-build.md
cvalarida: Should I be documenting existing concerns or new concerns that this process introduces?

platform/engineering/design-docs/2020-04-03-separate-content-build.md
```diff
@@ -0,0 +1,486 @@
+# Separate Content Build
+
+**Author(s):** Christopher Valarida  
+**Last Updated:** March 16 2020  
+**Status:** Draft | **In Review** | Approved  
+**Approvers:** Dror Matalon, Andrew Gunsch [x], Rian Fowler
```
rianfowler, line 6:
```suggestion
**Approvers:** Dror Matalon, Andrew Gunsch [x], Rian Fowler [x]
```

rianfowler (approved):

gunsch (approved):

platform/engineering/design-docs/2020-04-03-separate-content-build.md
```diff
@@ -0,0 +1,486 @@
...
+assets in the `/applications/` directory.
+
+![Coordinating deployments to S3](images/separated-content-build/coordinating-deployments-to-s3.png)
+
+To ensure that one build won't override files in the other, we'll have to
+1. Update
+   [`vets-website-deploy.sh`](https://github.com/department-of-veterans-affairs/devops/blob/2c63e6d62fa5ad2b6b87478e30d0a7bc2820c252/ansible/deployment/roles/deploy-vets-website/files/vets-website-deploy.sh)
+   to copy only to the `/applications/` directory
+1. Create a new deploy script for the content deployment which copies everything
+   over _except_ the `/applications/` directory
```
cvalarida, line 222: This part _mostly_ makes sense to me, but I'm not 100% solid on it yet. Am I missing anything? Should I add any clarification here?

platform/engineering/design-docs/2020-04-03-separate-content-build.md
```diff
@@ -0,0 +1,486 @@
...
+- The E2E test script in CI will start a process to serve these pages so the
+  tests can run a browser and access the applications
+- **Important note:** These HTML files will not be served in production; the
+  landing pages for the applications come from the content build
+  - To ensure these pages aren't made in production, the deploy script will copy
+    only the files in `/applications/`
+
+#### Miscellaneous
+- We need a separate content validation job that runs once a day on a schedule
+  - **Question:** How will we validate links to `vets-website` assets?
```
cvalarida, line 241: I think the answer is that we'll have a Docker container in Nomad which can `s3 sync` down all the files in production and run all content validation checks on that local copy.

Does this sound feasible, @wyattwalter ?

Will we need an interim solution until Nomad is ready to be used?

platform/engineering/design-docs/2020-04-03-separate-content-build.md
cvalarida: Is it enough to say that we can see the logs of what happened by looking in CircleCI and Nomad?

cvalarida (commented):
@wyattwalter I want to add a bit about running the application build in CircleCI (since we don't want it behind the TIC for this step, I think, right?) and bundling it up in a Docker container to ship to Nomad / Jenkins for deployment. There's enough in there that's pretty vague to me that I'm not really comfortable writing up an approach yet. How would you suggest we go about that?

1. Can we initiate a deploy from CircleCI?
    - @bkjohnson ?
1. Once it's built in Circle, how can we use the result and put it in a Docker container?
1. What would we do with Docker container to get it to Nomad?
    - Would it be stored as a GitHub release, mayhap?
      - Not sure how that plays into this
1. Once the container is in Nomad, my understanding is that we can have it run a script that does the `s3 sync` magic like we're doing now—does that assumption hold water?
1. Do we need an interim solution in Jenkins before Nomad is available for use?

drorva (approved):

cvalarida:
Merging this in now. Thanks for the approvals! :raised_hands: 

I'll make the additional changes in a separate PR so we can start referencing this one in `master`.

platform/engineering/design-docs/2020-04-09-separate-content-build.md
```diff
@@ -0,0 +1,498 @@
...
+The output of the application build **currently** live in the `/generated/`
+directory. To support the transition, the **new** application build will put its
+assets in the `/applications/` directory.
+
+![Coordinating deployments to S3](images/separated-content-build/coordinating-deployments-to-s3.png)
+
+To ensure that one build won't override files in the other, we'll have to
+1. Update
+   [`vets-website-deploy.sh`](https://github.com/department-of-veterans-affairs/devops/blob/2c63e6d62fa5ad2b6b87478e30d0a7bc2820c252/ansible/deployment/roles/deploy-vets-website/files/vets-website-deploy.sh)
+   to copy only to the `/applications/` directory
```
wyattwalter, line 219: This could be a separate bucket, if it's truly this clean that'd be super easy to separate and then we don't have to deal with as much complexity.

Did we get buy-in from IA folks on what this path would be? That's pretty generic and may already be in use.

platform/engineering/design-docs/2020-04-09-separate-content-build.md
```diff
@@ -0,0 +1,498 @@
...
+      content build
+- The E2E test script in CI will start a process to serve these pages so the
+  tests can run a browser and access the applications
+- **Important note:** These HTML files will not be served in production; the
+  landing pages for the applications come from the content build
+  - To ensure these pages aren't made in production, the deploy script will copy
+    only the files in `/applications/`
+
+#### Miscellaneous
+- Separating the content and application builds will require a standalone content validation job that runs outside of the CI pipeline- probably once a day on a schedule
```
wyattwalter, line 239: We'd want to run this before a deploy of the content, right? Otherwise we could deploy a set of HTML that references non-existence files. That's what this is trying to detect, right?

platform/engineering/design-docs/2020-04-09-separate-content-build.md
```diff
@@ -0,0 +1,498 @@
...
+- Application name
+- Whether the file should be made available in production
+
+When we do this, we will update the `manifest.json` documentation and remove
+unused properties from all the files.
+
+**Phase 2: Long-term solution**  
+All application landing pages will be created in the CMS. The page's published
+state will determine whether the application should be made available in
+production.
```
wyattwalter, line 266: Does Flipper play into this decision point at all?

platform/engineering/design-docs/2020-04-09-separate-content-build.md
```diff
@@ -0,0 +1,498 @@
...
+- To manage dependencies independently
+  - All dependencies in the content build repo are not client-facing
+    - There _may_ be exceptions later
+  - It will be harder for an application to use be able to use a dependency that
+    wasn't intended to be client-facing
+
+#### Multiple S3 Buckets
+There are a few challenges that multiple buckets present.
+1. ATO
+    - There's a lot of paperwork
```
wyattwalter, line 483: Nah, we are free to make this change.

platform/engineering/design-docs/2020-04-09-separate-content-build.md
cvalarida: > This could be a separate bucket

I talked with @nathanhruby and it sounded like it'd be more of a challenge for us to go that route. I'm honestly okay with whatever works best. I like the idea of separating the buckets, but if that's not as straightforward as my mental model makes it sound, I think it's probably find to keep it all in one bucket.

> Did we get buy-in from IA folks on what this path would be?

No. Great point! It's not a path that's in use currently, but when we go to implement this change, I'll be sure we check in with IA folks on this and get the green light first. :+1:

platform/engineering/design-docs/2020-04-09-separate-content-build.md
```diff
@@ -0,0 +1,498 @@
...
+1. Namespacing
+    - To use S3 to serve a static site, the bucket needs to be named after the
+      namespace, e.g. `www.VA.gov`
+    - We could maybe get around this with using a subdomain like
+      `javascript.VA.gov` or something, but again, paperwork and likely more
+      maintenance burden
```
wyattwalter, line 489: So actually, this is interesting.. I was originally thinking above that we could do multiple buckets and make the mapping happen in the revproxy. However, we don't even need that complexity.

Currently, the JS/CSS/etc files are served from a public S3 bucket. You'll notice that it doesn't have va.gov in the name at all. It's just a regular old S3 bucket and files are served via S3's website hosting feature. There'd be no trickery needed here and no additional paperwork. We're clear to serve this type of static content from a memo from the CTO. Technically we can also serve from a CDN as well, but we haven't gone through the vendor selection process there.

wyattwalter:
I know I'm late, but added a few comments. I love this approach overall, I think we can simplify it even further with the bucket stuff.

platform/engineering/design-docs/2020-04-09-separate-content-build.md
wyattwalter: The prod assets are currently loaded from https://prod-va-gov-assets.s3-us-gov-west-1.amazonaws.com. There's no `.gov` involved 😄

platform/engineering/design-docs/2020-04-09-separate-content-build.md
cvalarida: Ideally, we'd run this before a content deploy, yes. Ideally, it'd block a deploy so we wouldn't have "broken" content make it to production. That's the ideal veteran experience, I'd say.

But because of reasons, I think this isn't feasible. Because we'll have hundreds of content editors, we can't block the deploy of _all_ content if _one_ thing (or even some higher threshold) is bad. We've talked at length about this with the CMS team, and the priority we settled on was build and deploy speed, which means all validation has to be taken out of that process, unfortunately.

So this is the next best approach, I think. I hope.

For what it's worth, the current content-only deploy does no content validation. That's all done in the full build. That we can push out "broken" content won't be new with this approach.

platform/engineering/design-docs/2020-04-09-separate-content-build.md
cvalarida: It doesn't right now. We may change this later, but :man_shrugging:

platform/engineering/design-docs/2020-04-09-separate-content-build.md
cvalarida: Oh, interesting. I was maybe leading @nathanhruby down a path of misunderstanding when I had this conversation with him.

platform/engineering/design-docs/2020-04-09-separate-content-build.md
cvalarida: Interesting! I'd love to get some time on your calendar in the next few weeks to talk about this before we make the change to the deployments.

wyattwalter:
This is the link I meant to put in for the GitHub deployments stuff. Has a simple diagram that sort of explains it.

https://developer.github.com/v3/repos/deployments/

PR #7192, opened 2020-03-24
7171 getting started with forms

short000:
Adding some developer content to Shawna's "Getting Started With Forms". It also points to a new directory I created under teams/vsa/engineering/forms-system for capturing our comments, potential issues. I have two separate markdown files for that, so please review those as well. I believe this will help us bring more concrete things to VSP to inform their backlog, improve the overall forms documentation (my understanding is that the forms framework makes up anywhere from 60 to 80% of the front-end work).


ndsprinkle (approved):

cohnjesse (approved):
This looks good, this doc is obviously meant for designers so I think the Developer section has plenty there for the context of this document. We will continue to get the best practices documentation put together as well as your common issues doc as a foundation for more comprehensive "developer focused" documentation. I think it definitely makes sense to have the developer focused documentation live in that separate folder structure and not be mixed in with the this design focused document ( unless this document is meant to be a door into the developer focused documentation )

teams/vsa/design/getting-started-with-va.gov-forms.md
```diff
@@ -164,8 +167,15 @@ This document will not attempt to rewrite those documents, however, we will prov
    * Hierarchy/labelling
 
 ### Developers
-* Make sure you read up on and understand form builder [https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/forms](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/forms)/ and/or [watch the quick demo Chris V. gave to the design team](https://zoom.us/rec/share/x5FpP7XC1DJOYZ3d933GUagdMInqeaa82ilM-KcLzUxo_-q1CWuJcUJVwzaPXTp_) in Feb. 2020 (pw: vsadesign)
-   * Fill out a few forms on staging.va.gov to get a feel for the capabilities of the forms system
+* Make sure you read up on and understand form builder:
+   *  Fill out a few forms on staging.va.gov to get a feel for the capabilities of the forms system.
+   *  [Watch the zoom video of a quick demo Chris Valarida gave to the design team](https://zoom.us/rec/share/x5FpP7XC1DJOYZ3d933GUagdMInqeaa82ilM-KcLzUxo_-q1CWuJcUJVwzaPXTp_) in Feb. 2020 (pw: vsadesign)
```
short000: Perhaps rather than pointing to zoom (which we do not control), we should download the .mp4 somewhere, make a link to that instead. That way we reduce the risk of losing it, plus we can get rid of the password and control migration of the video to somewhere else later.

cohnjesse: I agree, I think adding it to Git LFS makes sense since this will live in a Github doc

mojaray2k (approved):

sshein:
@cohnjesse this document is not necessarily "meant only for designers" it's more meant as a "getting started" doc for all disciplines. So basically, if any discipline is tasked with creating a form on VA.gov, they should first read this document to get a feel for how things work and the standards in place. Then, they can jump off to more detailed documentation, as Paul mentioned. In terms of the Zoom URL, that video is pretty informal. I'm wondering if the dev team would like to have Chris give a more formal demo that's recorded, and put in a repository we control?

sshein (approved):

teams/vsa/design/getting-started-with-va.gov-forms.md
short000: Great, I added .mp4 to the repo's .gitattributes just now. I saw that @tlei123 has earlier added .m4v without much fanfare . You'll still have to to a -f with git add (force) because I left .mp4 alone in the .gitignore just to be safe.

short000:
Developer contributions to "Getting started with VA.gov Forms";
Added **teams/vsa/engineering/forms-system** as a place to include work-in-progress information regarding forms.

PR #7542, opened 2020-04-02
Update vsp-collaboration-cycle.md

jenstrickland:
Edited the request a Full 508 & Accessibility audit language so that it would be easier for the PM to know where the link is, where _click [here]()_ was easily overlooked.


meganhkelley:
For some reason I'm only seeing the initial changes you made? (Only to the full 508 audit collaboration point, and it does not mention that you can select from the ZenHub dropdown.) Am I missing something?

jenstrickland:
> For some reason I'm only seeing the initial changes you made? (Only to the full 508 audit collaboration point, and it does not mention that you can select from the ZenHub dropdown.) Am I missing something?

I had submitted the PR before your Slack response. I thought I'd fix it when I saw the PR review come through with the request.

jenstrickland:
> For some reason I'm only seeing the initial changes you made? (Only to the full 508 audit collaboration point, and it does not mention that you can select from the ZenHub dropdown.) Am I missing something?

The adjustment is in the PR now. I ended up creating bullet points to make it easy to digest.

meganhkelley:
Great! Maybe add "Or," to the beginning of the second bullet? The "Two options" note gets a bit lost in the header. Otherwise looks great, go for it 😎

meganhkelley (approved):

PR #7690, opened 2020-04-06
Update PII purge docs

billfienberg:
- Add sub-bullet to log additional info about the sensitive content that needs to be removed
- Add steps about finding/deleting references

### DIFF
--- a/platform/engineering/removing-sensitive-data-from-a-repository.md
+++ b/platform/engineering/removing-sensitive-data-from-a-repository.md
@@ -7,6 +7,7 @@ _Last updated Q1 2020 by VSP Content & IA team_
    - Add the following information to the [purge log](https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/blob/master/platform/engineering/purge-log.md)
      - Origin
        - Initial commit date
+       - Nature of sensitive info
        - Author
        - Initial commit SHA (SHA of commit where sensitive info was first introduced to repo)
        - Initial commit branch (`master` or something else)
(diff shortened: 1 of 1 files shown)


meganhkelley:
Thanks @billfienberg ! Is there anything in the updated purge log template that we think could be deleted? We should keep that to crucial info only so that we're not overdoing it on the burden of filling it out. (For example, we should definitely keep the "commit SHA where sensitive info was merged" bc we need that later, but do we need to identify the author of the sensitive info? The identifier?)

billfienberg:
> Thanks @billfienberg ! Is there anything in the updated purge log template that we think could be deleted? We should keep that to crucial info only so that we're not overdoing it on the burden of filling it out. (For example, we should definitely keep the "commit SHA where sensitive info was merged" bc we need that later, but do we need to identify the author of the sensitive info? The identifier?)

@meganhkelley, we could theoretically delete everything except the commit SHA. I added the extra info so that people could see a comprehensive history of an issue in a single place. Since the purge process involved multiple steps spread out over a long period of time (identification/deletion/sanitization/etc), I found myself repeatedly searching Slack and/or GitHub trying to track down specific pieces of info. I think logging all that info as it becomes available would make each successive step in the process easier. In my ideal world, a log entry would read like a timeline that could be linked and/or copy-pasted. 

I agree that it is a non-trivial amount of information, but my thought process was that this is something that shouldn't be happening often, so the total burden should be minimal. Additionally, I think the sensitive nature merits thorough documentation. 

That being said, I have no objection to deleting whatever you recommend.

PR #7768, opened 2020-04-08
Moves 10-minute downtime fallback into Option B

gunsch:
Proposed update for readability flow. To me, this note is a part of Option B (if you're running downtime and it runs too long, abort and go to option A!) rather than something that's part of the initial decision-making process about Option A vs Option B.


meganhkelley (approved):
LGTM! 🚀

PR #7962, opened 2020-04-13
Merge security review process doc into collab cycle

gunsch:
* Deletes old security review doc
* Removes a few links to it that shouldn't have been there in the first place
* Adds notes to collab cycle doc about how to handle security review issue


meganhkelley (approved):
Generally looks good — you've certainly updated some files in this batch that seem like they could use a more robust overhaul. 

In the VSA launch checklist, we'll need to remember to relink the ATO docs once they are updated with the new ATO process (we are incorporating this into the Collab Cycle in V2).

PR #8073, opened 2020-04-16
[Design Document] Integration Testing with CDCT

U-DON:


U-DON:
TODO:
- Clarify development workflow vs CI pipeline
- Get more detailed about passing contracts between FE and BE in both scenarios (development and CI), [alternative approaches to Pact Broker](https://docs.pact.io/getting_started/sharing_pacts#alternative-approaches), and how to use Pact Broker commands in both (`pact-broker can-i-deploy`, etc.)
- Talk about value of unblocking FE devs from waiting for BE devs to create stubbed API?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,275 @@
...
+- When run alone, contract tests don't simulate real users. However, we can use the generated pacts to create stubbed responses in the E2E tests. This would also address the existing issue of unvalidated and untrustworthy stubbed responses in those tests. There is already an official stub server that Pact has documented. Alternatively, in the spirit of consolidating dependencies, it might be simpler to leverage the stubbing capabilities of the E2E testing framework. Either way, using the the pacts grants the advantage of having stubbed responses that have gone through validation with the actual running API. The validation may occur in separate steps, but those steps come together to make the contract-supplemented E2E tests as trustworthy as full E2E tests.
```
rianfowler, line 258: So the idea here is that we could use the end output of this contract test in the functional FE tests?

rianfowler (commented):
this is great- have a couple of questions.

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,275 @@
...
+     ```
+     const mockApi = new Pact({
+       port: 3000,
+       consumer,
+       provider,
+       spec: 2,
+     });
+
+     ```
+   - Test each request in its own `describe` block. Further wrap in `context` blocks if necessary to test the same request when there are different BE states.
```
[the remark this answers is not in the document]
U-DON: I suppose it's not exactly *necessary*, but I would recommend doing context blocks that correspond to the state you want the provider to set up for that particular test case. To use search as an example:

```
contractTest('VA.gov Search', 'VA.gov API', mockApi => {
  describe('GET /search', () => {
    // The state for this interaction would be something like 'a single matching result exists'.
    context('with a single result', () => {
      it('responds with success', async () => { ... });
    });
  });
});
```

`vets-api` will depend on an external service for pretty much everything, so I anticipate most interactions will have to at minimum define a state to explicitly make the assumption that the external service being called is up and running.

I guess in the case that the state is just "this external service is up", the context block isn't necessary. But it's good for readability when you're testing different scenarios where the same request expects different results.

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,275 @@
...
+         },
+       },
+     };
+     ```
+     - **The `state` attribute will be used as a key for the API to set up the corresponding state when verifying the contract.** For example, this might entail spinning up a mock third-party service, seeding the database, or simulating a logged in session.
+     - **The body does not need to be comprehensive.** At a minimum, the expected response only needs to define the attributes that are relevant to the test. This means that any extra, unused, or unexpected properties of the response will not break the verification later.
+   - Invoke the relevant methods that make the request being tested.
+   - Make assertions on the expected response.
+
+   **To reiterate, the tests should focus on the components or functions that make requests to the API.** It's not meaningful to simply test a fetch call to an endpoint.
```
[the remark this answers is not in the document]
U-DON: I had some trouble myself figuring out what layer of abstraction to test the request when I was trying it on HCA as an example. There isn't going to be a blanket statement for every app of what component to test, but the general idea is that these are integration tests, so they should cover a couple of layers to see if multiple calls or components are working together as expected.

But for forms, which follow a pattern, we should figure out what component to recommend testing from, or better yet, provide a utility a la "form tester" that contract tests the forms from a common point.

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,275 @@
...
+           ...
+         })
+       })
+     })
+   });
+   ```
+2. The contract tests will run in CI when FE changes are pushed.
+   - **If they pass, they generate pacts in the form of JSON files in the `pacts` directory.** JSON files are generated for each pairing of consumer (the FE application) and provider (`vets-api`).
+   - If they fail, it's possible the request wasn't picked up by the mock server because request defined in the interaction didn't match the request that was actually made. It's also possible the component or function that made the request failed an assertion.
+3. **The CI job invokes a script to publish the pacts to a broker.** The command to publish should require a version to label the consumer and optionally accept tags. The published pacts should be versioned to the current commit hash of `vets-website`. For builds to be deployed, the pacts should also be tagged with the appropriate environment. *Tagging is important to ensure that we verify only the pact associated with the build that is getting deployed and not pacts generated from other branches.*
```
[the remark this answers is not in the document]
U-DON: I see this happening even before a PR is merged. It should happen as one of the build jobs.

We don't want PRs to merge that could introduce breaking changes because the `vets-api` master branch doesn't accommodate the changes yet. We also don't want to stop the `vets-website` master from deploying (to staging or production) when an app has a broken contract, because it would stop other apps from deploying their changes when they haven't broken their contracts.

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,275 @@
...
+
+   This assumes that the broker is accessible from CI. There may be some work involved in configuring the the broker to allow access depending on where it's hosted.
+4. Publishing the pacts should trigger a webhook in the broker that invokes a Rake task in `vets-api`:
+   ```
+   bundle exec rake pact:verify
+   ```
+   This task pulls all relevant pacts from the broker and verifies that the expected responses match the actual responses when running the requests through a running instance of the API.
```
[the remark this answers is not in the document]
U-DON: As the API tests each interaction, it reads the `state` of that interaction to set up the state before reading in the request. So in `vets-api`, there will be a `provider_states.rb` that sets up the states like this:

```
Pact.provider_states_for 'HCA' do
  provider_state 'enrollment service is up' do
    set_up do
      # Stub the response from the enrollment service here.
    end

    tear_down do
      # Any tear down steps to clean up the provider state
    end
  end
end
```

Notice the block for each state that starts off like: `provider_state '<the state description>' do`.

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
U-DON: That's correct. I should update the document to talk about why we don't want the functional tests to generate the contracts (Pact has an explicit answer in the FAQ about that), but I see the functional tests setting up the stubs using the end output from the contract tests. That end output will a set of mock responses that we can trust because they've passed validation from the API.

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+
+There are some problems with relying only on the current set of E2E tests as integration tests.
+- **The E2E tests don't actually incorporate the API.** The stubbed responses are not validated and don't necessarily reflect reality, so the tests don't give us confidence that we're testing the right thing.
+- FE developers are entirely responsible for coming up with stubbed API responses, but aren't accountable for *validating those responses*.
+- Without involving the BE developers maintaining the endpoints, FE developers are **prone to making invalid assumptions** about systems that they don't work on themselves.
+
+There has been at least one incident that revealed the ineffectiveness of the current test strategy in detecting breaking changes from the API. In this case, an API endpoint was modified independently of the form that invoked it, causing the form to break. Tests did not catch this as they were assuming an outdated expected response.
```
gunsch, line 36: do we have a ticket/postmortem we can link to here about this incident?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+> **Contract testing is a technique for testing an integration point by checking each application in isolation to ensure the messages it sends or receives conform to a shared understanding that is documented in a "contract".**
+
+> For applications that communicate via HTTP, these "messages" would be the HTTP request and response, and for an application that used queues, this would be the message that goes on the queue.
+
+The contract involves two parties, the **consumer** and **provider**. For our purposes, the FE application is the consumer consuming an endpoint from the provider, which is the API.
+
+In a consumer-driven context, the consumer generates the contract and the provider verifies it.
+
+The process of using Pact can be broken into multiple steps:
+1. Consumers run unit tests that test their request and response interactions with the provider's endpoints.
```
gunsch, line 59: I'm finding this a bit hard to follow. "The FE application is the consumer" above, but then "the consumer generates the contract" and "consumers run unit tests". Are you saying the FE application (which I take to mean "application code in vets-website") is generating the contract code, and running the unit tests? Is "consumer" being used to mean a few different things here?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+2. The tests generate contracts, also referred to as **pacts** in the Pact framework.
+3. The pacts are published to a central location or a broker to be versioned and shared with the provider.
```
gunsch, line 61: Would it be accurate to simplify much of this text to talking about `vets-website` and `vets-api` rather than generic "consumer" and "provider"?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+4. The provider verifies the pacts against an actual running instance of itself and validates that the actual response matches the expected response.
```
gunsch, line 62: again, does "provider" mean `vets-api` itself here, or would this be more accurate to say "a BE engineer verifies the pacts against a running instance of the BE"? how is this instantiated?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+   - If they fail, it's possible the request wasn't picked up by the mock server because request defined in the interaction didn't match the request that was actually made. It's also possible the component or function that made the request failed an assertion.
+
+   The assumption henceforth is that the CI workflow is done in Circle.
+3. **The CI job invokes a script to publish the pacts to a broker.** The command to publish should require a version to label the consumer and optionally accept tags. The published pacts should be versioned to the current commit hash of `vets-website`. For builds to be deployed, the pacts should also be tagged with the appropriate environment. *Tagging is important to ensure that we verify only the pact associated with the build that is getting deployed and not pacts generated from other branches.*
+
+   This assumes that the broker is accessible from CI. There may be some work involved in configuring the the broker to allow access depending on where it's hosted.
```
gunsch, line 143: this seems pretty handwavey for a design doc. can you propose answers to the questions of:
* whether we'll self-host a broker or not, and (if so) where/how in our infrastructure it will live (may require discussion with Ops)
* if we're confident that the broker would be accessible from Circle

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+   The broker maintains a matrix matching the consumer version and provider version.
+6. **The `vets-website` build runs a check to see if it can be deployed.**
+   ```
+   pact-broker can-i-deploy --broker-base-url=BROKER_BASE_URL
+   ```
+   This check is dependent on the published verification results.
+
+   **If verification was successful, the CI pipeline will proceed to deploy. If it failed, the pipeline will stop.**
+
+   We're assuming that the CI pipeline can wait for the asynchronous verification task to run and conditionally handle the output of that task.
```
gunsch, line 169: is this a valid assumption?

if not, is there a fallback plan?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+3. **The CI job invokes a script to publish the pacts to a broker.** The command to publish should require a version to label the consumer and optionally accept tags. The published pacts should be versioned to the current commit hash of `vets-website`. For builds to be deployed, the pacts should also be tagged with the appropriate environment. *Tagging is important to ensure that we verify only the pact associated with the build that is getting deployed and not pacts generated from other branches.*
+
+   This assumes that the broker is accessible from CI. There may be some work involved in configuring the the broker to allow access depending on where it's hosted.
+4. Publishing the pacts should trigger a webhook in the broker that invokes a Rake task in `vets-api`:
+   ```
+   bundle exec rake pact:verify
+   ```
+   This task pulls all relevant pacts from the broker and verifies that the expected responses match the actual responses when running the requests through a running instance of the API. For any states specified in the pacts, the API will set up any matching states already defined in `provider_states.rb`. If there are no matching states, the verification will fail.
```
gunsch, line 148: what running instance of the API? spinning up vets-api locally? running against dev-api.va.gov or something?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+   bundle exec rake pact:verify:at[http://build-box/MyConsumerBuild/latestSuccessful/artifacts/my_consumer-my_provider.json]
+   ```
+   The actual URL should point to the contract that was published from the `vets-website` feature branch. After the `vets-api` feature branch is merged to master, that pact should be able to pass verification. 
+5. **The verification results are published back to the broker** with the version of the API that verified the pacts. Again, we can use the commit hash to version, but this time it's from `vets-api`.
+   The broker maintains a matrix matching the consumer version and provider version.
+6. **The `vets-website` build runs a check to see if it can be deployed.**
+   ```
+   pact-broker can-i-deploy --broker-base-url=BROKER_BASE_URL
+   ```
+   This check is dependent on the published verification results.
```
gunsch, line 165: what happens if the verification task fails and doesn't post results back?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+   
+   [Pact Broker webhooks can be created using the CLI](https://github.com/pact-foundation/pact_broker-client/#create-webhook), whether it's from the Ruby gem or NPM package. **The webhook should just check for `contract_content_changed` because the broker will automatically pass any unmodified contracts that previously passed verification."**
+   
+   While the pacts are being verified, the `vets-website` build can proceed to E2E tests.
+
+   **If the pacts are coming from a feature branch and the API has not been updated to match the new expectations, verification should be expected to fail.** In that case, once the pacts have been published, the BE developers working on updating the API can run the following task during development to verify their changes:
+   ```
+   bundle exec rake pact:verify:at[http://build-box/MyConsumerBuild/latestSuccessful/artifacts/my_consumer-my_provider.json]
```
gunsch, line 156: do you have an example of what URL these contracts might actually be published at?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+### Code Location
+In `vets-website`:
+- Contract tests will be specific to each application and located within that application's `tests` directory: `src/applications/**/tests/contract/*.contract.spec.js`.
+- Interactions can exist in a sibling file to be imported into the tests: `src/applications/**/tests/contract/interactions.js`
+- Helper functions for writing contract tests will be at `src/platform/testing/contract`.
+- Pacts will be generated in the `pacts` directory, but this is not intended to be committed to version control.
+
+In `vets-api`:
+- Configuration for the provider verification will be at `spec/service_consumers/pact_helper.rb`.
+- Provider states will be at `spec/service_consumers/provider_states.rb`.
```
gunsch, line 182: what are "provider states"? this is the first time they're mentioned

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+
+### Testing Plan
+We can start with a trial run of using Pact to contract teste a VA.gov application. This will demonstrate the viability of incorporating contract testing into our process and reveal any unforeseen work that needs to be done.
```
gunsch, line 185: how close have you gotten to this in prototyping so far?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+   ```
+   bundle exec rake pact:verify
+   ```
+   This task pulls all relevant pacts from the broker and verifies that the expected responses match the actual responses when running the requests through a running instance of the API. For any states specified in the pacts, the API will set up any matching states already defined in `provider_states.rb`. If there are no matching states, the verification will fail.
+   
+   [Pact Broker webhooks can be created using the CLI](https://github.com/pact-foundation/pact_broker-client/#create-webhook), whether it's from the Ruby gem or NPM package. **The webhook should just check for `contract_content_changed` because the broker will automatically pass any unmodified contracts that previously passed verification."**
+   
+   While the pacts are being verified, the `vets-website` build can proceed to E2E tests.
+
+   **If the pacts are coming from a feature branch and the API has not been updated to match the new expectations, verification should be expected to fail.** In that case, once the pacts have been published, the BE developers working on updating the API can run the following task during development to verify their changes:
```
gunsch, line 154: I'm confused about this --- here you say verification should be expected to fail, but below you say failure will stop the CI pipeline. does that mean it's expected that FE developers can't submit their test in this case?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+       })
+     })
+   });
+   ```
+2. The contract tests will run in continuous integration (CI) when FE changes are pushed.
+   - **If they pass, they generate pacts in the form of JSON files in the `pacts` directory.** JSON files are generated for each pairing of consumer (the FE application) and provider (`vets-api`).
+   - If they fail, it's possible the request wasn't picked up by the mock server because request defined in the interaction didn't match the request that was actually made. It's also possible the component or function that made the request failed an assertion.
+
+   The assumption henceforth is that the CI workflow is done in Circle.
+3. **The CI job invokes a script to publish the pacts to a broker.** The command to publish should require a version to label the consumer and optionally accept tags. The published pacts should be versioned to the current commit hash of `vets-website`. For builds to be deployed, the pacts should also be tagged with the appropriate environment. *Tagging is important to ensure that we verify only the pact associated with the build that is getting deployed and not pacts generated from other branches.*
```
gunsch, line 141: if pacts are published during presubmit CI runs, how does the Pact broker distinguish between "pacts that are checked in" and "pacts that are being run as part of presubmit"?

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
```diff
@@ -0,0 +1,304 @@
...
+1. Consumers run unit tests that test their request and response interactions with the provider's endpoints.
+2. The tests generate contracts, also referred to as **pacts** in the Pact framework.
+3. The pacts are published to a central location or a broker to be versioned and shared with the provider.
+4. The provider verifies the pacts against an actual running instance of itself and validates that the actual response matches the expected response.
+5. Results of the verification can be published back to the broker if there is a broker involved.
+6. Builds will depend on successful verification results to pass.
+
+## Specifics
+
+### Detailed Design
```
gunsch, line 68: reading this (esp. steps 2-6) feels like it would benefit from some sort of visual diagram (sequence diagram maybe?) showing the control flow: CircleCI, vets-website, pact broker, vets-api, and when each type of test is run (especially since parts of it are async).

gunsch (commented):
nice description of all of this! I've got a lot of questions about things that need more clarification, but this seems on the right track so far.

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
U-DON: From reading Circle's documentation, I think it's supported, but I haven't wrapped my head around it yet.

The fallback plan would be to include running the `vets-api` verification task in the same `vets-website` CI pipeline but in a job following the publishing of the pacts. It can either pull from the broker (just to ensure that it's all there; I think this option is better) or from the artifacts generated by the publishing step in the Circle pipeline (if we want to store the pacts as a build artifact) and would avoid async concerns (and seems a lot easier).

The verification job triggered by `vets-website` pushes wouldn't be under the `vets-api` organization in Circle, but I think we can live with that.

The verifications triggered by `vets-api` pushes would still remain under that org, so I don't see this fallback approach affecting it.

I didn't mention this alternative originally, but it seems worth adding now that I've written it out.

platform/engineering/design-docs/2020-04-16-integration-testing-with-consumer-driven-contract-tests.md
U-DON: The `can-i-deploy` check would fail since there's no passing status on verification, so the code wouldn't be mergable/deployable until the verification task passes. In Circle, we can re-run specific steps, so if anything weird is going on in that verification step, we can resolve any issues with the task and re-run it.

gunsch:
One note per [design doc](https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/platform/engineering/design-docs)

> If a reviewer asks for clarifying technical details, don’t just answer the question inline; update the design doc with the extra details! (Goal: "The design doc has been updated to clarify your question.")

Thanks for the answers to my questions, but they should all be things that are updated in the doc :)

johnpaulashenfelter (commented):
I don't work with the VA/Oddball. Maybe you meant someone else in requesting the review?

PR #8442, opened 2020-04-27
Update external performance monitoring product docs

kfrz:


alexpappasoddball (approved):
Everything that we have discussed is here, it is clear to me what each link in the readme is.  All of the external services documents at first glance appear to have all the key aspects.  This is really solid and a ton of good information!

PR #8615, opened 2020-04-30
VAFS/RJSF Recipes & Best Practices

### LINKED ISSUE #8598: VAFS/RJSF Recipes & Best Practices
1. Organize our scattered notes for VAFS and RJSF into practical how-to's and guidelines. Emphasis should be on "what we have already done" (now) rather than "this is the right way to do it."

2. "**Custom React Component: Simple**": in Slack @cvalarida mentioned that he was able to update form state with an onChange handler within a custom widget. Take that reference to SelectArrayItemsWidget.jsx and use it as a simple how-to before we get into the more complicated scenario below. 

3. "**Custom React Component: Multiple/Complex**": take @bradl3yC 's write-up and add content from code fragments and conversations in Slack so that it is suitable as a stand-alone how-to.

4. Move @cohnjesse 's "**List Loop**," "**Recommend Code Structure**," and "**Using A Constants File**" out of eBenefits and into the main VSA Forms System folder.

short000:
Here is a first stab at pulling together recent Forms system content from various sources (mined from Slack, source code pulled out of GitHub). The discussions are easier to follow when you can see the appropriate code fragments, so I have made an attempt to extract the relevant parts. I'd like VSA developers to review and help me clean up first, then we can bring this over to VSP so that this can inform their planning with Forms documentation efforts. It's all markdown, not anything complex.


kelsonic (approved):
This seems very useful to those who are on-boarding! Thank you! 🙂

Mottie (approved):
👍 Thank you!

micahchiang (approved):

ndsprinkle (approved):

sanlouise (approved):

teams/vsa/engineering/forms-system/readme.md
```diff
@@ -1,17 +1,50 @@
...
+
+## Implementation Patterns
+These real-world implementation patterns are somewhat gnarlky in detail. In some situations we may be using the forms system beyond what it was originally designed for.
+
+- [List Loop](./list_loop.md) - This is needed to build a form that allows a user to input a few items and a separate page for each of those items to enter more information about that respective item.
+
+- [Custom React Component: Simple Case](./custom_component_simple.md) - Update form state with an onChange handler within a custom React component.
 
-### Other (external) useful forms-related guidance:
+- [Custom React Component: Multiple/Complex](./custom_component_multiple.md) - Using a custom React component requires that you pass in the form's schema for that component (map a small slice of form state to the component). In this scenario, we use the same component twice and we pass in a different instance of the schema to each instance of the component. Since each instance only has access to its own individual schema/form state, we are using Redux in order for our instances to see each other's form state.
```
short000, line 19: @bradl3yC and @mr0sari0 how accurate is this statement? If we compare this scenario against the simple case above, does it explain why we have to peek into redux form state?

teams/vsa/engineering/forms-system/custom_component_simple.md
```diff
@@ -0,0 +1,147 @@
+# Updating Form State with an onChange Handler Within a Custom React Component
+
+Based on answers from *Christopher Valarida* from Slack discussions:
+
+- **A field can not trigger a change to another field** \
+*We can! There's an ***updateFormData** available. Put that sucker in the **uiSchema** on a field and changes to it will trigger the callback you pass it. I highly recommend thinking hard about whether you need it first, however. It's a known footgun; if you're not careful, you can make a lot of sneaky bugs for your future self.*
```
short000, line 6: @bradl3yC this wasn't necessary nor sufficient for the complex scenario, right? (Due to "another field" being "another field in another react component?)

teams/vsa/engineering/forms-system/custom_component_multiple.md
```diff
@@ -0,0 +1,138 @@
...
+*Discussion on slack: https://dsva.slack.com/archives/CBU0KDSB1/p1588080662115800*
+
+## Why It's a Hack
+If it isn't obvious by now, rather than treating VSFS/RJSF as a back box, we're having to peek inside the form's internal Redux state (gray box). Since this wasn't thought through by design, you'll run into limitations such as the hidden field scenario we just mentioned.
+
+## Background: The Original Problem
+Here is the original problem we were trying to solve, and conversations we had in Slack related to it. First, here is the write-up from [mr0sari0](https://github.com/mr0sari0):
+
+### Form System Questions/Limitations
+Sharing state between `ui:fields`,
```
short000, line 80: @bradl3yC and @mr0sari0 is this accurate? Are the ui:fields under the same parent, or do they belong to or share state under different react components?

teams/vsa/engineering/forms-system/custom_component_multiple.md
```diff
@@ -0,0 +1,138 @@
...
+
+  ```
+
+## Side-Effect: Extra Component
+You still need to pass in a schema that accounts for the data you want to modify with the form system.  The form system will then add its own component for that data.  So you end up with an extra component.  We need to hide it!  The original way to do this was to leverage a hideIf param that exits in the uiSchema.  There is, however, a gotcha with this.  When you hide a component on the form, it removes its schema thus wiping out its form state.
+
+*[src/platform/forms-system/src/js/state/helpers.js](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/forms-system/src/js/state/helpers.js)*
+```
+export function expandArrayPages(pageList, data) {
+. . .
```
short000, line 57: @bradl3yC  is this really the right line of code?

short000 (commented):
Request for corrections: @mr0sari0 @bradl3yC @cohnjesse

teams/vsa/engineering/forms-system/custom_component_multiple.md
```diff
@@ -0,0 +1,138 @@
...
+     } else if (nextPage.arrayPath !== lastArrayPath && !!arrayPages.length) {
```
short000, line 58: @bradl3yC is this really the right line of code?

teams/vsa/engineering/forms-system/readme.md
```diff
@@ -1,17 +1,50 @@
...
+
+- [Recommended Code Structure](./code_structure.md) - A way to organize your code so that it is easy to maintain.
+
+- [Using a Constants File](./constants.md) - Using a central file for all the variables you will use over and over again.
+
+
+## Implementation Patterns
+These real-world implementation patterns are somewhat gnarlky in detail. In some situations we may be using the forms system beyond what it was originally designed for.
+
+- [List Loop](./list_loop.md) - This is needed to build a form that allows a user to input a few items and a separate page for each of those items to enter more information about that respective item.
```
short000, line 15: @cohnjesse could you cite where in the VSP doc they recommend this? My understanding is that they recommended this pattern but did not provide an example, so yours would be the example.

teams/vsa/engineering/forms-system/custom_component_multiple.md
bradl3yC: @short000 this is the block that removes it:

```export function removeHiddenData(schema, data) {
  // null is necessary here because Rails 4 will convert empty arrays to null
  // In the forms, there's no difference between an empty array and null or undefined
  if (isHiddenField(schema) || typeof data === 'undefined' || data === null) {
    return undefined;
  }

  if (schema.type === 'object') {
    return Object.keys(data).reduce((current, next) => {
      if (typeof data[next] !== 'undefined' && schema.properties[next]) {
        const nextData = removeHiddenData(schema.properties[next], data[next]);

        // if the data was removed, then just unset it
        if (typeof nextData === 'undefined') {
          return _.unset(next, current);
        }

        // if data was updated (like a nested prop was removed), update it
        if (nextData !== data[next]) {
          return _.set(next, nextData, current);
        }
      }

      return current;
    }, data);
  }

  if (schema.type === 'array') {
    const newItems = data.map((item, index) =>
      removeHiddenData(schema.items[index], item),
    );

    if (newItems.some((newItem, idx) => newItem !== data[idx])) {
      return newItems;
    }

    return data;
  }

  return data;
}```

Its in src/platform/forms-system/src/js/state/helpers.js line 566

File: teams/vsa/engineering/forms-system/custom_component_multiple.md (hunk 3 above)
Comment (bradl3yC) [reply teams/vsa/engineering/forms-system/custom_component_multiple.md:80]: This not - This should be done using `ui:widgets`

File: teams/vsa/engineering/forms-system/custom_component_simple.md (hunk 2 above)
Comment (bradl3yC) [reply teams/vsa/engineering/forms-system/custom_component_simple.md:6]: @short000 we didn't know about the existence of this function until after we had completed our work - I don't think it would work for our application, though as we wanted to update the state of another component

File: teams/vsa/engineering/forms-system/readme.md (hunk 1 above)
Comment (bradl3yC) [reply teams/vsa/engineering/forms-system/readme.md:19]: Yeah this pretty much sums it up - we used a permanentAddress and a temporaryAddress - each had their own data and we needed to change a 3rd slice of state called currentAddress - this lived outside of them but they had it mapped to determine if their radio button was selected or not

File: teams/vsa/engineering/forms-system/readme.md (hunk 6 above)
Comment (cohnjesse) [reply teams/vsa/engineering/forms-system/readme.md:15]: The page is actually missing from the VSP documentation, on [this](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/forms/available-widgets#arraycountwidget) page there is a widget called an ArrayCountWidget, in there you can see a link to `sequential duplicate form groups` and that link is supposed to go to a page that explains this.

File: teams/vsa/engineering/forms-system/custom_component_multiple.md (hunk 3 above)
Comment (short000) [reply teams/vsa/engineering/forms-system/custom_component_multiple.md:80]: ah, so whenever we originally said "ui:fields" we should have communicated something like, "ui:fields in different ui:widgets => ui:widgets" b/c only saying "ui:fields" would imply fields within the same widget?

Comment (short000) [reply teams/vsa/engineering/forms-system/custom_component_multiple.md:80]: @mr0sari0 knowing this, should we change:

Sharing state between `ui:fields`,
- A field can not trigger a change to another field
- Two schema fields cannot alter the state of each other within a `ui:field`.

to be this:

Sharing state between `ui:widgets`,
- A field can not trigger a change to another field that is in a different widget
- Two schema fields cannot alter the state of each other across widgets

?

File: teams/vsa/engineering/forms-system/custom_component_multiple.md [hunk 7]
```diff
@@ -0,0 +1,138 @@
+# Extending RJSF Behavior Using a Custom React Component
+
+(Based mostly on [bradl3yC](https://github.com/bradl3yC)'s write-up of his solution)
+
+## The Problem:
+Using a custom component requires that you pass in a schema for that component.  It essentially maps a small slice of form state to the component.  Because the same component is used twice but we are passing in a different schema, each of those components, each of those only has access to its own individual schema/form state. (Read [Background: The Original Problem](#background-the-original-problem) for further details.)
```
cohnjesse, line 6: I believe the part `, each of those components, ` is unnecessary here and makes this a bit confusing to read

cohnjesse (approved):
Just one small change

teams/vsa/engineering/forms-system/custom_component_multiple.md
```diff
@@ -0,0 +1,138 @@
...
+
+## The Solution/Hack:
```
ncksllvn, line 8: I feel a little strange about the most comprehensive document to be highlighting solutions as hacks. It feels like we're promoting an attitude of "we're stuck with this technology, so do what you gotta do" when we definitely have the ability to influence and directly contribute to the forms system.

I don't know, just thinking how/if we could we define guidance on how to identify when you need a custom component, what the right way to build it is vs. the temporary/urgent way (I'm thinking/hoping that's what we mean here by "back" because it implies accumulating tech debt IMO), and how to identify when _not_ to use the forms system, how to provide feedback to designers about steering design towards ease of implementation ("we can't do this out-of-the-box with the forms system so we'd have to build a new widget with a LOE of X. However, we could change the design to use a X, in which case this would be easy"). Maybe if we could define those things first we could segue into this approach. 

Also, a table of contents at this root of this directory could help us navigate these docs and their objectives.

teams/vsa/engineering/forms-system/custom_component_multiple.md
bradl3yC: I was saying it should be `ui:widgets` because thats the key that custom components should be used under.  It works with a ui:field but its not how you should use that particular key/value.

teams/vsa/engineering/forms-system/custom_component_multiple.md
sshein: I've been talking about exactly to my designers - the idea that it's best to build within the form system and if a design can't accommodate that, then to identify it in advance and to go through VSP to recommend changing the form system instead of hacking outside of the form system. In order to be able to do this, both designers and developers should be able to identify whether or not something will be harder / if we can't do it "out of the box." My designers sometimes struggle with this because they see examples in other forms and aren't sure if small tweaks will add hugely to the LOE / make something custom. So we need developers that can really dig into early designs and let designers know when something looks problematic. Or some process for this. I've suggested PMs consciously writing tickets where devs look at early designs and flag anything that might be problematic.

teams/vsa/engineering/forms-system/readme.md
short000: thanks I submitted a broken link issue in github

teams/vsa/engineering/forms-system/custom_component_multiple.md
short000: thanks, that was a weird typo

teams/vsa/engineering/forms-system/custom_component_multiple.md
short000: @ncksllvn and @sshein The scope of this particular item is limited to documenting patterns that we have used. That is a first step (data collection for empirical analysis), not the end goal. The analysis of the patterns is the next step (read-through with VSP), then after that we would get into design and solutions. What you are seeing here is essentially a capture of the AS-IS state so that empirical analysis can be conducted at the technical level between developers in VSA, VSP, and possibly VAOS:

I believe what both of you are looking for might be covered in my pending check-in which has some proposed solutions in a **wish list section**, especially the first two items:

- A visual design-oriented catalog that shows the forms components integrated on test pages (100% traceability).

- A live playground feature similar to the [one for the original RSFJ](https://rjsf-team.github.io/react-jsonschema-form/).

- Documentation on *formConfig* options such as *hideIf, hideOnReview, hideOnReviewIfFalse, updateFormData, updateSchema, expandUnder* (including limitations we have run into with HideIf, expandIf, and why.)

- Visible backlog of VAFS/RJSF items in VSP that VSA developers can see so that we have transparency into what is known and not known as far as defect, features, workarounds, etc. WHEN or IF we get any of these improvements is a totally different story, but we at least need to keep track of the issues so that we don't keep stepping into them. When we get into the how/solutioning, VSA developers would be expected to work with VSP (joint effort, not something that would be expected to be handed down).

- More data mined from Slack conversations and the existing forms in source control. For example, forms 2346, 526EZ, 686, 686c-674, 996 are listed under [src/applications/disability-benefits](https://github.com/department-of-veterans-affairs/vets-website/tree/master/src/applications/disability-benefits). Would like to have an empirical analysis that summarizes *all the forms* we have built in va.<span/>gov so far and identify what patterns we are using. Are 80% just out-of-the box, and 20% special cases? Would like to know which ones are special, and for those special cases, what implementation patterns were used.

short000: We should also come up with some kind of rating system on patterns we have found, such as "recommended", "workaround", "obsolete", "avoid"

short000: > Also, a table of contents at this root of this directory could help us navigate these docs and their objectives.

@ncksllvn the root readme.md does this; however, I'll make it more obvious (put things under Contents, rename Other Guidance to External Links, etc.)

teams/vsa/engineering/forms-system/custom_component_multiple.md
mr0sari0: If widgets are preferred as best practice,I can update the doc to be widgets instead of fields.

teams/vsa/engineering/forms-system/readme.md
mr0sari0: This is pretty accurate @short000 ! The simple case just outlines how you can use `updateFormData` to modify your formData. This scenario describes a need not mentioned in that case; the need being a way to share state between different schema properties.

teams/vsa/engineering/forms-system/custom_component_multiple.md
mr0sari0: 100% visibility across design and development so everyone knows what's going on will solve a lot of confusion.

short000:
formal review has ended, but this is just markdown documentation, so further edits can be made post-merge. thanks for all the feedback

PR #8616, opened 2020-04-30
datadog design doc

wyattwalter:


gunsch (commented):
thanks for this overview! couple questions that come to mind reading it:

* are there integrations for Datadog to accomplish the following: GitHub login / GitHub Teams? BigQuery? PagerDuty?
* right now a lot of our Grafana dashboards are just configured in the UI rather than something that's checked into code, making the config management situation a bit unwieldy. can we fix that if we go to Datadog?
* we know that Prometheus / PromQL is fairly difficult to use today (noted in this doc) and we don't have many folks ramped up. are we confident that Datadog will be significantly easier?

nathanhruby:
A few things:
* DataDog's auth model is based on either accounts made in the app, or SAML.  There's no direct GitHub integration for logins so we'd need to manauge users via some crafty github+datadog terraform resources, use VA Okta/ADFS/SSOi, or build our own "VSP Auth" with Cognito 
* We are a sub-org of the main VA account, I'm not sure if we can create our own sub-orgs.  That said, everything in DD works on tag, and there is a magical "Environment" tag that we should populate in order to provide pivoting/filter.  Most DD services also allow pivoting based on the tag results from faceted search.
* There's no direct "BigQuery" integration.  Datadog isn't a a BI/Analytic platform specifically, so unsure what that would look like.  There are webhooks, an API, so coupled with custom metrics and the event timeline we could handle some specific use cases.
* DataDog [dashboards can be managed with Terraform](https://www.terraform.io/docs/providers/datadog/r/screenboard.html) though that are not pretty, this is a case where we'd want to do so for a limited set of things
* I think most people would prefer the GUI nature of DD's query builders and visualizations with autocomplete.  Also most metrics have names/descriptions provided by the integration, so it's easier to see what some of these numbers are (aids in discoverability, not really for comprehension).

wyattwalter:
I'm not sure how to best capture the "how difficult is it to learn" question. You are guided pretty well through writing a query as opposed to Prometheus' sort of blank line ready for you to input a query. It's certainly less intimidating, but I am struggling to measure intimidation 😄

joanneesteban (approved):

PR #8978, opened 2020-05-11
edited form detail ia updates

jenniferlee-dsva:
@mnorthuis - please see a couple of comments and questions. 

@beeyayjay - There's one part here that I'd like to get your confirmation on re how form number data is pulled into the CMS (as this is how part of the form detail landing page URL will be generated). Please see section where I've @'d your name.


mnorthuis (changes requested):
I have made corrections to my doc, please do not merge this. 
- removed note indicating we would not strip additional charaters from data in form name (that will be up to Drupal to handle)
- corrected breadcrumb and URLs in examples (totally missed this, thank you)
- deleted duplicate bullet

PR #9142, opened 2020-05-14
Update readme.md

kfrz:
Updates documentation to reflect (minor) change in https://github.com/department-of-veterans-affairs/vets-api/pull/4277

Also fixes a couple small typos (/vsf/vfs/)


alexpappasoddball (approved):
Nice catch on the slack channel names as well.  That should be sufficient though.

johnpaulashenfelter (approved):
Wondering a little if we just use sprints for releases since semantic versioning doesn't mean much in this context. But its 👍

PR #9153, opened 2020-05-15
Create Product SLO Tools folder

kfrz:
## Background

This PR adds initial drafts for both the product outline and the design document for Product SLO Tools. 

These docs are under review and still considered in draft status until stakeholder approval is met.


products/platform/product-slo-tools/README.md
```diff
@@ -0,0 +1,157 @@
...
+- All SLO performance config could be in one place rather than disparate places (PagerDuty, Prometheus Alerts, Grafana,
+    etc)
+
+Proposed sample configuration, using search product:
+
+  ```yaml
+    ---
+    search:
+      action: 'index' # which action?
+      controller: 'v0/search' # which endpoint?
```
omgitsbillryan, line 83: Does SLO/SLA apply also to backend services? (like EVSS for example) - Or is that outside the scope of what we want to accomplish?

products/platform/product-slo-tools/README.md
```diff
@@ -0,0 +1,157 @@
...
+- This approach requires license & volume pricing for all teams, and costs scale per node + team
+
+#### Build a dashboard manually in Grafana for each product
+
+We could build a distinct SLO dashboard in Grafana for each product.
+
+- This approach is slow and laborious
+- Would require training VFS teams on Grafana & PromQL (or future tech)
+- Tightly couples SLO configuration to dashboarding solution
+- Data integrity and configuration requires updating & exporting from Grafana directly
```
omgitsbillryan, line 152: Would it be possible to have a single dashboard with some kind of big dropdown list of endpoints? ...I suppose the actual thresholds would also need to be some kind of dropdown 🤔

platform/engineering/design-docs/2020-06-product-slo-tools.md
```diff
@@ -0,0 +1,360 @@
...
+
+- [Product SLO Tools - Product Outline](../../products/platform/product-slo-tools/product-outline.md)
+- [Chapter 2](https://landing.google.com/sre/workbook/chapters/implementing-slos/) of the Google SRE book explains the basics of setting and implementing SLOs
+- Datadog maintains [well written docs](https://docs.datadoghq.com/monitors/service_level_objectives/monitor/) for SLOs
+
+### High Level Design
+
+The high-level design of the Product SLO lifecycle includes these steps:
+
+1. VFS Product team works in conjunction with the VSP Tools teams to define SLOs
```
rroueche, line 46: @kfrz I would love to hear more opinions on this, but my take is that we need to set baseline SLO standards that all VA.gov products must live up to, so I'm hesitant to define SLOs w/the VFS team on a per product basis, unless that collaborative definition would be to define service level objectives above and beyond the standards.

products/platform/product-slo-tools/README.md
kfrz: It'll be outside of the scope for this juncture, we will likely want to look at ways that we can normalize external and on-platform if possible.

products/platform/product-slo-tools/README.md
kfrz: Yeah, I'm looking into Grafana 7.0 features that allow for variable thresholds

platform/engineering/design-docs/2020-06-product-slo-tools.md
kfrz: I agree, we should have reasonable defaults, so long as those defaults are a floor that maintain a reasonable user experience -- but we *should*, in my opinion, make sure we tailor a given product's SLO to the particular user experience, whatever that may be. 

That said we can dictate the initial baseline SLOs and build from there

PR #9503, opened 2020-05-26
update brief

melissajschaff:
I think I figured out the issue with the brief/duplicative information that we were all confused by and removed it


artsymartha68:
Report looking great so far. Push away!

Reminder that I would like to list how Vets currently make their appointments as a call-out section since we asked and since there seems to be many methods.

lenaecb (approved):

PR #9637, opened 2020-05-29
Updates to readme

karlbrown-va:
* Adds link to documentation knowledge hub
* removes last-updated and stats from readme as these are out of date
* Updates LFS bit on CMS PDF


README.md
```diff
@@ -1,44 +1,25 @@
 # VA.gov Team Repository
-*Last updated: [9/17/2019](#document-history)*
 
 This project aims to improve the lives of Veterans, their families, and their caregivers by facilitating the creation of interactive experiences on VA.gov on par with the private sector. It's comprised of a platform (the Veteran-facing Services Platform, aka VSP) which serves as the common framework for building cohesive products on VA.gov, and the VA.gov products themselves (the Veteran Facing Services, aka VFSs).
 
-This repository is used to store *non-code* documents and files created and used by teams building the VSP and VFSs.
+This repository is used to store *non-code* documents and files created and used by teams building the VSP and VFSs. It is also used for issue tracking for many parts of the VSP platform.
```
meganhkelley, line 5: The va.gov-team repo is used for issue tracking across VSP and VFS, so this could say something like: "It is also used for issue tracking across the platform and teams building on VA.gov."

README.md
```diff
@@ -1,44 +1,25 @@
...
 
 ***An important note about security for contributors to this repository: this is a public, open source space! Which means it's great for collaborating, but decidedly not great for storing sensitive information (like credentials, PII, etc). So be mindful as you work in this repo. Don't store anything sensitive here!***
 
 For information on how to use this repo, see the [Repo Guidelines](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/working-with-vsp/orientation/repo-guidelines.md).
 
-#### Top level stats:
```
meganhkelley, line 12: These stats were added by @rroueche in September 2019, I believe, so I would want to check with her on ownership of these and whether we should delete or assign someone to keep them updated. I can ask in Slack.

README.md
```diff
@@ -1,44 +1,25 @@
...
-- VSP-hosted VA.gov page loads occuring in < 5s: 88.6%
-- Average uptime last 30 days: 99.1%
-
----------------
-
 ## How to navigate this repository
 
+### Knowledge hub
+The best starting place is the knowledge hub, that provides links to the content in this repository and elsewhere:
+https://department-of-veterans-affairs.github.io/va.gov-team/
```
meganhkelley, line 15: We tend to direct people to the TinyURL version of this since it's easier to remember: www.tinyurl.com/vagovdocs 

Also haven't ever called it a knowledge hub, but I will confirm with my team that folks are fine with that term.

README.md
```diff
@@ -1,44 +1,25 @@
...
+
 ### [/platform](./platform)
-All resources, templates, guides, and processes that support development of VFSs on VA.gov.
+All VSP resources, templates, guides, and processes that support development of VFSs on VA.gov.
```
meganhkelley, line 18: Not all platform resources are owned by VSP — for example, "research" is something that sits pretty squarely with DEPO, and "CMS" is included as part of the platform folder but is not part of VSP.

README.md
```diff
@@ -1,44 +1,25 @@
...
 ### [/products](./products)
-All VFSs (aka Va.gov features and products)
+All VFSs (aka VA.gov features and products)
 
 ### [/teams](./teams)
-All teams currently building VSP or building VFSs.
-
----------------
-
-## Document history
```
meganhkelley, line 36: Is there a reason you deleted the document history?

README.md
```diff
@@ -1,44 +1,25 @@
...
-| Updated | Description  | User  |  
-|---|---|---|
-| 9/16/2019 at 3:00 PM | Added link to repo guidelines file | Mikki Northuis |
-| 9/4/2019 at 3:30 PM | Updated to reflect new repo structure | Mikki Northuis |
-| 6/28/2019 at 1 PM | Stripped most content, reroute users to `va.gov-team` or `va.gov-vfs-teams` as appropriate. | Andrew Gunsch |
-| 6/1/2019 at 1:45 PM  | Added "TBD" for sections still being fleshed out  | Ryan Luu |
-| 6/1/2019 at 11:06 AM  | Removed POC, added link for Platform users, added Document history  | Layla Soileau |  
+All teams currently building the VSP or building VFSs.
```
meganhkelley, line 24: Suggestion: Internal resources for all teams currently building on or in support of VA.gov.

meganhkelley (commented):
Left some comments in line, and can follow up with a couple folks based on the action items I assigned myself here :) 

Re: CMS file, I assume you deleted this because it was causing an LFS issue? Did you re-upload via LFS, and/or reach out to the author to resolve the problem? (I can reach out to provide instructions for future, but wondering current status)

README.md
karlbrown-va: That sounds good. Will make that change.

README.md
karlbrown-va: IMO it's a bad idea to have such things committed to a readme. They will always be out of date. Instead we could provide a link to https://www.va.gov/performance-dashboard or the google analytics dashboard for those with access?

README.md
karlbrown-va: The top of the page calls it a knowledge hub which is why I used that terminology, but happy to use the team's preferred verbiage...

README.md
karlbrown-va: gotcha, updatd. this. thanks!

README.md
karlbrown-va: Linked to the tinyurl url

README.md
meganhkelley: We don't have a formal stance on this (doc-level change summaries). It's nice to see changes at a glance (particularly because the GH history can be inaccurate — migrated files are shown as created by the person who migrated, and renamed files or files that move folders are shown as new upon renaming/moving), but we don't recommend one way or the other currently on VSP. On this readme — which does not have another "owner" or "point of contact" it also serves as a quick way to show who's been making significant updates. But again, in the absence of a formal stance/process here, I could go either way.

README.md
meganhkelley: Checked with Rachael — let's cut it for now, and can add back analytics later if/when we do a more complete overhaul of an intro page of sorts.

Issue #503, opened 2019-07-05
Va.gov Redirects - Establish Drupal as source of truth for redirects on www.va.gov


ricetj:
## User Story or Problem Statement
As a content editor, I need to be able to change redirects as my site is being migrated to VA.gov so that Veterans can access the information they need.

### High Level User Story/ies or Description: 
We currently keep redirect information in vagov-content, but we should use the Drupal API as the source of truth for redirects going forward.

We'll need to get a schema for the structure of redirect data in Drupal as well as some test data in the dev Drupal instance to work on this.

## VSP Objective (Q1 & Q2 2020)
O1: Any volume of simultaneous customers have a delightful, self-service, end-to-end experience using the Platform to build on VA.gov.

## Measuring Success (KRs/KPIs)
- [ ] [VSP KR] Reduction in volume of “non-human-needed” support requests month over month
- [ ] [VSP KR] Increase in customer NPS

## Goals
Set Drupal as the source of truth for redirects.

ricetj:
Connected PR: https://github.com/department-of-veterans-affairs/devops/pull/4282

ricetj:
(03/20/2019)
This has been removed from the MVP. I'm going to drop from current sprint and put back to ready for now.

ricetj:
https://dsva.slack.com/archives/CDJQ0JCE4/p1572633390094700?thread_ts=1572551914.087100&cid=CDJQ0JCE4

meganhkelley:
Please keep VSP Content & IA posted as y'all begin working on this!

ricetj:
Need to talk to Robbie and Jeff B @ricetj on the future of this.

PR #9658, opened 2020-05-29
Content validation design doc

billfienberg:
This design doc builds upon [Chris V.'s Content Separation design doc](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/design-docs/2020-04-09-separate-content-build.md), and the [`content-build` repo that Brooks has been working on](https://github.com/department-of-veterans-affairs/content-build).


platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,148 @@
...
+
+##### Content Writers
+
+Content writers want to quickly draft, publish, and deploy content. However, `vets-website` is only deployed to production once per day, and `vets-website` deploys to staging take 20+ minutes. To allow content writers to move faster, there is a partial deploy, which only includes static page changes (`vagov-content` and Drupal).
+
+Unfortunately, the partial deploys have no content validation. That means that content writers don't have a proactive reactive way to learn if their content includes broken links or accessibility errors.
+
+##### Users
+
+Content-only deploys create an opportunity for end users to encounter broken links and/or accessibility errors from CMS content that wasn't validated before being deployed.
```
cvalarida, line 76: Expand on this a little. This happens because content can bypass any content validation checks if it's deployed via the content-only deployment.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,148 @@
...
+
+##### FE Tools team
```
cvalarida, line 78: The scope is much broader here. This affects all teams across all contracts who are trying to ship code on the VA.gov front end, not just the FE Tools team.

cvalarida (commented):
So far so good. I look forward to the Specifics section getting fleshed out.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,148 @@
...
+The `check-broken-link` plugin does the following:
+
+1. Loop through all of the HTML files in the Metalsmith pipeline
+2. Extract all of the broken `href`/`src` values using the `getBrokenLinks` helper
+3. Formats the result into useful console output
+4. Breaks the build on production, and logs the output on lower environments
+
+##### Accessibility Checking
+
+
```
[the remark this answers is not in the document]
billfienberg: When the integration step of the build fails, there is a generic notification sent to the #devops-deploys channel via https://github.com/department-of-veterans-affairs/vets-website/blob/f29aa3a385d03479308f78c0c22c292ecae93f20/Jenkinsfile#L86, correct? If so, the violation must be reported manually because the error message isn't descriptive, correct?

For example: https://dsva.slack.com/archives/C37M86Y8G/p1591099447374800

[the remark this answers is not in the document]
cvalarida: I believe that's all correct, yeah.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,148 @@
...
+### Work Estimates
+
+COMING SOON
+
+### Alternatives
+
+#### Highest quality, but lowest speed
+
+If we could proactively validate CMS content before it's published, we could eliminate a source of downstream broken links and/or accessibility errors, which would improve the experience for end users and/or reduce the number of frontend deploys that require manual intervention. The downside of this approach would be figuring out how to add the content validation to the CMS, and/or adding friction to the content writers workflow.
```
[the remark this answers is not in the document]
swirtSJW: @cvalarida We added link checking to the CMS several months back On node save every link both internal and external are tested and a report generated for that node.   
image
A bright red box appears if there are any that are broken.
image

508 compliance is a little tougher because the front end templates do not match the cms templates,

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,148 @@
...
+
+
+
+#### Pain points
+
+##### Content Writers
+
+Content writers want to quickly draft, publish, and deploy content. However, `vets-website` is only deployed to production once per day, and `vets-website` deploys to staging take 20+ minutes. To allow content writers to move faster, there is a partial deploy, which only includes static page changes (`vagov-content` and Drupal).
+
+Unfortunately, the partial deploys have no content validation. That means that content writers don't have a proactive reactive way to learn if their content includes broken links or accessibility errors.
```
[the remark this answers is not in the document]
swirtSJW: Content writers do have validation from the node link report.

Where it still has the potential to fall down right now is that  if a path to a page in drupal is changed, drupal automatically creas a redirect from the old path to the new path, so when the links are checked, they still work.   The original plan was that the Front end would import redirects from drupal into Nginx when the content build takes place.  However that is not happening, so the front end does not have the redirects, so a moved page, results in broken links from pre-existing content links.

platform/engineering/design-docs/2020-05-26-content-validation.md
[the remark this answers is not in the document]
swirtSJW: This checker even reaches out to the public website in case it finds pages that are not known to drupal.  

All contributed back to benefit Drupal's everywhere :)  
https://www.drupal.org/project/node_link_report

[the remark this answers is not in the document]
swirtSJW: It is on my roadmap to add link accessibilty checking to the link checker which would warn on an editor creating a link with no text in it.   This was a recent issue that blocked a content release a couple of weeks back.
https://www.drupal.org/project/node_link_report/issues/3145477

platform/engineering/design-docs/2020-05-26-content-validation.md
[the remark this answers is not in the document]
kevwalsh: So many of the broken links come from URLs that change, according to decisions by the Content & IA team, usually in conjunction with Public Websites. In that workflow, an engineer (usually @ncksllvn, i believe, but others seem to do it too) makes a code change for the redirect, and has to time the release of the redirect code to match the content release of the URL change in the content layer. It's a fragile and very time-consuming process. 

At one point there was [an epic on the operations team's backlog to establish Drupal as source of truth for redirects](https://github.com/department-of-veterans-affairs/va.gov-team/issues/503). but i just saw that the initial progress towards that was just removed in [PR 6999](https://github.com/department-of-veterans-affairs/devops/pull/6999)

On our end, we have [a new user story](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/1937) that may help a bit with this, much like the "outgoing" node link report mentioned by Steve. But these solutions rely heavily on humans to do things in the right order for things to not fall down.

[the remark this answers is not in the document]
kevwalsh: I should add, as we add hundreds of editors from outside the tight Github/Slack communication flows (eg Public Affairs Officers at 2000+ VA facilities), we won't be able to rely on this type of coordination for redirects.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+
+Once the `content-build` repo is successfully extracted from `vets-website`, the Metalsmith static content build pipeline will be deleted from `vets-website`. Webpack will handle the application build in `vets-website`, and Metalsmith will handle the static content build in the `content-build` repo. This will create two separate pipelines for built code to be deployed to S3, which will prevent failures in one from blocking the other.
+
+The validation of the static content coming from the CMS will occur during the Metalsmith build.
+
+## Specifics
+
+### Detailed Design
+
+Currently, the static CMS content is only validated during the full deploy, which only happens once per day. To maintain some form of parity with that that process, the content validation will happen in a daily scheduled job.
```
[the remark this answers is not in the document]
rianfowler: This section should have a lot more information on the implementation. What is going to run? What is going to be tested? Is there documentation on the existing process that can be referenced? What are the inputs and outputs? 

If you've already documented this in the background, then reference it here. something like, "the existing link checking process will be used. it will be triggered by a daily job blah blah."

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+2. The [`vagov-content` repo](https://github.com/department-of-veterans-affairs/vagov-content/blob/master/pages/404.md)
+
+This design doc will focus on validating content from the Drupal CMS.
+
+##### Amazon S3
+
+The frontend of VA.gov is served as static files via Amazon S3. Since `vets-website` currently handles both application and content building, a failure in one blocks both from being deployed to S3 during a full deploy.
+
+###### Broken Link Checking
+
```
rianfowler, line 77: summarize what this validates

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+The [`check-broken-link` middleware](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/stages/build/plugins/check-broken-links/index.js) is a Metalsmith plugin that happens near the end of the Metalsmith static content build pipeline. It does the following:
+
+1. Loop through all of the HTML files in the Metalsmith pipeline
+2. Extract all of the broken `href`/`src` values using the `getBrokenLinks` helper
+3. Formats the result into useful console output
+4. Breaks the build on production, and logs the output on lower environments
+
+###### Accessibility Checking
+
```
rianfowler, line 86: summarize what this validates

rianfowler:
Good start- I think we need some more details as noted. I intend to review this with andrew and dror who are not intimately familiar with what this stuff.

platform/engineering/design-docs/2020-05-26-content-validation.md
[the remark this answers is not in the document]
billfienberg: @swirtSJW @kevwalsh, thank y'all for the feedback so far. I'll let you know once I'm done updating the design doc to reflect this improved understanding.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+2. [Run the content build](https://github.com/department-of-veterans-affairs/vets-website/blob/43ea0bdccd5e53886e3e38ea27d3f8e8e7bd9038/script/build.sh#L61)
+
+Previously, Webpack was entangled in the content build, but no longer. Brooks is currently working on moving the content build into a [`content-build`](https://github.com/department-of-veterans-affairs/content-build/) repo to separate it from `vets-website`.
+
+##### Content Sources
+
+Static content currently comes from two sources:
+
+1. The Drupal CMS
+2. The [`vagov-content` repo](https://github.com/department-of-veterans-affairs/vagov-content/blob/master/pages/404.md)
```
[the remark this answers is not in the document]
ncksllvn: Also - we still have static content in vets-website - the entire `src/site/assets` directory iirc

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+
+###### Accessibility Checking
+
+The accessibility checks only happen in CI ([the Integration step in Jenkins](https://github.com/department-of-veterans-affairs/vets-website/blob/master/Jenkinsfile#L81), which only happens during the full `vets-website` build.
+
+Before accessibility checks can be run, the sitemap is generated via the [`create-sitemap` Metalsmith plugin](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/stages/build/plugins/create-sitemaps.js). Nightwatch iterates over all the paths found in the sitemap, and runs the [aXe accessibility checker](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/platform/testing/e2e/nightwatch-commands/axeCheck.js) to look for violations.
+
+When violations are found, the build breaks and violations must be reported manually.
+
+The accessibility checks are skipped during a content-only deploy.
```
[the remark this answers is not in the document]
ncksllvn: All tests (except link checking) are skipped during content-only deploys by design - for speed of publishing. The link-checking runs as part of the build itself rather than as a separate step whereas unit tests run before before e2e and accessibility tests afterwards.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+The frontend of VA.gov is served as static files via Amazon S3. Since `vets-website` currently handles both application and content building, a failure in one blocks both from being deployed to S3 during a full deploy.
+
+###### Broken Link Checking
+
+The [`check-broken-link` middleware](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/stages/build/plugins/check-broken-links/index.js) is a Metalsmith plugin that happens near the end of the Metalsmith static content build pipeline. It does the following:
+
+1. Loop through all of the HTML files in the Metalsmith pipeline
+2. Extract all of the broken `href`/`src` values using the `getBrokenLinks` helper
+3. Formats the result into useful console output
+4. Blocks the deployment on production or logs the error output on lower environments
```
ncksllvn, line 83: Some historical info might be useful - originally (on vets.gov), all content was in vets-website in a `content/` directory and you would open a PR to edit content. CI would then run against that PR. Link-checking would run as part of the build the same way it does today. If a broken link was found, CI would fail, and you wouldn't be able to merge. This meant that broken links never made their way into the master branch.

Accessibility checks had the same deal - errors never made their way into master.

Also - I think Drupal contains some link-checking internally nowadays but it isn't as comprehensive or something.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+
+The accessibility checks are skipped during a content-only deploy.
+
+Additionally, we have the preview server which shows content editors any accessibility violations, but they must deliberately go to the server from a link in Drupal to see this.
+
+#### Pain points
+
+##### Content Writers
+
+Content writers want to quickly draft, publish, and deploy content. However, `vets-website` is only deployed to production once per day, and `vets-website` deploys to staging take 20+ minutes. To allow content writers to move faster, there is a partial deploy, which only includes static page changes (`vagov-content` and Drupal).
```
ncksllvn, line 101: vets-website deploys take 20 mins to go to staging bc CI runs again on master - the deploy to staging itself takes like a minute.

But also, staging.va.gov has pretty much zero value to content writers, so it's probably not even worth mentioning. Writers are editing in Drupal prod and looking at their changes via the preview server. So they are basically always in prod. Drupal staging is wiped by a downstream migration from Drupal prod every now and then, don't know when that job runs. But it basically means that Drupal staging reflects old Drupal prod content and any changes you make are temporary, so basically, there is no value to anyone to make content changes to Drupal staging.

(related but not completely on topic - this whole environment config between staging.va.gov and drupal staging has been a serious pain point for app teams, but i think folks roll their eyes now whenever i bring it up 🙄 )

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+
+Unfortunately, the partial deploys have no content validation. That means that content writers don't have a proactive way to learn if their content includes broken links or accessibility errors.
+
+##### Users
+
+Content-only deploys create an opportunity for end users to encounter broken links and/or accessibility errors. This happens because content can bypass any content validation checks if it's deployed via the content-only deployment.
```
ncksllvn, line 107: Is this true for broken links? See my earlier comment - the broken links run as part of the build (website compilation), not as an isolated step. Do we skip link-checking during content deploys? We would have had to do that via a flag passed into the build command or something. Seems like a bad idea.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+
+##### Anybody that is trying to ship code for the VA.gov front end
+
+If invalid content is breaking the production build, and therefore blocking the production deploy, then no team on any contract is able to ship their frontend code to VA.gov.
+
+### High Level Design
+
+Once the `content-build` repo is successfully extracted from `vets-website`, the Metalsmith static content build pipeline will be deleted from `vets-website`. Webpack will handle the application build in `vets-website`, and Metalsmith will handle the static content build in the `content-build` repo. This will create two separate pipelines for built code to be deployed to S3, which will prevent failures in one from blocking the other.
+
+The validation of the static content coming from the CMS will occur during the Metalsmith build.
```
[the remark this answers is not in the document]
ncksllvn: > The validation of the static content coming from the CMS will occur during the Metalsmith build.

I think what we're going for here is a separation of concerns or delineation of responsibilities, but so we're clear, engineers will always need to support the content writers. It's cool that we're going for content validation not holding up app deploys, but also, we can't just put the content layer aside and act like it isn't our problem anymore.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Specifics
+
+### Detailed Design
+
+The content validation will happen in a daily scheduled job.
```
ncksllvn, line 127: To run in a daily scheduled job, I assume that it's report-only at that point? Because as an isolated job, it won't be blocking any deployments.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+
+##### Accessibility errors
+
+1. Confirm that the content build is passing
+2. Modify an HTML file, like [`header.html`](https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/includes/header.html) to have an accessibility error
+3. Run the build again to confirm that it is failing
+
+### Caveats
+
+- Our [function to determine if a link is broken](https://github.com/department-of-veterans-affairs/vets-website/blob/e8b68850fc83fdae4386fe4ab392770dde38faee/src/site/stages/build/plugins/check-broken-links/helpers/isBrokenLink.js#L16) doesn't check if external links are broken.
```
ncksllvn, line 159: I don't think this this worth mentioning really because the proposed design here doesn't change this behavior. We've never validated broken links.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+
+- Neither the current content validation nor the proposed content validation will **prevent** invalid content from making it to production. But this proposal would prevent unrelated teams from being blocked by invalid content.
+
+#### Questions
+
+### Work Estimates
+
+The following estimates vary greatly depending on who's doing the work.
+
+1. Finish extracting the `content-build` repo out of the `vets-website` repo - ?
```
ncksllvn, line 182: I'm assuming the content-build repo will need to be sibling to vets-website at all times, correct? CMS content is everywhere, from the mega menu to user flows. Does this mean vets-website, content-build, and vagov-content will all need to be cloned side by side?

platform/engineering/design-docs/2020-05-26-content-validation.md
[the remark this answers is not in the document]
ncksllvn: @swirtSJW how come when we see reports of broken link they seem to originate from failing builds? 

If we're confident with Drupal's link validation, then I think we should just delete link validation from vets-website.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,201 @@
...
+
+Additionally, we have the preview server which shows content editors any accessibility violations, but they must deliberately go to the server from a link in Drupal to see this.
+
+#### Pain points
+
+##### Content Writers
+
+Content writers want to quickly draft, publish, and deploy content. However, `vets-website` is only deployed to production once per day, and `vets-website` deploys to staging take 20+ minutes. To allow content writers to move faster, there is a partial deploy, which only includes static page changes (`vagov-content` and Drupal).
+
+Unfortunately, the partial deploys have no content validation. That means that content writers don't have a proactive way to learn if their content includes broken links or accessibility errors.
```
ncksllvn, line 103: Does ` partial deploy` == content deploy?

Someone should definitely confirm broken links don't block content deploys because I thought they did.

ncksllvn (commented):
One of the biggest sources of confusion for me with these deployments has been the "time traveling" we do during the content deployments. As in -

1. vets-website release goes out at 3 PM ET
1. people merge to master fifty times
1. content-only deployment is issued at 9 AM ET the following day
1. the deployments checks out the last vets-website release, fifty commits behind
1. the deployments builds using latest drupal content and latest vagov-content
1. the finished deployment is archived to github somewhere, and I think it replaces the commit SHA of the latest vets-website release on s3

Maybe it is just me, but the whole process of rewinding vets-website master to use the latest release has always been super confusing to me.

ncksllvn:
Meta question - do these go into `platform/engineering/design-docs/` nowadays? I figured this was an RFC.

billfienberg:
> Meta question - do these go into `platform/engineering/design-docs/` nowadays? I figured this was an RFC.

Chris V. put his recent [design doc about separating the content build from the app build](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/design-docs/2020-04-09-separate-content-build.md) in `platform/engineering/design-docs/`, so I just followed his example.

platform/engineering/design-docs/2020-05-26-content-validation.md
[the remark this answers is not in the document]
swirtSJW: @ncksllvn I think most now are either related to something being unpublished or moved to a new path.  The node link report warns if linking to something that is unpublished, but it does not / can not prevent it.

platform/engineering/design-docs/2020-05-26-content-validation.md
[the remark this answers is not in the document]
cvalarida: :point_up: That context is probably good to add to the doc, @billfienberg .

platform/engineering/design-docs/2020-05-26-content-validation.md
cvalarida: Oh hey, look at that. So they do.
image
From `checkForBrokenLinks` in `jenkins/common.groovy`.

I even made that change, but I totally forgot. My mistake! I've led you astray, @billfienberg.

platform/engineering/design-docs/2020-05-26-content-validation.md
cvalarida: Right, yeah. Looks like broken links are accounted for after all. I gave Bill the wrong information there. :raised_hand_with_fingers_splayed: : My bad.

The accessibility checks aren't run, but broken links are checked.

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: When I say `partial deploy`, I am referring to the content deploy. I got the term `partial deploy` from the [Separate Content Build design doc](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/design-docs/2020-04-09-separate-content-build.md#deployment), which links to the [Deploy page of the VA.gov workflow docs](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started/workflow/deploy/#partial-deploy--static-page-changes-only).

billfienberg: Oh yea. There it is in `common.groovy`.

```
    // Only break the build if broken links are found in master
    if (IS_PROD_BRANCH || contentOnlyBuild) {
      throw new Exception('Broken links found')
    }
  } else {
    echo "Did not find broken links."
  }
```

https://github.com/department-of-veterans-affairs/vets-website/blob/master/jenkins/common.groovy#L177

platform/engineering/design-docs/2020-05-26-content-validation.md
swirtSJW: CMS staging and dev are rebuilt every time a commit is merged into master.   When it is rebuilt, it pulls a new db copy from prod.

platform/engineering/design-docs/2020-05-26-content-validation.md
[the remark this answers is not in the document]
swirtSJW: Also, our link validation happens on a save in drupal.  It is not blocking, only reported.   It is up to editors to note and fix.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,208 @@
...
+The frontend of VA.gov is served as static files via Amazon S3. Since `vets-website` currently handles both application and content building, a failure in one blocks both from being deployed to S3 during a full deploy.
+
+###### Broken Link Checking
+
+The [`check-broken-link` middleware](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/stages/build/plugins/check-broken-links/index.js) is a Metalsmith plugin that happens near the end of the Metalsmith static content build pipeline. It does the following:
+
+1. Loop through all of the HTML files in the Metalsmith pipeline
+2. Extract all of the broken `href`/`src` values using the `getBrokenLinks` helper
+3. Formats the result into useful console output
+4. Blocks the deployment on production or logs the error output on lower environments
```
billfienberg, line 87: Slack notification sent to the #cms-team channel informing them of broken links found in the build. 

https://dsva.slack.com/archives/CT4GZBM8F/p1591303292290600

image

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Code Location
+
+The content validation script and scheduled job config will live in the [new `content-build` repo](https://github.com/department-of-veterans-affairs/content-build).
+
+### Testing Plan
+
+1. Configure a scheduled job to run a script once per workday
+2. Write an initial script that logs when the script is run
+3. Once the scheduled job is running at the correct time, replace the initial script with the actual script that runs the content validation
```
bkjohnson, line 145: This feels a bit confusing to me.  Can you try rewording it or explaining it differently?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Logging
+
+Currently, invalid content is logged during staging deploys. And failed builds trigger generic [Slack notifications](https://dsva.slack.com/archives/C37M86Y8G/p1591099447374800).
```
bkjohnson, line 149: Do you plan on changing this?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,208 @@
...
+
+##### Accessibility errors
+
+1. Confirm that the content build is passing
+2. Modify an HTML file, like [`header.html`](https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/includes/header.html) to have an accessibility error
+3. Run the build again to confirm that it is failing
+
+### Caveats
+
+- Our [accessibility check](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/assets/js/execute-axe-check.js#L8) only renders a banner on the preview server.
```
bkjohnson, line 167: This seems like it could easily get confused with the accessibility tests that Jenkins runs.  Can you differentiate them a bit more?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Privacy Concerns
+
+There are no new privacy concerns with a separated build process.
+
+### Open Questions and Risks
+
+#### Risks
+
+- Neither the current content validation nor the proposed content validation will **prevent** invalid content from making it to production. But this proposal would prevent unrelated teams from being blocked by invalid content.
```
bkjohnson, line 181: This might be me being picky, but I'm not sure this is 100% true.  When you say "production" here, are you talking about the live va.gov site, or do you mean the CMS production server?  We have had incidents of bad content breaking the content build, which prevents new content from going to prod on va.gov

platform/engineering/design-docs/2020-05-26-content-validation.md
bkjohnson: Yes, that's correct.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Alternatives
+
+#### Highest quality, but lowest speed
+
+If we could proactively validate CMS content before it's published, we could eliminate a source of downstream broken links and/or accessibility errors, which would improve the experience for end users and/or reduce the number of frontend deploys that require manual intervention. The downside of this approach would be figuring out how to add the content validation to the CMS, and/or adding friction to the content writers workflow.
+
+#### Fastest speed, but lowest quality
+
+We could stop breaking the `vets-website` build due to invalid content. That would reduce the burden on the content writers and/or reduce the number of deploys requiring manual intervention. However, that would likely mean end users experience broken links and/or accessibility errors more often.
```
bkjohnson, line 200: This line seems like it can be interpreted as contradicting an earlier statement:

> We could stop breaking the `vets-website` build due to invalid content.

The previous statement:
> Neither the current content validation nor the proposed content validation will **prevent** invalid content from making it to production.

bkjohnson (commented):
Me being picky, also asking for some clarifications.

billfienberg:
@swirtSJW @kevwalsh, I have made a bunch of updates based on your feedback.

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: I actually removed this caveat because I added it before I realized that there are actually two different types of a11y checks. 

1. During the Metalsmith build
2. During the Integration stage of Jenkins

ncksllvn: FYSA -

> During the Metalsmith build

This is totally non-blocking and runs only on the preview server. It _can_ be run during a regular build or watch task, but that's not what it was designed for.

> During the Integration stage of Jenkins

This _does_ block regular full deployments. The gist is that -

1. Everything is good and CI is passing
1. A new page is published in Drupal
1. CI runs on vets-website master or on a PR branch
1. The new page from Drupal is served into vets-website CI
1. The new page has an accessibility issue, and now all branches in vets-website have to fall back to the cached Drupal data on S3 in order to build successfully.
        - This doesn't always work. Sometimes the cache is considered invalid if the GraphQL queries have changed since the last cache was uploaded, so if there's an accessibility issue at the same time, CI is blocked for all branches.
1. Full deployments can't go out. 
    - Content-only deploys skip accessibility tests, so content can (ironically) still go out.
    - If vets-website master passed okay during the last build, we can do an app-code deployment by skipping the pre-release job in Jenkins. (Just unchecking the box for the deploy job)

So...we could technically achieve a full deployment by first doing an app-code deploy, then doing a content-only deploy. But we can't do them both at the same time if there are accessibility issues on a Drupal page.

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: The new content validation script will continue including logs in the build log. On top of that, it will send an automated report of invalid content to the #cms-team Slack channel.

bkjohnson:
One thing that may be worth considering in the future is adding a way to version the content, since it is my understanding that we are just deploying the latest content paired with a specific release of `vets-website`.  Instead, if we have a specific content version paired with a specific `vets-website` version, that may allow us to be more confident that what we are deploying is valid and doesn't contain broken links or other accessibility errors.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,227 @@
...
+
+#### Pain points
+
+##### Content Writers
+
+Content writers want to quickly draft, publish, and deploy content. However, `vets-website` is only deployed to production once per day. To allow content writers to move faster, there is a [partial deploy](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started/workflow/deploy/#partial-deploy--static-page-changes-only), which only includes static page changes (`vagov-content` and Drupal).
+
+Unfortunately, the content-only deploys have no accessibility checking. That means that content writers don't have a proactive way to learn if their content includes accessibility errors. Adding accessibility checking to the CMS is on the roadmap, but it doesn't currently exist.
+
+##### Users
```
rianfowler, line 120:
```suggestion
##### VA.gov Users
```

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,227 @@
...
+
+To enable a faster publishing workflow, all tests (except link checking) are intentionally skipped during content-only deploys. That creates an opportunity for end users to encounter accessibility errors. Additionally, inconsistencies between how the CMS handles redirects and how `vets-website` handles redirects, it's possible for users to encounter broken links in production.
+
+##### FE Tools team
+
+Invalid content causes warning logs during staging deploys, and breaks the build during production deploys. These broken builds then need manual intervention, and require coordinating with another team.
```
rianfowler, line 126:
```suggestion
Invalid content causes warning logs during staging deploys, and breaks the build during production deploys. These broken builds then need manual intervention, and require coordinating with the CMS team.
```

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,227 @@
...
+
+##### Anybody that is trying to ship code for the VA.gov front end
+
+If invalid content is breaking the production build, and therefore blocking the production deploy, then no team on any contract is able to ship their frontend code to VA.gov.
+
+## High Level Design
+
+Once the `content-build` repo is successfully extracted from `vets-website`, the Metalsmith static content build pipeline will be deleted from `vets-website`. Webpack will handle the application build in `vets-website`, and Metalsmith will handle the static content build in the `content-build` repo. This will create two separate pipelines for built code to be deployed to S3, which will prevent failures in one from blocking the other.
```
rianfowler, line 134:
```suggestion
As part of the [build separation wor](https://github.com/department-of-veterans-affairs/va.gov-team/issues/2719)k, two separate pipelines will be created for for built code to be deployed to S3 to prevent failures in one from blocking between CMS and application deployments. 

The Metalsmith static content build will be extracted to the `content-build` repo and deleted from `vets-website`, and its deployment pipeline. Webpack will handle the application build for the `vets-website` repo, and Metalsmith will handle the static content build for the `content-build` repo. 
```

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,227 @@
...
+
+The validation of static content will take place in a scheduled job that runs every workday. Since the broken link check happens during the Metalsmith build, and the accessibility check relies on the output of the content build, the scheduled job will run a script to produce a build that only has the steps necessary for content validation. That build will only be used for reporting broken links and accessibility errors, and will have no impact on deploys. Those reports will be delivered to the #cms-team channel via existing Slack integrations.
```
rianfowler, line 136:
```suggestion
The validation of static content will take place in a scheduled job that runs every workday instead of ???. Since the broken link check happens during the Metalsmith build, and the accessibility check relies on the output of the content build, the scheduled job will run a script to produce a build that only has the steps necessary for content validation. That build will only be used for reporting broken links and accessibility errors, and will have no impact on deploys. Those reports will be delivered to the #cms-team channel via existing Slack integrations.
```

rianfowler: This is what we're making a decision about on this document so we should emphasize that somehow.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Code Location
+
+The content validation script and scheduled job config will live in the [new `content-build` repo](https://github.com/department-of-veterans-affairs/content-build).
+
+### Testing Plan
+
+1. Once the content validation script is written, we can run the script locally.
+2. Once the script is behaving correctly in local environment, we can call the script in a scheduled job.
+3. Once the script is being called in the scheduled job, we can monitor and optimize it.
```
rianfowler, line 154: We should add some more details to this. We should at least do positive and negative testing (locally and in the CI) and validate that errors are pushed to the right notification channels. Describe how you intend to do the different types of testing.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,227 @@
...
+
+Currently, invalid content is logged in the build log during staging deploys. And failed builds trigger generic [Slack notifications](https://dsva.slack.com/archives/C37M86Y8G/p1591099447374800).
+
+The updated content validation approach will continue logging invalid content errors in the build log, and will also post an automated report of the invalid content in the #cms-team Slack channel.
+
+### Debugging
+
+##### Broken links
+
+1. Confirm that the content build is passing
```
rianfowler, line 166: What can be used for this? Jenkins logs? Slack notifications?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,227 @@
...
+2. Modify an HTML file, like [`header.html`](https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/includes/header.html), to have a broken link
+3. Run the build again to confirm that it is failing
```
rianfowler, line 168: this sounds like a test plan. debugging should probably just detail how to find and interpret the errors in jenkins: 

>How users can debug interactions with your system. When designing a system it's important to think about what tools you can provide to make debugging problems easier. Sometimes it's unclear whether the problem is in your system at all, so a mechanism for isolating a particular interaction and examining it to see if your system behaved as expected is very valuable. Once a system is in use, this is a great place to put tips and recipes for debugging. If this section grows too large, the mechanisms can be summarized here and individual tips can be moved to another document.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,227 @@
...
+
+##### Accessibility errors
+
+1. Confirm that the content build is passing
+2. Modify an HTML file, like [`header.html`](https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/includes/header.html), to have an accessibility error
+3. Run the build again to confirm that it is failing
```
rianfowler, line 174: This also sounds like a test plan.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Caveats
+
+- The current broken link checking in Drupal can handle redirects, but those redirects are not currently synchronized with `vets-website`. When the redirects are not synchronized between `vets-website` and Drupal, a page that is moved in Drupal may pass the Drupal broken link check but fail the `vets-website` broken link check.
```
rianfowler, line 178: I don't understand why this is a caveat. anchor tags from VA's CMS shouldn't point to redirects- redirects are for external referrers i.e. content we don't control.

rianfowler (approved):

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: I added this based on [a comment from Steve W](https://github.com/department-of-veterans-affairs/va.gov-team/pull/9658#discussion_r434864015). 

Do you think this paragraph needs to be moved (perhaps to the **Background** section), is incorrect, is confusing, and/or something else?

billfienberg:
@swirtSJW, I've made some updates since the last time you reviewed this. Now that this is ready for review, will you rereview this PR and either request changes or approve it?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,278 @@
...
+
+As part of the [build separation wor](https://github.com/department-of-veterans-affairs/va.gov-team/issues/2719)k, two separate pipelines will be created for for built code to be deployed to S3 to prevent failures in one from blocking between CMS and application deployments.
+
+The Metalsmith static content build will be extracted to the `content-build` repo and deleted from `vets-website`, and its deployment pipeline. Webpack will handle the application build for the `vets-website` repo, and Metalsmith will handle the static content build for the `content-build` repo.
+
+The validation of static content will take place in a scheduled job that runs every workday instead of ???. Since the broken link check happens during the Metalsmith build, and the accessibility check relies on the output of the content build, the scheduled job will run a script to produce a build that only has the steps necessary for content validation. That build will only be used for reporting broken links and accessibility errors, and will have no impact on deploys. Those reports will be delivered to the #cms-team channel via existing Slack integrations.
+
+## Specifics
+
+### Detailed Design
```
gunsch, line 150: this makes sense as a starting point, but I think needs a _lot_ more detail.

each of the sentences you have below should probably have their own heading describing that decision and explanations about how to implement each, and maybe describing how they affect the pain points. but at the moment there isn't enough described to feel like this really communicates what's happening.

some initial questions that come to mind as examples of what detail this still needs:

> Using the current content build script as a reference, we will write another build script that only includes the steps necessary for the content validation. 

* Does that mean you'll remove content validation from the current content build script?
* What runs this new build script?
* What happens when this new build script detects invalid content?
* Will invalid content detection block a build or deployment?

> We will reuse the existing broken link check from the current content build. 

* does that mean just the checking logic? what about how reporting is done?

> And using the existing accessibility tests as a reference, we will configure the accessibility tests to not stop on failure so we can identify all the accessibility errors.

* Wait, are you using the accessibility tests as a reference for other work? Or changing the accessibility tests? This sounds like both, but to me those are mutually exclusive things
* What does "not stop on failure" mean --- just making sure we check for _all_ errors instead of stopping at the first error? Or does it mean not stopping a build/deploy?


To be clear, these are all separate questions that I feel like the design doc doesn't answer currently.

gunsch: A few other questions I have reading this:

* Where will this script run? (Jenkins? CircleCI? somewhere else?)
* How often will it run? (you said "every workday" above, does that mean once/day? if it's just giving us notifications, should we run it more frequently?)
* What content will it run against? Fetched directly from Drupal? Running against what's live on the website?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,278 @@
...
+
+### Debugging
+
+- Local build logs
+- Jenkins build logs
+- Slack notifications
+
+### Caveats
+
+- The current broken link checking in Drupal can handle redirects, but those redirects are not currently synchronized with `vets-website`. When the redirects are not synchronized between `vets-website` and Drupal, a page that is moved in Drupal may pass the Drupal broken link check but fail the `vets-website` broken link check.
```
gunsch, line 229: what does it mean to say "can handle redirects, but those redirects are not currently synchronized with `vets-website`"?

* does "handle redirects" mean "follow all 301/302s on the live site"? (if so, I don't understand the "synchronized with vets-website" comment)
* does it mean "understands redirects configured within Drupal"? (if so, could you add that?)

gunsch (commented):
hey, apologies for the delay on getting to this. bunch of questions inline. I appreciate all the context, I feel like I understand the _current state_ pretty well from this, but I have a lot of questions about what you're proposing we move to next. let me know if they need any clarification!

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: It means that the redirects from drupal are not being imported into nginx during the content build. 

See the following quote from Steve W. for more info. 
> Where it still has the potential to fall down right now is that if a path to a page in drupal is changed, drupal automatically creates a redirect from the old path to the new path, so when the links are checked, they still work. The original plan was that the Front end would import redirects from drupal into Nginx when the content build takes place. However that is not happening, so the front end does not have the redirects, so a moved page, results in broken links from pre-existing content links. - https://github.com/department-of-veterans-affairs/va.gov-team/pull/9658#discussion_r434864015

I'll update this section to be more clear.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,278 @@
...
+
+Invalid content causes warning logs during staging deploys, and breaks the build during production deploys. These broken builds then need manual intervention, and require coordinating with the CMS team.
+
+##### Anybody that is trying to ship code for the VA.gov front end
+
+If invalid content is breaking the production build, and therefore blocking the production deploy, then no team on any contract is able to ship their frontend code to VA.gov.
+
+## High Level Design
+
+As part of the [build separation wor](https://github.com/department-of-veterans-affairs/va.gov-team/issues/2719)k, two separate pipelines will be created for for built code to be deployed to S3 to prevent failures in one from blocking between CMS and application deployments.
```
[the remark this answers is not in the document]
billfienberg: > wait, is this "two separate deploy pipelines" what you're proposing with this new design? or is this part of the context of other work already happening?

The two separate pipelines I'm referring to are part of the context of other work already happening. When I say "two separate deploy pipelines," I'm referring to the distinct content and application builds from [Chris' Separate Content Build design doc](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/design-docs/2020-04-09-separate-content-build.md#high-level-design).

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,278 @@
...
+
+The Metalsmith static content build will be extracted to the `content-build` repo and deleted from `vets-website`, and its deployment pipeline. Webpack will handle the application build for the `vets-website` repo, and Metalsmith will handle the static content build for the `content-build` repo.
+
+The validation of static content will take place in a scheduled job that runs every workday instead of ???. Since the broken link check happens during the Metalsmith build, and the accessibility check relies on the output of the content build, the scheduled job will run a script to produce a build that only has the steps necessary for content validation. That build will only be used for reporting broken links and accessibility errors, and will have no impact on deploys. Those reports will be delivered to the #cms-team channel via existing Slack integrations.
```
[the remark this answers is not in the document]
billfienberg: Because the content-only deploy, AKA [partial deploy](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/getting-started/workflow/deploy#partial-deploy--static-page-changes-only), only runs the drupal broken link checks, and none of the `vets-website` broken link checks, we've already indirectly decided that fast content deploys are worth the risk of broken links landing in production. 

In the **Alternatives** section, I've included a [**Highest quality, but lowest speed** option](https://github.com/department-of-veterans-affairs/va.gov-team/blob/cfae598310780e75243d3b5f6cbad8f1031079aa/platform/engineering/design-docs/2020-05-26-content-validation.md#highest-quality-but-lowest-speed).

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -139,19 +172,50 @@ If invalid content is breaking the production build, and therefore blocking the
...
+For anybody trying to ship code for the VA.gov front end, this approach will eliminate a source of failed application builds, which should allow them to ship code faster.
+
+For VA.gov users, this approach will continue allowing users to quickly see updated content. The downside of this approach for VA.gov users is that it won't eliminate existing opportunities for users to see invalid content. Fortunately, the improved error reporting should allow content writers to easily identify, remedy, and deploy valid content, which should minimize the user's chance of encountering invalid content.
+
+In summary, this approach will have the following impact for the following groups:
+
+- **Content Writers**: Improve error reporting while maintaining current convenience
+- **Frontend Tools team**: Reduce build noise, failure rate, and number of builds requiring manual intervention
+- **VA.gov users**: Maintain existing time to see updated content and risk of seeing invalid content
```
gunsch, line 193: hmm, is this true? my understanding was that broken link checking currently gates deployment, so if we move it to a daily job that _doesn't_ gate deployment, doesn't that increase risk of seeing invalid content?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -139,19 +172,50 @@ If invalid content is breaking the production build, and therefore blocking the
...
+
+#### Improved invalid content reporting
+
+Currently, accessibility errors need to be manually communicated from the Frontend Tools team to the CMS team. And broken links need to be extracted from the build log. We would improve that reporting by generating a file using Winston, and send that file to the #cms-team channel using existing Slack integrations.
+
+#### New scheduled job in Jenkins
 
-Lastly, we will write a scheduled Jenkins workflow to run the content validation script every workday at midnight.
+Lastly, we will write a scheduled Jenkins workflow to run the content validation script every workday at midnight Eastern. That way, the CMS team should have an accurate report of invalid content when they start their workday.
```
gunsch, line 218: how long do we anticipate this job taking? could we run it more frequently? (hourly? every 10 minutes?)

platform/engineering/design-docs/2020-05-26-content-validation.md
gunsch: Thanks for the extra detail! I feel much more able to understand your proposal now.

I didn't find an answer to this question of mine though:

> What content will it run against? Fetched directly from Drupal? Running against what's live on the website?

gunsch (commented):
couple last comments, but this read a lot clearer to me. thanks for the updates 👍

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: It looks like the existing `e2e` and `accessibility` steps, which run in parallel, take about 16 minutes to run. There's no reason I can currently think of why we couldn't run the job more frequently. I think hourly makes sense. Do you have any concerns about noise from an increased frequency?

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: Your understanding is correct. I apologize for the confusion. The current system does block deploys when broken links are detected in `master`. I've updated the design doc to clarify this in https://github.com/department-of-veterans-affairs/va.gov-team/pull/9658/commits/c3389e0b45f31d14079cc1a8d26a898d0a280de8.

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: > Thanks for the extra detail! I feel much more able to understand your proposal now.
> 
> I didn't find an answer to this question of mine though:
> 
> > What content will it run against? Fetched directly from Drupal? Running against what's live on the website?

@gunsch, sorry for missing that question. To answer your question, we will first try to pull the latest content from Drupal. If we are unable to pull the latest content from Drupal, we will use the cached Drupal content.

I've updated the design doc to include that answer in https://github.com/department-of-veterans-affairs/va.gov-team/pull/9658/commits/6077200b080141dd0b7946cd6060fb54ff9e42ee.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,378 @@
...
+
+## High Level Design
+
+Once the [build separation work](https://github.com/department-of-veterans-affairs/va.gov-team/issues/2719) is complete, there will be separate content and application builds. These distinct builds will prevent failures in one from blocking deployments to the other.
+
+The Metalsmith static content build will be extracted to the `content-build` repo and deleted from the `vets-website` repo and deployment pipeline. Webpack will handle the application build for the `vets-website` repo, and Metalsmith will handle the static content build for the `content-build` repo.
+
+The validation of static content will take place in a scheduled job that runs every workday, instead of happening as part of the full deploy. Since the broken link check happens during the Metalsmith build, and the accessibility check relies on the output of the content build, the scheduled job will run a script to produce a build that only has the steps necessary for content validation. That build will only be used for reporting broken links and accessibility errors, and will have no impact on deploys. Those reports will be delivered to the #cms-team channel via existing Slack integrations.
```
gunsch, line 209: > That build will only be used for reporting broken links and accessibility errors, and will have no impact on deploys.

is this still true?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,378 @@
...
+For anybody trying to ship code for the VA.gov front end, this approach will eliminate a source of failed application builds, which should allow them to ship code faster.
+
+For VA.gov users, this approach will continue allowing users to quickly see updated content. The downside of this approach for VA.gov users is that it won't eliminate existing opportunities for users to see accessibility errors. Fortunately, the improved error reporting should allow content writers to easily identify, remedy, and deploy valid content, which should minimize the user's chance of encountering invalid content.
+
+In summary, this approach will have the following impact for the following groups:
+
+- **Content Writers**: Improve error reporting while maintaining current convenience
+- **Frontend Tools team**: Reduce build noise, failure rate, and number of builds requiring manual intervention
+- **VA.gov users**: Maintain existing time to see updated content and risk of seeing accessibility errors
+- **Anybody trying to ship code for the VA.gov front end**: Reduce failed deploys which should allow them to ship faster
```
gunsch, line 224: given our conversations about broken link checking still blocking deploys after all... what builds will we no longer be failing?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,378 @@
...
+      - [Highest quality, but lowest speed](#highest-quality-but-lowest-speed)
+      - [Fastest speed, but lowest quality](#fastest-speed-but-lowest-quality)
+    - [Future Work](#future-work)
+    - [Revision History](#revision-history)
+
+## Overview
+
+### Objective
+
+The objective here is to define an updated approach to validate the CMS content. Specifically, this design doc focuses on identifying accessibility errors and/or broken links coming from the CMS content.
```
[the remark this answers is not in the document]
billfienberg: I made the following change to the objective in https://github.com/department-of-veterans-affairs/va.gov-team/pull/9658/commits/76febfe19d753aa2535ef231ad8e29d4d9359090. 

What do you think of this?
```
- The objective here is to define an updated approach to validate the CMS content. Specifically, this design doc focuses on identifying accessibility errors and/or broken links coming from the CMS content.
+ The objective here is to ensure that CMS content is still being validated once the `content-build` has been successfully extracted from the `vets-website` repo.
```

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: That is no longer true. Now that we have determined that broken links do block content only deploys, I have removed the mention of broken links. That build will now only be used for reporting accessibility errors, and that build will have no impact on deploys.

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: With this approach, the application build will no longer fail due to accessibility errors. 

Phrased another way:
**Current**: a11y content errors break the application build, but not the content build
**Proposed**: a11y content errors will not break the application build, and continue not breaking the content build

billfienberg:
@gunsch, after reading your comments, reading more code, and talking with Chris, I’ve made another round of updates to the design doc.

Now that we’ve learned that broken links do break content-only builds, the focus of this proposal can be narrowed from both broken links and a11y errors to just a11y errors.

Currently, a11y content errors can break the application deploy, but not the content-only deploy.

So the updated, simplified goal of this design doc is to run the content a11y checks in a scheduled job, so a11y content errors stop breaking the vets-website build, and improve the reporting of those a11y content errors so it’s more convenient for content writers.

platform/engineering/design-docs/2020-05-26-content-validation.md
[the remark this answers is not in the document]
gunsch: Awesome! Thanks, even just that small change helps me put the rest of the doc in a clearer context.

platform/engineering/design-docs/2020-05-26-content-validation.md
gunsch: The tradeoff seems to be "increased noise" vs "increased time to catch a Veteran-facing issue", right?

If that's the case, I'm inclined to lean more toward "increased noise". We can always come up with creative ways to deal with noise (i.e. some extra logic to avoid alerting multiple times about the same issue), not as much for user-visible accessibility issues on the site waiting to be caught by validation.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,388 @@
...
+
+#### New scheduled job in Jenkins
+
+Lastly, we will write a scheduled Jenkins workflow to run the content validation script hourly during workdays. That way, the content team learns about invalid content at a convenient cadence.
+
+### Code Location
+
+The broken link checks, axe checks, and scheduled job config will live in the [new `content-build` repo](https://github.com/department-of-veterans-affairs/content-build).
+
+### Testing Plan
```
gunsch, line 267: Btw, it's OK for testing plan to be much more high-level than this! This is extremely detailed, and I think just saying we'd:
* Validate that broken link checking still works
* Validate that accessibility validation still works
* Validate that accessibility validation on CI sends alerts to #cms-team 

is probably sufficient. (No need to change on this doc, just a note for future docs!)

platform/engineering/design-docs/2020-05-26-content-validation.md
gunsch: Re: whether or not this caveat needs to be here

Does this affect your design doc's plans in any way? It's true information, but I don't think it's relevant to the changes you're proposing in this doc.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,388 @@
...
+### Debugging
+
+- Local build logs
+- Jenkins build logs
+- Slack notifications
+
+### Caveats
+
+- The current broken link checking in Drupal can handle redirects, but those redirects are not currently synchronized with `vets-website`. When the redirects are not synchronized between `vets-website` and Drupal, a page that is moved in Drupal may pass the Drupal broken link check but fail the `vets-website` broken link check.
+- Many of the broken links come from URLs that change, according to decisions by the Content & IA team, usually in conjunction with Public Websites. In that workflow, an engineer makes a code change for the redirect, and has to time the release of the redirect code to match the content release of the URL change in the content layer. It's a fragile and very time-consuming process, and will likely not be scalable as we add hundreds of editors from outside the tight Github/Slack communication flows (e.g. Public Affairs Officers at 2000+ VA facilities).
```
gunsch, line 337: (nit) Also doesn't really feel like a caveat, maybe background info?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,388 @@
...
+
+#### Improved invalid content reporting
+
+Currently, accessibility errors need to be manually communicated from the Frontend Tools team to the CMS team. And broken links need to be extracted from the build log. We would improve that reporting by generating a file using Winston, and send that file to the #cms-team channel using existing Slack integrations.
+
+#### New scheduled job in Jenkins
+
+Lastly, we will write a scheduled Jenkins workflow to run the content validation script hourly during workdays. That way, the content team learns about invalid content at a convenient cadence.
```
gunsch, line 261: Can we run scheduled jobs in CircleCI? (i.e. what do we do post-Jenkins, since Ops team is tearing that down before too much longer?)

gunsch (commented):
Couple last comments/questions, but this seems in good shape to me overall. Thanks for the iterations!

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: Moved from caveat to background in https://github.com/department-of-veterans-affairs/va.gov-team/pull/9658/commits/152e818308af23b7148b4acbec1de630e52fe6fd

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: CircleCI does support scheduled workflows. 

https://circleci.com/docs/2.0/workflows/#scheduling-a-workflow

I actually attempted a prototype of a CircleCI scheduled job in a [branch](https://github.com/department-of-veterans-affairs/vets-website/compare/add-nightly-build-to-circleci) a couple months ago.

gunsch: great! is there a reason we wouldn't just implement this in CircleCI to start?

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: It doesn't affect the design doc's plans in any way. I've moved it from caveats to background in https://github.com/department-of-veterans-affairs/va.gov-team/pull/9658/commits/4efc42728ba68c5c467fded098d3791f73d635fa.

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: 🤔 

On one hand, I can't cite a concrete reason off the top of my head. 

On the other hand, I could see an argument that switching CI tools is slightly outside the scope/spirit of this design doc, and that task would be a good candidate for the **Future Work** section.

gunsch: my 2c: it seems like a strong likelihood that we're going to switch fully to CircleCI. so options are:
- implement this in Jenkins now, have to move it to CircleCI later
- implement in CircleCI, don't have to move later
if that's the case, should we just implement in Circle right away and save the trouble?

billfienberg: That makes sense. I've updated the Detailed Design to specify CircleCI instead of Jenkins in https://github.com/department-of-veterans-affairs/va.gov-team/pull/9658/commits/bdc89d7347e9ca7df9ac23524d52eaf848fafe3e.

gunsch (approved):

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,395 @@
...
+4. `src/site/stages/build/plugins/check-broken-links/index.js` calls the `getBrokenLinks` helper
+5. `src/site/stages/build/plugins/check-broken-links/helpers/getBrokenLinks.js` calls the `isBrokenLink` function
+6. `src/site/stages/build/plugins/check-broken-links/helpers/isBrokenLink.js` checks if an internal `href`/`src` value is included a specific array of paths
+7. The `check-broken-links` plugin calls the `getBrokenLinks` helper to build up an array of broken links
+8. The array of broken links from the `getBrokenLinks` helper is used by the `getErrorOutput` helper to build a CSV of broken links
+9. The `check-broken-links` plugin logs the broken links in the build console
+
+![screenshot of code path for check-broken-links Metalsmith plugin](https://user-images.githubusercontent.com/6130520/87714554-8849ef00-c771-11ea-8f20-ed52af7fcd3a.png)
+
+If there are broken links, the `glean-broken-links` script the build log, and create a file of broken links in a Comma-Separated Value (CSV) format. If broken links are found on the `master` branch, then an exception is thrown during the Jenkins job to block the deploy.
```
U-DON, line 141:
```suggestion
If there are broken links, the `glean-broken-links` script parses the build log and creates a file that lists the broken links in a Comma-Separated Value (CSV) format. If broken links are found on the `master` branch, an exception is thrown during the Jenkins job to block the deploy. 
```

meganhkelley:
Hey @billfienberg what else needs to happen here before this can be merged? Anything? Would like to get it into master so that it's in place with the other design docs in this folder https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/platform/engineering/design-docs

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,395 @@
...
+
+The Metalsmith static content build will be extracted to the `content-build` repo and deleted from the `vets-website` repo and deployment pipeline. Webpack will handle the application build for the `vets-website` repo, and Metalsmith will handle the static content build for the `content-build` repo.
+
+The broken link validation will continue using the existing the `check-broken-link` Metalsmith plugin and `glean-broken-links` script. Since broken links currently block content-only deploys, then we will continue blocking content-only deploys when broken links are detected.
+
+The validation of accessibility errors will take place in a scheduled job that runs hourly during workdays, instead of happening as part of the full deploy. Since the accessibility check relies on the output of the content build, the scheduled job will run a script to produce a build that only has the steps necessary for accessibility checking. That build will only be used for reporting accessibility errors, and will have no impact on deploys. Those reports will be delivered to the #cms-team channel via existing Slack integrations.
```
timwright12, line 225: I saw above that there's a plan to get a11y testing into the CMS, maybe it's worth accelerating that so the site doesn't risk falling out of section 508/WCAG AA compliance? I think code reviews should mitigate it from the apps end, but content I could see tripping up quite a bit - maybe if it's just trying to recommend a Drupal plugin or something (cc @meganhkelley)

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,395 @@
...
+
+### Open Questions and Risks
+
+#### Risks
+
+- Neither the current content validation nor the proposed content validation will **prevent** invalid content from making it to the live VA.gov site. But this proposal would prevent unrelated teams from being blocked by invalid content.
+
+#### Questions
+
+- What is the optimal frequency to run the scheduled job? Daily seems too slow. Every 10 minutes might be too fast for our builds. Hourly seems feasible. Is hourly acceptable? Should we be c concern about noise from a higher frequency?
```
timwright12, line 362: I think hourly is OK, but errors should pile up. Like, if one error reports and isn't fixed it should be reported again after an hour and keep reporting until it is fixed

platform/engineering/design-docs/2020-05-26-content-validation.md
meganhkelley: Accessibility checks for content on VA.gov are split over a variety of places — part of the CMS, part of automated tests on vets-website, and as manual reviews done before and after products and features launch. Features within the CMS itself would fall under purview of the CMS team — but we currently maintain the CMS build pipeline, which has some accessibility checks, which is why it came up in this design doc. 

In other words, let's chat more!

swirtSJW (approved):
I approve of this plan, but offer an additional improvement option for the future.  This design doc  
https://github.com/department-of-veterans-affairs/va.gov-team/pull/14092
proposes a way to process broken link errors after the fact and notify editors.  for immediate follow-up to rectify the situation, but not block timely content from going out.
It would be fairly easy to extrapolate that method to also read an accessibility failure report and process it in a similar way to notify the appropriate editors.

platform/engineering/design-docs/2020-05-26-content-validation.md
swirtSJW: I'd suggest that this should run with every content release.  Otherwise the reporting could be completely out of sync with what is actually out there and have editors thrashing to chase down reported issues they already resolved.

platform/engineering/design-docs/2020-05-26-content-validation.md
swirtSJW: The issue is most of the content from the CMS is in fields that lack the bigger html wrappers, so it can not be accurately tested since it is not rendered the same way it is by the  metalsmith build.  

There are some WYSIWIG instances in the cms but those are fairly restricted as far as what html tags people can use and what they can add.  It is not a guarantee that what comes out of them is accessible, but it is often the case that what may be needed for accessiblity may not be attainable in the wysiwig.    There is some scanning we could do and should do, but it will not be the panacea we want it to be.

platform/engineering/design-docs/2020-05-26-content-validation.md
timwright12: Yeah, I agree there, but I would also scan hourly in case there's a large gap in between content releases. So something doesn't sit there for too long and risk getting missed.

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,395 @@
...
+
+### Future Work
+
+- Delete the `accessibility` step of the `integration` stage from the `vets-website` Jenkins pipeline because that step will no longer be necessary in `vets-website`. It will no longer be necessary because the application accessibility checks are included in the tests run by the `e2e` step.
+
+### Revision History
+
+| Date         | Revisions Made | Author        |
+| ------------ | -------------- | ------------- |
+| May 26, 2020 | Initial draft  | Bill Fienberg |
```
mpelzsherman, line 395: There have been many commits to this document since May 26, so this section is misleading. Consider deleting this since the revision history can be obtained from git?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,395 @@
...
+
+#### Local
+
+##### Broken links
+
+###### Positive
+
+1. Confirm that the content build has zero broken links
+2. Run the content validation script
+3. Confirm that the script reports zero broken links
```
mpelzsherman, line 286: Aside from broken links, is there any plan to validate that the content is rendered correctly?

On the facilities team, we have started using [Cypress e2e tests](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/layouts/tests/vamc/smoke.mobile.cypress.spec.js) to validate our work.
Will this approach be supported going forward?

platform/engineering/design-docs/2020-05-26-content-validation.md
```diff
@@ -0,0 +1,395 @@
...
+1. Confirm that the content build has zero accessibility errors
+2. Modify an HTML file, like [`header.html`](https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/includes/header.html), to have an accessibility error
+3. Confirm that the script reports the newly introduced accessibility error
+
+#### CI
+
+##### Positive
+
+1. Confirm that the content build has zero accessibility errors and zero broken links
+2. Run the job that calls the content validation script
```
mpelzsherman, line 314: Aside from broken links, is there any plan to validate that the content is rendered correctly?

On the facilities team, we have started using [Cypress e2e tests](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/site/layouts/tests/vamc/smoke.mobile.cypress.spec.js) to validate our work.
Will this approach be supported going forward?

timwright12: @mpelzsherman nothing currently, but I'm happy to run/take a look at anything anyone has

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: I see what you mean. Initially, I included more entries in the revision history, but I removed them per the following PR comment:

>I think you should only need to add revision history after the document has been approved and merged. - https://github.com/department-of-veterans-affairs/va.gov-team/pull/9658#discussion_r433466403

platform/engineering/design-docs/2020-05-26-content-validation.md
billfienberg: The hourly idea came from the previous platform lead:

>how long do we anticipate this job taking? could we run it more frequently? (hourly? every 10 minutes?) - https://github.com/department-of-veterans-affairs/va.gov-team/pull/9658#discussion_r458442981

platform/engineering/design-docs/2020-05-26-content-validation.md
timwright12: we decided that since the a11y tests take 30mins to run, hourly didn't makes sense, so we're backing this off to daily

PR #10574, opened 2020-06-23
Changed page title and headings section

gmaciolek:
- Changed key point.
- Edited a bit to make it more concise.
- Tried to make it more consistent in the use of the terms "section" and "heading" 
- Added a bullet about using descriptive words in headings. 
 Question: Do we want to talk about SEO here? While that is important for page titles and headings in general, for our purposes, I think it's more important to emphasize the role of titles and headings in making the content focused, scannable, and clear.


kristinagore (approved):
Looks good!

PR #10692, opened 2020-06-26
Added edits and additional observers

jenniferlee-dsva:
Hi Liz - take a look at the edits. I've also added observers, some who were part of the original design studio workshop; not all of them will attend, but I want them on the original invites, so they can choose to attend. 

This is looking really good. 

There is one thing that Ryan and I def wanted to get user input on. It's a little granular, so maybe more for the conversation guide than the research plan. When users click on the 'View all articles...' link, we currently just take them to the search results page with a list of all of the LC articles tagged with that benefit category (and user may have to paginate if there are lots of articles in the future).

We'd love to learn if that's usable for users or if they expect/prefer/need to see a list of articles as a link farm - just a list of text links, all on one page. We didn't design for this because we just weren't sure what would be more useful to people. The concern is whether users find it hard to scan and find the for topic they're looking for when it's given as a search results display of "tagged" content.


andaleliz (approved):
These edits/additions look good to me. Thanks for adding them!

PR #10957, opened 2020-07-07
use Include to load custom VA ssh config as standalone file

mchelen-gov:
This PR does 2 things:
- use browser "save as" instead of copy+paste URL into CURL
- use `Include` in `~/.ssh/config` to allow standalone file


gunsch (approved):
drive-by approval, I like this change :+1:

mchelen-gov:
@gunsch thanks! i've been tinkering to see if it can be simplified/clarified any further :D

mchelen-gov:
spun off config script to https://github.com/department-of-veterans-affairs/va.gov-team/pull/11442

PR #10973, opened 2020-07-07
Update domo-monthly-report-survey-synthesis.md

joanneesteban:
@Kimberley2019 - I'd modify the synthesis summary. Looking at the data, 2/3 of the participants haven't actually used the contact center report at all and just glanced at it (assumptions here: they may be VSP or haven't had incentive to look at the data). The one person who actually uses the contact center report stated that there are pros/cons either way, but that Excel is easier to filter (which plays into our KPI dashboard research that folks need walk throughs of how to filter on Domo).


joanneesteban:
Also, given this is research based off a timeline, we should add the dates of this survey & a link of the survey to the Github doc.

PR #11376, opened 2020-07-17
Adds Issue Template for Initiative Dev Checklist

gunsch:
This includes:

* A deletable preamble explaining how it's intended to be used
* The checklist itself (last draft was still in [Google Docs](https://docs.google.com/document/d/1j_9zne5lEfXLjkkrxhcjL-Rg2-esIdGOtdp8C7yjoLI/edit?ts=5efdff9b&pli=1#))
* Checkboxes next to items people should complete
* "Required Artifacts" within each section for which pieces should be explicitly linked to show completion, or where the justification should live
* A number of TODOs for the process / documentation pieces we haven't clearly finished yet. I think this is okay for "usability testing" level but we should probably have something for each of those before launching this team-wide.

The "required artifacts" framing is a little new, since I didn't find something we had like this that felt clear to see where/how artifacts are provided. I'm not 100% in love with this solution but it felt the cleanest compared to interlacing artifacts with each bullet point or putting the artifacts all together at the end.


andreahewitt-odd (approved):
I think this looks great! Sorry for the lag in review. Hopefully we'll wrap up the remaining artifacts this week.

jchasia-adhoc (approved):

PR #11385, opened 2020-07-18
Updates Slack channels across documentation

gunsch:
Makes Slack channels across documentation meet the [style guide recommendations for Slack channels](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/documentation/style-guide/formatting-and-organization.md#slack-channels).

This isn't 100% coverage, but should be a vast enough majority of the instances to immediately shift what's "normal".

This was performed by a series of find/replace regexes. Specific regexes are in the commit messages of each commit making up this PR.


gunsch:
It's a bit hefty of a PR, but I've taken the time to go through file-by-file and make sure things look good. The find-and-replace approach definitely screwed a couple instances up, but I'm about 99% confident I found and fixed all those cases in self-review.

gmaciolek:
I spot-checked a few docs and it looks good to me. It seems like there are probably a few here and there that were missed? Like the design Slack channel link here: https://github.com/department-of-veterans-affairs/va.gov-team/tree/gunsch-slack-channels/platform/research but if this caught most of them, it's a good start.

gunsch:
Yeah, I'm expecting there are a few places I missed --- this wasn't intended to be 100% coverage but to catch the bulk of them with some find/replace logic. Thanks for looking!

PR #11580, opened 2020-07-23
use team mention instead of user assignment

mchelen-gov:
would this make the process easier or harder for @department-of-veterans-affairs/vsp-content-ia team?

### DIFF
--- a/.github/ISSUE_TEMPLATE/documentation-support-request.md
+++ b/.github/ISSUE_TEMPLATE/documentation-support-request.md
@@ -2,7 +2,6 @@
 name: Documentation Support Request
 about: Incoming documentation support requests triaged by VSP Content & IA team. 
 labels: content-ia-team, documentation-support
-assignees: gmaciolek
 
 ---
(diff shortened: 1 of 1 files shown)


gmaciolek:
Funny, I was in this issue template editing as you made this suggestion. I am trying to make this template generically useful for both VSP and VFS teams, so I don't think directing people to vsp-content-ia is quite the right approach here. vsp-content-ia is theoretically for VSP internal discussion and we direct VFS teams to the vfs-platform-support channel. However, I don't follow that channel closely and wouldn't notice a documentation question there unless some one @ mentioned me specifically. I'm going to update the template to say that the initial response time will be within 3 days, probably faster, and revisit things in a couple months to determine a smoother escalation path.

mchelen-gov:
@gmaciolek ah interesting, thanks for explaining! #vfs-platform-support is definitely helpful for VSF teams who have platform issues or questions. it does seem like there are times when questions are (thankfully) resolved but do not make it back into VSP docs. any thoughts you have about how to catch those would be great to hear!

PR #11833, opened 2020-07-29
add pact readme WIP

### LINKED ISSUE #12945: Review Pact docs for FE & BE Tools teams
### What is your documentation request, question, comment, or issue?

VSP FE & BE Tools teams are currently working on setting up Pact. Pact is an integration testing tool that will allow VFS teams to run automated tests on their apps, specifically to test the frontend of the app (the UI on VA.gov)'s integration with the backend. 

Lindsey and Eugene have drafted VFS-facing documentation about how to set up Pact. That can be seen in the `be-tools/pact-readme` branch: https://github.com/department-of-veterans-affairs/va.gov-team/blob/be-tools/pact-readme/platform/testing/integration-testing/how-to-use-pact-for-integration-testing.md

I reviewed and left some comments on the PR, many of which have now been addressed: https://github.com/department-of-veterans-affairs/va.gov-team/pull/11833

As I see it, there are 3 primary sections to this doc: introduction & details about Pact, how to set up the frontend, and how to set up the backend. We've talked about maybe splitting these out into separate docs, but I'm not sure whether that is necessary and am curious for y'alls documentation advice. 

We hope to begin testing Pact with VFS teams mid-next week — this will include getting feedback from engineers on the documentation (in other words, it doesn't need to be 100% perfect right now). I'm hoping to have another round of edits completed (from y'all) before we begin that testing.

@U-DON @LindseySaari @alexpappasoddball @rianfowler please add any details here as needed! Thanks all. 👍

LindseySaari:
Adds a WIP pact readme including implementation details and background info on PACT as a CDCT tool


teams/vsp/teams/tools/backend/pact/readme.md
```diff
@@ -1 +1,420 @@
...
+
+#### VCR
+Many of the vets-api endpoints call out to external services. To mock external service calls, the pact helper has VCR configured for usage. VCR cassettes can be used to mock external service responses for many of the vets-api third party service calls. The defined provider state in the search example from above makes use of a VCR cassette.
+
+### Configure the pact_uri/broker_url
+
+To work with one pact in the broker vs all pacts, point the `pact_uri` in the `pact_helper.rb` file to a specific contract in the broker (see [here](https://github.com/department-of-veterans-affairs/vets-api/pull/4612/files#diff-79dfb3765c26716457a005b9d343a160R38)), otherwise the rake task will run all the pacts pushed to the [heroku broker](https://vagov-pact-broker.herokuapp.com/).
```
U-DON, line 349: Should this section mention using [`rake pact:verify:at`](https://github.com/pact-foundation/pact-ruby/wiki/Verifying-pacts#verifying-a-pact-at-any-url-using-rake-pactverifyat)?

meganhkelley:
Hey y’all @U-DON @LindseySaari thanks so much for putting these docs together! I’ve made a few edits (style, formatting), and have some higher level comments & requests below. 

(1) I added a table of contents — wow! This is a very hefty doc. We don’t need to leave this full table of contents in the final doc (although we could), but it’s a good tool for us to evaluate the formatting. Does the current formatting of this doc make sense when y’all look at that ToC? Should this be broken out into multiple, smaller, cross-linked docs? Should the “how to set up” sections be way higher up and more the focus of the doc? Are things using the right header level? (Header level 5 ends up with a font size smaller than the body text which is not ideal — do we need to nest things that deep?) _I don’t feel like I have enough technical understanding to answer these Qs myself, although would be happy to have a work session with someone where we talk through this stuff, if that would be helpful._

(2) By `Requirements` do we mean this section describes integration test coverage required by VSP? If so let’s make that clear. If there’s a doc we can link to for `required code review process` here let’s do so.
* Check box when addressed [ ]

(3) In the `Broker` section, the `example setup instructions` link to HackMD — where are these supposed to link? Actually, a few other links throughout also link to HackMD (Pact example > Local json Files > details _and_ How to configure the provider codebase > quick reference checklist). Could someone please review the doc for lingering HackMD links? Thanks!
* Check box when addressed [ ]

(4) In `Implementation details` the link to `workflow details` goes here which is a really odd place to send people (the PR for a design doc). What are we trying to convey here? If linking to the design doc is the right call, could we please merge that design doc and link to the actual file? 
* Check box when addressed [ ]

(5) The `Pact example` setup instructions link goes to a really odd place as well — this seems to be a doc with some discovery info for Pact that’s on a branch in a folder that doesn’t make sense. If this is useful info that we want to reference, let’s move it to the `platform/testing/integration-testing` folder — either in a new doc, or we can move this whole doc over if we give it a better name. Also, we should always link to master wherever possible.
* Check box when addressed [ ] 

(6) Are the `Broker` and `Live Pacts` sections duplicates? Or are these conveying different info? 
* Check box when addressed (as applicable) [ ] 

(7) We refer to “the pact helper” a couple times, but never actually explain what this is, how to use it, etc. Is that implied, or does that info need to be called out? 
* Check box when addressed (as applicable) [ ]

U-DON:
I think it makes sense to split the FE and BE setup instructions into their own docs and cross link from here. It's likely there will be more content to add in both sections, and I agree the whole thing is already rather hefty as it is now.

There are probably a few things we could probably leave out, like the Pact Broker local setup. VFS teams using Pact should be able to adequately test their pacts by using the locally generated pacts and verifying locally. Only VSP will have to be concerned with maintaining the broker.

LindseySaari:
Thanks for the feedback @meganhkelley! Going to work through your feedback points above and push up changes. Thanks for pointing out that the links point to hackmd! Forgot to update those when doing a copy/paste from the original workspace.

LindseySaari:
@U-DON Any thoughts on this point? (4) In Implementation details the link to workflow details goes here which is a really odd place to send people (the PR for a design doc). What are we trying to convey here? If linking to the design doc is the right call, could we please merge that design doc and link to the actual file?

U-DON:
Maybe we could do without that section too,  or at least for now. From a VFS perspective, I think the reason people visit this doc would mainly be to figure out how to use Pact, and the design doc might be too dense to be useful to them.

PR #11855, opened 2020-07-30
Add mobile research guidelines

emilywaggoner:
Adding the first draft of mobile research guidelines for the research folder. What do you all think of the Pro/Con tables - they are formatted a little weird, should we keep them?

### DIFF
--- a/platform/research/planning/mobile-research-guidelines.md
+++ b/platform/research/planning/mobile-research-guidelines.md
@@ -28,7 +28,7 @@ This photo shows what it looks like:
 [Source, and more about the laptop hugging method](https://medium.com/@beparticular/were-still-hugging-our-laptops-8c7f22ed800e)
 
 Pros | Cons
------------- | -------------
+--------- | ----------
 Ability to see the full movement of the fingers | People without laptops or working web cam can’t use this method
 Relatively easy instruction and setup | The quality of the webcam could be blurry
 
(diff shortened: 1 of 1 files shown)


platform/research/planning/mobile-research-guidelines.md
```diff
@@ -0,0 +1,87 @@
...
+Ability to see the full movement of the fingers | People without laptops or working web cam can’t use this method
+Relatively easy instruction and setup | The quality of the webcam could be blurry
+
+#### Screensharing method
+The screensharing method requires the participant to download the Zoom app on their phone and share their screen as they complete the tasks of the usability study. 
+
+Pros | Cons
+--------- | ----------
+People without laptops or webcams can use this method | Moderators can not see hand movements
+x | Could be difficult for non-tech savvy people (there might have to be some reconfiguration in the settings section of the phone)
```
sshein, line 41: another "pro" is that Perigean has already helped folks run studies in this way, so it's been socialized with them already. That said, we could work with them on the laptop hugging method

sshein (approved):
looks good! I like the pros/cons tables. You could also put them as lists if the tables feel like too much to you

CrystabelReiter (approved):
This looks good! We could add after the line " Invision, UXPin, and staging are viable options for testing mobile flows on VA.gov" that "Note: Participants with assistive tech can only use staging for testing"

andaleliz (approved):
Looks good! Thanks for putting this together.

PR #11934, opened 2020-08-03
change to vagov for updated naming convention

jbritt1:
Working on updating the docs in https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/internal-tools.md and phasing out references to the old `vetsgov` naming convention. This PR updates key naming to newer version / matches to changes being made to the internal-tools doc.

Additional reference for transparency: https://github.com/department-of-veterans-affairs/va.gov-team/pull/11950

### DIFF
--- a/scripts/socks/README.md
+++ b/scripts/socks/README.md
@@ -19,7 +19,7 @@ Make sure to comment out any `IdentitiesOnly yes` directives set at the top leve
 Once this is saved, try the following commands:
 
 ```
-$ ssh-add -K ~/.ssh/vetsgov_id_rsa
+$ ssh-add -K ~/.ssh/*gov*
 $ ssh socks -D 2001 -N
 ```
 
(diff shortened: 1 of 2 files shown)


jbritt1:
Would also be open to wildcarding it so it works for those who already have this script as well as newcomers (i.e. `socks_add_key_cmd="ssh-add -K ${VA_SOCKS_KEYFILE:-~/.ssh/v*gov_id_rsa}"`

karlbrown-va:
I think maybe the wildcard (if it works) might be better for backwards compatibility?

jbritt1:
Tested locally:
```
> ssh-add -l
The agent has no identities.
> ll
total 16
drwxr-xr-x 2 jbritt jbritt 4096 Aug  3 14:30 .
drwxr-xr-x 3 jbritt jbritt 4096 Aug  3 14:30 ..
-rw------- 1 jbritt jbritt 1766 Aug  3 14:30 id_rsa_vetsgov
-rw------- 1 jbritt jbritt 1696 Aug  3 14:30 id_rsa_vtesttestgov.pem
> ssh-add id_rsa_v*gov*
Enter passphrase for id_rsa_vetsgov:
Identity added: id_rsa_vetsgov (id_rsa_vetsgov)
Identity added: id_rsa_vtesttestgov.pem (id_rsa_vtesttestgov.pem)
> ssh-add -l
2048 SHA256:ShkbdHKQqDwgONLv8/1qiYlX20kX9IPp3uV56ATp3c8 id_rsa_vetsgov (RSA)
2048 SHA256:5aOW1FonD9NDdZMgsCC5xK3s8rRAM0BvbvS5WOp+Pqw id_rsa_vtesttestgov.pem (RSA)
```

jbritt1:
Added wildcarding as well as flipped the order to match the current order in [the docs](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/internal-tools.md) (`id_rsa_vetsgov` vs `vetsgov_id_rsa`).

karlbrown-va:
I actually prefer the flipped version (e.g. vagov_id_rsa) - that way when you sort files in the directory they can be sorted by their domain, not by id_rsa.

also would likely need to update the readme, possibly - for this tool.

jbritt1:
I hear ya, but I would argue that most users aren't going to be as concerned with that. So I have shifted to using wildcards for the whole key name / structure to basically anything with "gov" in the name. Also updated the readme to follow this new pattern.

jbritt1:
Any other thoughts on this one @karlbrown-va ?

karlbrown-va:
hmm. I'm of two minds - part of me likes the flexibility, but it also means people who happen to have lots of other *gov* keys will get them all loaded. maybe this is a rare case - I'm not sure. I think let's go with it, and then if people complain we could change later.

karlbrown-va (approved):

PR #11975, opened 2020-08-04
Updates user guide Markdown

gunsch:
Minor updates to the user guides content:
* Updates one link to point to master
* Replaces pile of table HTML with Markdown table


gunsch:
Hey @gmaciolek --- this is basically my "review feedback" on your commit as a PR back at you.

I'm kinda curious in particular what you think of the Markdown (GFM) vs. HTML tables. The GFM tables forces these really long lines --- my research didn't turn up any ways to insert line breaks in the table --- but it has a ton less boilerplate HTML. Tradeoffs, I guess. Do we have a stance on this in the style guide?

gmaciolek:
@gunsch we don't have a stance on this in the style guide. I accidentally discovered that the HTML would work using a Markdown converter within Google docs--the converter output HTML for the table and I didn't know whether we had an opinion on that, so I tried it out and noticed that it solved the line break problem I've had when working with tables. I haven't been able to figure out how to do line breaks any other way. Do we want to keep the HTML? Format it without a table? Use a comma-separated list for the example links instead of line breaks?

gunsch:
idk! The line break thing is nicer, but on the other hand my table is ~6 lines compared to ~50 for the full HTML table.

the HTML gets even simpler if I replace the bulleted-list HTML with simple line breaks (`<br>`) instead.

after sitting on this a day, I think I'd like to merge this change, but I don't think I have a strong stance on this needing to be in the style guide. I'm happy to let things play out for a bit and see what people do for tables in other docs?

gmaciolek:
Yeah, just chalk this one up to my Markdown ignorance! I think people tend to format tables without HTML, so we probably don't need to add anything to the style guide urgently.

PR #12071, opened 2020-08-04
Rewrites logging README

gunsch:
* Simplifies using new Product Brief template
* Reframes around how to access/view logs

---

Hey TJ, I took a stab at what I'm envisioning this README could/should look like. (I redesigned the Product Brief template _after_ you filled it out, this is now in line with the new template). I tried to organize it less around the specific tool choice / remove "how it works" in favor of "how to use it". What do you think about this?

### DIFF
--- a/platform/logging/readme.md
+++ b/platform/logging/readme.md
@@ -1,34 +1,27 @@
-# Loki Product Brief (WIP)
+# Logging (WIP)
 
-### What is it?
+Logs on VSP are currently viewable in AWS Cloudwatch. For access, [file a ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/new/choose) using "Environment Access Request Template".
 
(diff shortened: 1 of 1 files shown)


ricetj (commented):
This template makes a lot more sense to me. It always felt a bit weird to put how it works in a customer-facing doc. The one question I do have is what do you envision these docs should look like?: 
- user docs about logging from FE + BE
- user docs about accessing view + logs 
Do you envision these being different docs and if so how?

ricetj (approved):

gunsch:
Yeah I do!

> user docs about logging from FE + BE

To me, these would be "how do you log things from FE" and "how do you log things from BE", both developer-focused docs

> user docs about accessing view + logs

Putting aside my incoherent language here, this to me would be more "how do you view / search logs in our logs UI" (Cloudwatch today, Loki tomorrow) --- probably more general audience than specifically developer focused

PR #12482, opened 2020-08-14
Mobile SSOe Design Doc

kreek:
This design doc lists options for how the mobile app could, via IAM's SSOe service, perform authentication in vets-api. There's also a  parent design doc detailing [why and where the mobile app is building on vets-api](https://github.com/department-of-veterans-affairs/va.gov-team/pull/12614).

PR #12614, opened 2020-08-19
Mobile API Design Doc

kreek:
This design doc has our initial thoughts on why and how we'd like to leverage vets-api to serve the mobile app's endpoints. There's also a companion design doc goes into lower-level details on [how the mobile app would perform authentication](https://github.com/department-of-veterans-affairs/va.gov-team/pull/12482).


platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
```diff
@@ -0,0 +1,173 @@
...
+will be in flux, and parallel va.gov changes could cause versions to climb.
+
+To limit the disruption to the API it's best to perform the optimizations in a module with a `/mobile` base path. 
+This allows us to:
+
+- Add mobile authentication without changes to the core API codebase.
+- Stay out of the way as we tweak endpoints in a mobile sandbox.
+- Test out optimizations that can be later ported back to the core API.
+
+Expanding on the last point it's possible that once the mobile optimized versions of the endpoints have stabilized that we permanently
```
gunsch, line 61: can you say more about API/endpoint versioning?

one of my biggest concerns is about mobile clients adding constraints to vets-api over time. mobile clients are likely to permanently lag behind in updates in ways that client JS does not. some questions that come to mind, though I think this is a bigger area to discuss than just these immediate questions:

* how will we know which endpoints/fields/etc. are depended on by mobile clients in the wild, and therefore can't (ever?) be changed?
* because of this concern, should mobile endpoints remain separated from web endpoints? why or why not? (I don't have a strong opinion here, but I think we should discuss pros/cons)
* how might a discovery API facilitate flexibility here?
* will we have a deprecation policy for mobile endpoints, slash, how out-of-date can a user's VA.gov mobile app be before we stop supporting it?
* follow-up: assuming we do eventually stop supporting old versions of mobile apps, will that include functionality in the mobile app to alert the user that they need to upgrade? (i.e. can we control that from an API endpoint of ours?)

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
```diff
@@ -0,0 +1,173 @@
...
+move them back to core API. The aim is to avoid confusion about why a pereceived newer version of an endpoint would be missing fields
+without context that it was actually a 'mobile only' endpoint.
+
+#### Application Controller
+
+Authentication for the mobile app is via IAM's SSOe OAuth service. It is 100% API based and does not use callbacks or cookies. 
+The mobile `ApplicationController` would extend the core `ApplicationController` and override its authentication methods.
```
gunsch, line 68: maybe rename to something with a different identifier? (`MobileApplicationController`?)

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
```diff
@@ -0,0 +1,173 @@
...
+
+```ruby
+mount Mobile::Engine, at: '/mobile'
+```
+
+Some endpoints such as facilities may not require optimization. They will pass-through via routes:
+
+```ruby
+# modules/mobile/config/routes.rb (redirect starts at the app root so this is /v0/mobile/.. pointing to )
+get '/v0/facilities/va/:id' => redirect('/v1/facilities/va/%{id}')
```
gunsch, line 86: to clarify, this would indicate that calls to `api.va.gov/v0/mobile/v0/facilities/va/:id` would auto-route to `api.va.gov/v1/facilities/va/:id`?

assuming yes, could you could talk about what kind of testing would ensure that a change to the `/v1/facilities/va/:id` endpoint output doesn't break mobile? (or rather, at least alerting developers that the proposed change affects mobile apps, not just website clients)

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
```diff
@@ -0,0 +1,173 @@
...
+or over-deliver, should implement fieldsets and includes in REST url params so no extra data is returned. JSON-API defines
+sparse [fieldsets](https://jsonapi.org/format/#fetching-sparse-fieldsets) and [includes](https://jsonapi.org/format/#fetching-includes) which
+can be implemented with custom code or through gems such as [JSON-API Resources](https://jsonapi-resources.com/) or [Graphiti](https://www.graphiti.dev/guides/).
+
+
+### Code Location
+The code will live in a 'mobile' directory within the [modules](https://github.com/department-of-veterans-affairs/vets-api/tree/master/modules) directory within vets-api.
+
+### Testing Plan
+To ensure changes to shared backend services will not break code, tests will be written to assert the mobile API requirements. Any breaking changes to upstream services will fail our tests.
```
gunsch, line 102: could you say more about this?

most of the `vets-api` developers will be thinking about the website primarily. how do we ensure that the output of mobile endpoints remains stable while development actively happens in the rest of the codebase? (i.e. the comment I left in your Routing section)

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
```diff
@@ -0,0 +1,173 @@
...
+### Security Concerns
+#### Potential threats
+
+##### SSOe introspect DoS
+Adding more surface area to the API does enable new possibilities for denial-of-service attacks. We plan on making a call to SSOe (introspect) with every vets-api session cache miss, to validate the access token and establish a session. An attacker could DoS the introspect endpoint by passing multiple requests all with invalid access tokens.
+
+Remediation could include caching each “miss” for a period of time, protecting the introspect endpoint.
+
+##### CSRF for API POST requests
+We know we can disable CSRF checking on the server-side for the mobile POST routes, but that seems unwide.
```
gunsch, line 122: it should be OK to only enforce CSRF prevention when cookie-based authentication is used

if you're authenticating with a mobile auth token (presumably passed via HTTP Header? I haven't read the other design doc yet), then CSRF prevention isn't necessary

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
```diff
@@ -0,0 +1,173 @@
...
+
+#### Path and Protocol Alternatives
+Once we decided on building within vets-api we analyzed our options for how to do so. The short term solutions were to use 
+existing endpoints, version endpoints and make changes there, or create a mobile base path (module) so the mobile routes and 
+auth are isolated. Below is a table listing the sizes for; Lift, or the estimated effort to implement a solution. Latency, 
+the number of calls and size of the payload per mobile screen. Flexibility, how easy it is to make near and long term changes. 
+Ubiquity, the penetration/adoption rate of a technology. And Disruption, how new a technology is to VA and likelihood of 
+requiring a new ATO.
+
+<img src="images/mobile-api/api-decision-matrix.png" alt="API Decision Matrix" width="650" />
```
gunsch, line 156: (lol, this table is tough to comment on)

I take mild issue with the "Existing Endpoints" x "Latency" cell --- partly because I think you have overlapping concerns with similar web UIs. if Profile UI takes three calls to get all the data you need... well, that means the web UI takes three calls too and would probably _also_ benefit from optimizing those.

but I agree with the other concerns, you're sending more data than necessary (and by a bit), and being able to more seriously consider options like protobuf optimization in the future could be a big win

gunsch (commented):
thanks for sharing an awesome design doc! couple comments of some areas to dig into more, but overall I think this technical direction should work well for the team. (that said, take the other reviewers' word more seriously than mine on that 😉 )

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
jholton, line 122: typo?
```suggestion
We know we can disable CSRF checking on the server-side for the mobile POST routes, but that seems unwise.
```

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
```diff
@@ -0,0 +1,173 @@
...
+new mode of communication to allow Veterans to interact more easily with the department. Since VA.gov is powered by 
+vets-api and VSP, it seemed like a natural fit for the mobile app to be powered by APIs in the same space.
+
+### High Level Design
+Since vets-api already contains integrations to backend services that the mobile app will need 
+(such as MVI, Claims, Appointments), the mobile API will be built inside of the vets-api codebase, and exposed in the same 
+ way that VA.gov API endpoints are exposed.
+
+The mobile API will provide functionality that is currently exclusive to VA.gov as well as some already-public functionality 
+from Lighthouse, so existing integrations will be reused. A Rails engine will house the mobile API surface area, with necessary
```
jholton, line 34: > A Rails engine will house the mobile API surface area

Does this mean all calls from the mobile client will be made to endpoints in the `mobile` Rails engine?

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
jholton: I think the `redirect` here returns a 301 to the client, which would then result in the mobile client making a subsequent request directly to `/v1/facilities/va/%{id}`, bypassing the `mobile` module endpoint(s). I'm not sure if that's the desired behavior.  Please correct me if I'm wrong.

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
```diff
@@ -0,0 +1,173 @@
...
+- Stay out of the way as we tweak endpoints in a mobile sandbox.
+- Test out optimizations that can be later ported back to the core API.
+
+Expanding on the last point it's possible that once the mobile optimized versions of the endpoints have stabilized that we permanently
+move them back to core API. The aim is to avoid confusion about why a pereceived newer version of an endpoint would be missing fields
+without context that it was actually a 'mobile only' endpoint.
+
+#### Application Controller
+
+Authentication for the mobile app is via IAM's SSOe OAuth service. It is 100% API based and does not use callbacks or cookies.
```
jholton, line 67: Just curious, do these SSOe OAuth sessions overlap with "standard" SSOe sessions?
If a user logs-in on the app, then opens a browser and logs in via the website, are there two active SSOe sessions for that user from the SSOe service's perspective?

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
kreek: @gunsch answers from @jjulian and myself for the above questions:

> How will we know which endpoints/fields/etc. are depended on by mobile clients in the wild, and therefore can’t (ever?) be changed? - because of this concern, should mobile endpoints remain separated from web endpoints? why or why not? (I don’t have a strong opinion here, but I think we should discuss pros/cons)

Mobile clients will only access endpoints under `/mobile`. Optimizations could find their way back to the core API but any changes there would not need to take the mobile app(s) into account.

As we have more discussions about the mobile app it’s becoming clear that it’s key difference from va.gov will be that it’s a distilled experience. e.g. You’re not going to fill out a multi page 526 form on the mobile app, but you will get updates as it moves through the system. This difference in how the data is consumed is why mobile apps would always use `/mobile`.

**Pros**:

- Web and mobile endpoints can serialize customized data for their views
- There will be mobile only endpoints, all the mobile endpoints will live in one place
- Easier to document the mobile API as separate entity
- Less version creep, each client manages breaking changes independently
- More consistent in the core API (versions naturally evolve, a v1 mobile endpoint doesn’t live between web’s v0 and v2)
- Easier to move the mobile API out of vets-api completely should the need arise

**Cons**:

- Although routes and controllers will be unique, services will be shared and must not break wither contract.
- More tests! Vets-api takes a while to run its test suite, will be more than if they shared endpoints.
- There’s now three entry points to the API core, lighthouse, and mobile

> How might a discovery API facilitate flexibility here?

Because mobile apps are less frequently updated (if at all for some users) it will definitely use a discovery endpoint which accepts an app version and returns a directory of routes. Similar to how [Github’s](https://api.github.com/) does. It could also help in a shared scenario mapping routes for both web and mobile. Assuming web wants a discovery API and the added complexity of looking up its routes.

> Will we have a deprecation policy for mobile endpoints, slash, how out-of-date can a user’s VA.gov mobile app be before we stop supporting it? - follow-up: assuming we do eventually stop supporting old versions of mobile apps, will that include functionality in the mobile app to alert the user that they need to upgrade? (i.e. can we control that from an API endpoint of ours?

We think deprecation requires more analysis. For discussion’s sake, let’s say we release 1.0.0 in Nov, 1.1.0 in Jan, and 1.2.0 in April. There will most likely be several point releases between each of these. We don’t imagine setting a hard line when an app version becomes unsupported, but as time goes on I’m sure we’ll notice trends. For example, if we are on 2.1.0, and there is a burden to retaining the 1.0.0 functionality in one or more of our /mobile endpoints, then we can choose a sunset date for 1.0.0 depending on the user base and activity of that version’s cohorts. Although React Native (using Expo) allows for over-the-air code updates, we have no plans to use that feature. However, the discovery endpoint can be used to inform an app user that their experience will be diminished and they should attempt to upgrade to 1.1.0.

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
kreek: Yes @jholton is correct it is 301 which is not how we want the route to work. Instead we'll have our own route definitions, controllers, and serializers. There would be some code duplication. Specs would guard services from changing out from under us.

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
kreek: We're now thinking of having our own routes and controllers and only sharing services. Request specs with [JSON schemas](http://json-schema.org/) would protect mobile responses from core API service changes.

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
kreek: yes

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
kreek: I'll convert this to a list of alternatives and why we didn't choose them. GraphQL and gRPC can go in future work as they weren't really options for MVP. Agree that the profile optimization should be API wide. Mobile still might have a different serializer if it didn't need all the fields.

platform/engineering/design-docs/2020-08-17-mobile-api-design-doc.md
gunsch: Thanks for the answers, that all sounds great. Can you work those into the design doc for posterity?

re: `/mobile` endpoints, SGTM 👍 

re: discovery API, 👍 

re: deprecation policy:

> the discovery endpoint can be used to inform an app user that their experience will be diminished and they should attempt to upgrade to 1.1.0.

cool, I think this is the part I was hoping for in particular

gunsch:
@kreek  since you merged this --- In terms of getting this to approval, do you want to continue discussion on this closed PR? or are you planning to open a follow-up PR?

kreek:
@gunsch we can continue the discussing here. Sorry that merge was more to finalize the areas we weren't sure about initially 😄 It's in its proposed form now. i.e. we'll build in an engine, @ `/mobile` and will share services but not controllers with the core API.

PR #12696, opened 2020-08-21
Misc. updates

gmaciolek:
I made a variety of changes, summarized below. (Feel free to disagree with any of it!) As for where to put the doc, it seems to me like you have it in a logical spot, but  I'm not that familiar with this particular folder and it seems like the README is out of date, so I'll check in with @emilywaggoner and @CrystabelReiter 
- Added a few introductory sentences to tee up the purpose of the doc
- Added/tweaked the headings so that the headings provide an outline of the document for easier scanning
- Shortened and simplified sentences 
- Adjusted capitalization in a few places(I struggle the most with being consistent on this, so if the capitalization still seems arbitrary, it might be!)
- Removed the Path A, B, C designations for the different options. I'm assuming people wouldn't actually reference the options by letter and thought they might be more confusing than helpful. If you had plans for the letter options, though, we can work them back in.

### DIFF
--- a/platform/design/design-system/governance-process/edge-cases.md
+++ b/platform/design/design-system/governance-process/edge-cases.md
@@ -22,11 +22,11 @@ Edge cases arise when, in step 3 of the happy path, teams can't isolate the requ
 ## Edge case governance process
 
 ### For changes that are launch-blocking
-**If it would impact multiple features/teams:** 
+**If it would impact multiple features/teams:** \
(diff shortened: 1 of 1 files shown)


gmaciolek:
I thought of another question: Is there an official/preferred way to escalate to a VSP DEPO lead?

rroueche (approved):

PR #12768, opened 2020-08-24
VACMS-2297 Add cms redirects design doc.

swirtSJW:
Design doc submitted for review following https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/platform/engineering/design-docs

### DIFF
--- a/platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
+++ b/platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
@@ -1,0 +1,224 @@
+# Front End Build Redirects from CMS
+
+_Replace the previous line with the the title of your project or component and replace the following lines with your name(s), email and the date._
+**Author(s):** Steve Wirt
+**Last Updated:** Aug 21, 2020
+**Status:** Draft | **In Review** | Approved
+**Approvers:**
+   - [ ] Rian Fowler
+   - [ ] Andrew Gunsch
+   - [ ] Mikki Northuis
+   - [ ] Nick Sullivan
+   - [ ] Taylor Skinner (originally worked on redirect import)
+   - [ ] ??
+
+## Overview
+
+
+### Objective
+The CMS allows editors to create pages and move pages to other locations within the VA.gov website.  The CMS keeps track of when a page's path is changed and builds a redirect from one to the other.  The problem is, there is nothing that moves the redirects known within the CMS to the public website. It is in the best interest of editors, developers and veterans to have the front-end website know about and act on the redirects known to the CMS.
+
+### Background
+If pages are moved within the CMS under the right conditions the end result is a combination of internal broken links that currently cause failure of CMS content to be released to the front-end, and 404 errors for veterans.
+
+Within the CMS we have put in a lot of [safeguards to detect broken links](https://www.drupal.org/project/node_link_report) within content at the instant of saving a page of content and warning the edtitor in real time.  There are limits to this method however as it can not account for editors changing the location of a page being linked to or the published status of the link target at the time of content release (CR).
+
+| Situation | Current Result | Result if Redirects as Content |
+| --------- | -------------- | ------------------------------ |
+| Link correctly created in CMS to page in CMS that has been moved | 200 | 200 |
+| Link created by url copy paste in WYSIWG to page in CMS that has been moved | 404 & CR Fail | 200|
+| Link from VA.gov page not in CMS to CMS page that has been moved | 404 & CR Fail | 200 |
+| Link from external site to page in CMS that has been moved | 404 | 200 |
+| Browser bookmark / email link to page in CMS that has been moved | 404 | 200 |
+
+The current state of managing redirects for content, involves content editors and a developer coordinating and choreographing the release of the contnet with the addition of the redirect(s).  It is time intensive and risk prone.  VA needs a redirect system that is easier to manage and less reliant on developers timing releases perfectly to match.
+
+### Guiding Principles
+1. Veterans should not get 404 errors.
+2. Content on VA.gov should not have broken links to content also on VA.gov.
+3. As the CMS scales from ~20 editors and daily content releases, up to ~500 editors and continuous content releases, we can not allow a broken link to prevent content releases.
+4. Developers currently waste far too much time chasing down broken links.
+
+
+### High Level Design
+
+Three parts of the system work together:
+
+#### I. CMS
+At the highest level, the CMS manages redirects already.  It tracks them, sorts out and prevents circular references (infinite redirects), creates new redirects for moved pages and can handle query parameters, if needed for legacy handling. The CMS even provides a nice user interface for adding, deleting, and editing source, destination and status codes.  The CMS currently makes redirects available to both GraphQL and the static file content exports.  Only one minor change is needed on the CMS side to translate node/# paths to the proper alias.  We simply need a way for the Front-End to act on the redirects in the CMS. The CMS treats redirects like content, and the Front-End should too.  When a title changes, in most cases, the URL changes, so a redirect is needed.  Any Front-End system that does not account for that is a 404 factory, which is our current condition.
+
+#### II. Vets-website content creation.
+Local environments that are not AWS based would need to function correctly with automated tests and human users.  So one part of this strategy is to create client side redirects by creating an html file with a meta refresh tag
+`<meta http-equiv="refresh" content="0; URL='/new/path/destination'" />` for each redirect object found in the content export.  This will happen at the same time that content html pages are built.
+
+#### III. Vets-website deploy to AWS
+During the vets-website deploy to AWS S3, the system currently generates an object for each static page of the VA.gov website.  Via the same object creation process we can add an [S3 object with redirect metadata](https://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html) that describes both the destination and the return code for each HTML page that contains a meta-refresh tag. These S3 objects get discarded and rebuilt on each content release, so they would **always** be in sync with the content from the CMS.  By adding these, the website when running in an AWS environment would have the more desireable server side redirects instead of the browser-based `meta-refresh` redirects.
+
+## Specifics
+
+### Detailed Design
+
+#### I. CMS
+The redirects are output in the content export like this:
+```
+{
+    "rid": [
+        {
+            "value": 494
+        }
+    ],
+    "uuid": [
+        {
+            "value": "3d72f51f-fb7a-4cce-a16e-0771505e552a"
+        }
+    ],
+    "hash": [
+        {
+            "value": "NSUuoJQb5WlkVtO2TgzqvnRzwOvwR-ZX9U7rztw5v8w"
+        }
+    ],
+    "type": [
+        {
+            "value": "redirect"
+        }
+    ],
+    "uid": [
+        {
+            "target_type": "user",
+            "target_uuid": "fd71292f-3dc7-409d-b0e8-500204cb005a"
+        }
+    ],
+    "redirect_source": [
+        {
+            "path": "pittsburgh-health-care\/volunteer-or-donate",
+            "query": []
+        }
+    ],
+    "redirect_redirect": [
+        {
+            "uri": "internal:\/node\/281",
+            "title": "Server",
+            "options": []
+         }
+    ],
+    "language": [
+        {
+            "value": "en"
+        }
+    ],
+    "status_code": [
+        {
+            "value": 301
+        }
+    ],
+    "created": [
+        {
+            "value": "2019-11-26T18:44:28+00:00",
+            "format": "Y-m-d\\TH:i:sP"
+        }
+    ],
+    "metatag": {
+        "value": {
+            "title": "Edit URL redirect | Veterans Affairs",
+            "twitter_cards_type": "summary_large_image",
+            "og_site_name": "Veterans Affairs",
+            "twitter_cards_title": "Edit URL redirect | Veterans Affairs",
+            "twitter_cards_site": "@DeptVetAffairs",
+            "og_title": "Edit URL redirect | Veterans Affairs"
+        }
+    }
+}
+
+```
+
+The only modification needed at the CMS export is to add a destination element that would take drupal node paths
+```
+ "redirect_redirect": [
+        {
+            "uri": "internal:\/node\/281",
+            "title": "Server",
+            "options": []
+         }
+```
+
+and translate `"uri": "internal:\/node\/281",` to an actual path like
+```
+"destination": [
+        {
+            "uri": "pittsburgh-health-care\/about\/volunteer-or-donate",
+         }
+```
+This would be a trivial to add.
+
+#### II. Vets-website content creation.
+  _(Covered in High Level Design section.)_
+
+#### III. Vets-website deploy to AWS
+  _(Covered in High Level Design section.)_
+
+
+### Code Location
+
+Step I: Addition of relative path for destination would be added to the [CMS repo here](https://github.com/department-of-veterans-affairs/va.gov-cms/blob/master/docroot/modules/custom/va_gov_content_export/src/TomeExporter.php#L97)
+
+Step II: Build the HTML files with meta refresh during content file generation in the [vets-website repo here](https://github.com/kevwalsh/vets-website/blob/1d6a53cd9b53ec71b69ae2bce78bd5694eaf1d7f/src/site/stages/build/drupal/metalsmith-drupal.js#L51).
+
+Step III:  Building the AWS S3 objects with redirect metadata would happen in the [Devops repo here](https://github.com/department-of-veterans-affairs/devops/blob/a58dfbc762466905d5727cb4cd626d5e6dad9c3f/ansible/deployment/roles/deploy-vets-website/files/vets-website-deploy.sh)
+which is called from [here](https://github.com/department-of-veterans-affairs/devops/blob/6fd9fc8f5ebcb2b685e4e0a20450c472655fcef2/ansible/deployment/roles/deploy-vets-website/tasks/main.yml#L2-L7)
+
+
+### Testing Plan
+The current content release method already does link checking.  This method would be tested to a large degree by this existing test.  The CMS already does a significant amount of testing of a front end build in a non-AWS environment so we would add a Behat test to makes sure the that the client side redirect is working.
+
+Additional testing could be added on the Vets-website deployment tests to verify that a set of redirects work properly on the server side redirects in AWS.
+
+### Logging
+Logging would not be required.  There may be some statistics that could be tracked within AWS, but that may not be necessary.  Google Analytics may also be used to track instances of displaying the 404 page that may in some cases include referral data.  Server logs track HTTP requests and are likely already used to track requests ending in 404 errors.
+
+### Debugging
+CURL would be the simplest and most useful debugging method of looking at and debugging redirects on any of the BRD environments as well as local environments.  On local environments simply inspecting a page's meta refresh tag would provide debug specific data.
+
+### Caveats
+Section level redirects that are outside of the CMS could still be handled as they are at the server or F5 level.
+
+### Security Concerns
+None at this time.
+
+### Privacy Concerns
+None at this time.
+
+### Open Questions and Risks
+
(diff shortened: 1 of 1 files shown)


platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
```diff
@@ -0,0 +1,224 @@
...
+   - [ ] ??
+
+## Overview
+
+
+### Objective
+The CMS allows editors to create pages and move pages to other locations within the VA.gov website.  The CMS keeps track of when a page's path is changed and builds a redirect from one to the other.  The problem is, there is nothing that moves the redirects known within the CMS to the public website. It is in the best interest of editors, developers and veterans to have the front-end website know about and act on the redirects known to the CMS.
+
+### Background
+If pages are moved within the CMS under the right conditions the end result is a combination of internal broken links that currently cause failure of CMS content to be released to the front-end, and 404 errors for veterans.
```
gunsch, line 22: what are "the right conditions" that cause this? is this the table you have below?

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
```diff
@@ -0,0 +1,224 @@
...
+   - [ ] Taylor Skinner (originally worked on redirect import)
+   - [ ] ??
+
+## Overview
+
+
+### Objective
+The CMS allows editors to create pages and move pages to other locations within the VA.gov website.  The CMS keeps track of when a page's path is changed and builds a redirect from one to the other.  The problem is, there is nothing that moves the redirects known within the CMS to the public website. It is in the best interest of editors, developers and veterans to have the front-end website know about and act on the redirects known to the CMS.
+
+### Background
```
gunsch, line 21: can you also include in Background:

* how do redirects currently get to the FE website? I don't think "nothing" is the current answer, my understanding is it's just manual, and this doc should describe that process. (we're wanting to improve the process, rather than to build a process where one doesn't exist, right?)

* description of the redirect review process, i.e. how does IA get involved and review URL redirects today?  (and, in the rest of the doc, how would your proposed accommodate this process?)

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
```diff
@@ -0,0 +1,224 @@
...
+4. Developers currently waste far too much time chasing down broken links.
+
+
+### High Level Design
+
+Three parts of the system work together:
+
+#### I. CMS
+At the highest level, the CMS manages redirects already.  It tracks them, sorts out and prevents circular references (infinite redirects), creates new redirects for moved pages and can handle query parameters, if needed for legacy handling. The CMS even provides a nice user interface for adding, deleting, and editing source, destination and status codes.  The CMS currently makes redirects available to both GraphQL and the static file content exports.  Only one minor change is needed on the CMS side to translate node/# paths to the proper alias.  We simply need a way for the Front-End to act on the redirects in the CMS. The CMS treats redirects like content, and the Front-End should too.  When a title changes, in most cases, the URL changes, so a redirect is needed.  Any Front-End system that does not account for that is a 404 factory, which is our current condition.
```
gunsch, line 48: > The CMS treats redirects like content, and the Front-End should too.

what does it mean when you say the FE should "treat redirects like content"?

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
```diff
@@ -0,0 +1,224 @@
...
+Local environments that are not AWS based would need to function correctly with automated tests and human users.  So one part of this strategy is to create client side redirects by creating an html file with a meta refresh tag
+`<meta http-equiv="refresh" content="0; URL='/new/path/destination'" />` for each redirect object found in the content export.  This will happen at the same time that content html pages are built.
+
+#### III. Vets-website deploy to AWS
+During the vets-website deploy to AWS S3, the system currently generates an object for each static page of the VA.gov website.  Via the same object creation process we can add an [S3 object with redirect metadata](https://docs.aws.amazon.com/AmazonS3/latest/dev/how-to-page-redirect.html) that describes both the destination and the return code for each HTML page that contains a meta-refresh tag. These S3 objects get discarded and rebuilt on each content release, so they would **always** be in sync with the content from the CMS.  By adding these, the website when running in an AWS environment would have the more desireable server side redirects instead of the browser-based `meta-refresh` redirects.
```
gunsch, line 55: does this match how we serve content from S3? could we actually use this redirect functionality on S3, and would that work with our revproxy setup?

cc @cvalarida @billfienberg while Rian's out

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Specifics
+
+### Detailed Design
+
+#### I. CMS
+The redirects are output in the content export like this:
```
gunsch, line 62: this format seems to assume that all paths (source and destination) are simply paths on the same domain, but my understanding is that we handle redirects across VA.gov domains. can you speak to that?

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Work Estimates
+1. CMS: Add destination path to CMS content export. (~3 story points)
+2. PWS: Add build of HTML pages that have meta-refresh along side the current HTML page builds for each content export redirect.*.json file.  These create the client side redirects for use in non-AWS environments. (~3 story points)
+3. PWS Devops : Add processing to vets-website deploy to create an AWS Object with redirect metadata for any HTML page that contains a meta-refresh tag. This would create server side redirects for any AWS environments. (~10 story points)
+
+### Alternatives
+
+#### Import redirects from CMS into nginX or F5
+This [epic was planned and then dropped](https://github.com/department-of-veterans-affairs/va.gov-team/issues/503).
```
gunsch, line 202: I heard a rumor that this approach was shot down because having Drupal as central authority doesn't allow a place to inject any review of URLs being used for redirects --- that any content editor can create URL redirects without oversight. is this true? if so, how does this approach address that?

gunsch (commented):
thanks for writing up this design doc! some high-level feedback + questions from my first read

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
swirtSJW: Yes that is covered by the conditions in the table.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
swirtSJW: These are parts that I do not know.  They currently do not get from the CMS to the front end by any automated method.     I can consult with @ncksllvn as I know he has been involved with setting up some of the redirects that currently exits.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
swirtSJW: I mean that when content gets updated, redirects should get updated.  When the title of a page in the cms changes, its path changes.  So the redirects that Drupal already builds in the CMS, should get built on the front end too at the time of content release.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
swirtSJW: Those would still be handled wherever they are now  (either at the nginx or F5 level.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
mnorthuis: Current redirect process is documented here: https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/information-architecture/request-redirect.md

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
mnorthuis: We should be able to decouple URL changes from H1 changes. H1s may evolve and be tweaked but URLs are not directly tied to them for core benefit content.  This approach may be different for other types of content (i.e. news articles) but that needs to be defined yet.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
mnorthuis: There are templates/content types that oversight is required (i.e. core benefit content, VAMC location pages), there are other cases where we may let them happen organically (i.e. new articles)

swirtSJW: I did not see that rumor listed on the epic that was closed.  Drupal can have all kinds of workflows.  If there is a moderation of redirects that is needed, it could be added.  However, there is limited oversight now if an editor changes the name on a page that page gets moved to a different path. For some editors and some pieces of content (depending on its owner) the change of title goes through editorial review, while others can just make the change and publish it.  Either way,  they have likely created a 404.  

Other than the automated creation of redirects stated above, the other CRUD operations  of redirects are permed and currently assigned  to roles in the CMS 
image

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
mnorthuis: Does the CMS only automatically build the redirect when the node ID is used?  Also, sort of related to this process, how do we ensure all links across the site are updated appropriately?

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
mnorthuis: Does the CMS eliminate multiple jumps or is that something that needs to be manually managed?  In other words, if a URL A changes to B and A is redirected, then B changes to C, we want the original A URL to redirect to C, not B and then C.  Multiple jumps can hurt our SEO value.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
mnorthuis: Would like to still have a support capability within Drupal to find all links and update them when a URL outside the CMS is changed.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
```diff
@@ -0,0 +1,224 @@
...
+
+
+### Testing Plan
+The current content release method already does link checking.  This method would be tested to a large degree by this existing test.  The CMS already does a significant amount of testing of a front end build in a non-AWS environment so we would add a Behat test to makes sure the that the client side redirect is working.
+
+Additional testing could be added on the Vets-website deployment tests to verify that a set of redirects work properly on the server side redirects in AWS.
+
+### Logging
+Logging would not be required.  There may be some statistics that could be tracked within AWS, but that may not be necessary.  Google Analytics may also be used to track instances of displaying the 404 page that may in some cases include referral data.  Server logs track HTTP requests and are likely already used to track requests ending in 404 errors.
+
```
mnorthuis, line 177: This may not be related to logging per se, but we need to be able to manage the redirects over time as they age out, so we aren't managing thousands of redirects over time.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
mnorthuis: Please double check this, as far as I am aware, an H1 change does not always result in an automatic URL change in certain content templates - namely benefit content.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
gunsch: I don't think we can just automate redirect publishing if it discards the IA and approval processes. Can you speak to where IA approval processes might fit in if we automate this?

I realize this question's a little more product-y than technical since it's more about overall user stories --- but I'm worried this technical proposal changes the user stories in ways that aren't actually OK. curious for your thoughts! (or alternately, is there a PM on your team who's looking at this from the user-stories angle?)

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
cvalarida: Right now, we're using `aws s3 sync` to sync up the contents of the build directory in Jenkins to the S3 bucket. I don't know offhand how we'd manage the creation, updating, and removal of these redirects, but it seems feasible, given the documentation Steve links to.

This definitely needs an experiment run to verify we can do what we'd need to.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
meganhkelley: +1 to these Qs from Mikki and Andrew about how the governance and oversight process will work. Are there user roles in the CMS that could support that (redirects could be limited to certain user roles)? How else might we preserve a redirect governance model? 

I understand if the governance is not your expertise, Steve! Happy to direct these Qs to the PM for this work, just let me know who I should be talking to.

meganhkelley:
Here is another relevant redirect doc that Mikki, Wyatt, and Nick and I wrote up a few months ago — not sure if it's out of date by now, but adding here as context: https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/redirect-implementation-strategy.md

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
```diff
@@ -0,0 +1,224 @@
...
+| Link from external site to page in CMS that has been moved | 404 | 200 |
+| Browser bookmark / email link to page in CMS that has been moved | 404 | 200 |
+
+The current state of managing redirects for content, involves content editors and a developer coordinating and choreographing the release of the contnet with the addition of the redirect(s).  It is time intensive and risk prone.  VA needs a redirect system that is easier to manage and less reliant on developers timing releases perfectly to match.
+
+### Guiding Principles
+1. Veterans should not get 404 errors.
+2. Content on VA.gov should not have broken links to content also on VA.gov.
+3. As the CMS scales from ~20 editors and daily content releases, up to ~500 editors and continuous content releases, we can not allow a broken link to prevent content releases.
+4. Developers currently waste far too much time chasing down broken links.
```
meganhkelley, line 40: Thanks for adding these guiding principles! These also may be a good to start to defining what success looks like for a redirect strategy (which is more of a product question than the engineering design doc usually strives to answer, but still something I'm thinking about).

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
swirtSJW: The CMS aggregates multiple jumps to 1 jump.  They are server side redirects so they would always only be one jump. It would have only a positive impact on SEO, as it would greatly reduce our 404's

We should be able to decouple title (H1) changes to the path, but that is now how things are set up now and would make management of urls confusing if they were decoupled.   This is something that Drupal has handled pretty seamlessly for a decade and a half.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
swirtSJW: They would not need to be removed, as both html pages and these redirects would all be created anew on each release.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
swirtSJW: Right TItles  in Drupal (H1 are up to the front end) are not tied to paths on all content types. They are on several.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
swirtSJW: Any redirects known to drupal can be manged in a nice searchable interface
image

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
mnorthuis: Decoupling H1s and URLs is something we have been talking about for the past year with Kevin W.  In some cases we need them decoupled to keep URLs cemented in place for SEO and communication purposes...in some cases, we may need to set up a separate field for a shortend title/description to cap characters and drive a URL...and in other cases we may have the H1s tied directly to the URLs.  Each of these is a different product and we should be able to define the right way to handle it based on business processes and access levels.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
rianfowler: [HTML redirects](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#Alternative_way_of_specifying_redirections) should work with the current nginx setup. I don't think headers pass through but this would be inside of the body which nginx only changes to support CSP. Either way we could do a simple test on staging by adding a document with this metatag.  

It's preferable to use an HTTP redirect but I don't think there's a tangible difference. Users will get a 200 instead of a 301 but the browser will treat it like a permanent redirect. 

This approach doesn't couple us to S3 since it's just creating an html document with a metatag in it.

rianfowler (commented):
From an engineering perspective, I think this is preferable to the current approach. I think we should add validation to the build process to detect and fail on circular redirects and add a section to this document mentioning that.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
swirtSJW: @rianfowler returning misleading headers is not great for SEO. It would also make it impossible for our own link checkers (both on front end and inside the CMS) to surface whether something is pointing to a redirected page, it would CURL the page to see if it exists, it would get a misleading 200 response.  
We spent a lot of time building the functionality to detect redirects and warn editors
image
This shortcut would prevent that from working.

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
swirtSJW: > Does the CMS only automatically build the redirect when the node ID is used? 

@mnorthuis If the path of node/123 gets change from `/here` to `/there` A redirect gets built in the CMS.  If the link in the content is done via the best practice in the CMS then the is updated automatically to show that it is pointing to `/there`.  If however the link is created just in the wysiwyg as an html path then it would still point to `/here`, resulting in a 404 on the front end if we do not have redirects being built with the content releases. 

> Also, sort of related to this process, how do we ensure all links across the site are updated appropriately?

If this proposal were implemented, the link to the redirected page would show up in our node link report that shows on each node for editors.  It would show as a redirected link with a suggestion to fix it.
image

platform/engineering/design-docs/2020-08-11-redirects-from-cms.md
mnorthuis: A 200 response will impact SEO as the browser and search engine think all is well, but we need search engines to know that a page has been permanently redirected so that equity is passed on and SERPs are updated.   We need 301s to retain the SEO value of pages.

jsev-io (approved):

jbritt1:
@swirtSJW, is this work still relevant / good to merge?

meganhkelley:
@jbritt1 this work is still relevant, but this design doc has not been approved by anyone yet so it is not ready to merge

jbritt1:
> @jbritt1 this work is still relevant, but this design doc has not been approved by anyone yet so it is not ready to merge

Understood, thank you for the additional information @meganhkelley!

drorva:
What's the priority of this? @meganhkelley if it's high, I can carve some time to review and push it forward.

jbritt1:
This has become quite stale, can we close this until the work need/prio comes around?

swirtSJW:
Thanks @jbritt1 I agree this is stale, but it is still needed and is only going to become more critical as more editors are onboarded to CMS.    However this may be impacted and or resolved by the new accelerated publishing prioritization.  @mchelen I am going to let you advise on where this should sit in terms its current need and where it may fit with Accelerated Publishing.

jhouse-solvd:
@swirtSJW / @drorva - could we possibly mark this as a draft until we prioritize and confirm that it's ready for review? 

That has the positive impact for Ops (et al) of the following: 

"Draft pull requests cannot be merged, and code owners are not automatically requested to review draft pull requests."
_more info here:_ https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests

Thanks in advance for your consideration.

PR #13301, opened 2020-09-09
Updates to deployment policy

gunsch:
Proposing the following updates to the deployment policy, based on Friday's incident:

* Removes some outdated / non-policy bits near the end
* Adds introductory stanza about our overall stance
* Adds section about resolving production bugs, prioritizing reverts, not rolling back

### DIFF
--- a/platform/working-with-vsp/policies-work-norms/deployment-policy.md
+++ b/platform/working-with-vsp/policies-work-norms/deployment-policy.md
@@ -1,5 +1,13 @@
 *Last updated 2020-01-27*
 
+# Deployment Policies
+
+## Deployment Strategy
+
+We intend to keep our repositories **always deployable from master branches**. The deployment schedule below describes when deployments typically happen, but deployments from master branches may occur at any time. VA.gov does not use release branches; instead, we maintain a focus on keeping the master branches healthy.
+
+For finer control over when your checked-in feature is released, see [feature toggles](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/platform/tools/feature-toggles/).
+
 # Automated Deploys
 
 There are automated deploys of `vets-api` and `vets-website` Monday through Friday. The release is created from `master` at 2 pm Eastern and deployed at 3 pm Eastern. Code merged to `master` after 2 pm Eastern will not be added to a release and deployed until the next business day.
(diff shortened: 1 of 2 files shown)


platform/working-with-vsp/policies-work-norms/deployment-policy.md
```diff
@@ -58,16 +66,12 @@ When in doubt on whether an issue is _critical_ enough for out-of-band deploymen
 1. Once approved, contact the [VSP DevOps oncall](https://dsva.pagerduty.com/schedules#PGIEA8Q) through the [#oncall channel](https://dsva.slack.com/archives/C30LCU8S3) to coordinate the release.
 1. Within two business days of this incident, send a PR with a [postmortem](https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/tree/master/Postmortems), including any relevant stakeholders, VSP DEPO leadership, and any VSP team members involved in resolving the incident as reviewers.
 
-## Step-by-step process for deployment
-
-1. To deploy this change to production, the release job is run with the git SHA for the commit to deploy.
-1. The auto-deploy boolean flag must be set to `FALSE` on the job.
-1. The release tag is captured and then used as input for the matching deploy job.
-1. There are "special" jobs for `vets-website` and `vets-api` which will run both the release and deploy of `master` in one operation.
+## Resolving production bugs
```
gunsch, line 69: curious for y'all's thoughts on location on this section. I'm pretty sure this is absolutely the *wrong* place for this information, but I don't have a better place easily handy. (_possibly_ the Incident Response playbook, but that might be a little out of the way for this information? I'm open to ideas.)

rianfowler, line 69:
```suggestion
## Resolving post deploy issues 
```
This is tricky. We want to resolve issues as quickly as possible but I don't think reverting is the right call for lower priority production bugs.

rianfowler: This seems fine to me for now.

rianfowler (approved):

platform/working-with-vsp/policies-work-norms/deployment-policy.md
dginther: I would say this could go in the Engineering SOPs

gunsch: Hmm, I'd like these policies to remain public if possible, and `devops` isn't.

I've tried moving this to a new location, under a doc called [resolving critical issues](https://github.com/department-of-veterans-affairs/va.gov-team/blob/gunsch-deploy-policy/platform/engineering/resolving-critical-issues.md), adding some context, and linking between this doc and that one:

that way, this doc can remain focused on _deployments_, and the other doc can talk about how we fix issues or not. lemme know what you think.

gunsch: I ended up changing this heading entirely after separating it to the "resolving critical issues" doc. it's not really about "production" or "post deploy" anymore.

If we want the guidance to be "keep master deployable", to me that's equivalent to "keep master free of critical issues" (where "critical issues" is defined in this doc already). so I think the guidelines on how to fix (fix in five minutes, or revert) apply whether or not the issue has reached production.

philipwritescode (approved):

PR #14092, opened 2020-09-28
VACMS-2417 Add broken link design doc.

swirtSJW:
Design doc for consideration as a way to handle broken link issues in the future.

Here is the formatted view of the MD https://github.com/department-of-veterans-affairs/va.gov-team/blob/7785e85e3c68452857ec510ed5d2cb22bbd5f6cc/platform/engineering/design-docs/2020-09-27-broken-links.md

I have included both technical people on this doc as well as stakeholders because the approach depends on the commitment to not let timely content be blocked by the presence of a broken link.

### DIFF
--- a/platform/engineering/design-docs/2020-09-27-broken-links.md
+++ b/platform/engineering/design-docs/2020-09-27-broken-links.md
@@ -5,7 +5,7 @@
 **Status:** **Draft** | In Review | Approved
 **Approvers:**
    - [ ] Rian Fowler
-   - [ ] Andrew Gunsch
+   - [ ] Michael Fleet
    - [ ] Mikki Northuis
    - [ ] Nick Sullivan
    - [ ] Dave Conlon
--- a/platform/engineering/design-docs/2020-09-27-broken-links.md
+++ b/platform/engineering/design-docs/2020-09-27-broken-links.md
@@ -15,7 +15,7 @@
 ## Overview
 
 ### Objective
-To address broken links in a way that notifies the human source quickly, clearly, and without human intervention so that the link may be resolved.  If we can get broken links resolved rapidly, and we are doing rapid content releases every 3-5 minutes, we should not be holding up content releases for broken links.  Alerts, situation updates, and all other content should not be held hostage by a broken link somewhere in the website.
+To address broken links in a way that notifies the human source(s) quickly, clearly, and without human intervention so that the link may be resolved.  If we can get broken links resolved rapidly, and we are doing rapid content releases every 3-5 minutes, we should not be holding up content releases for broken links.  Alerts, situation updates, and all other content should not be held hostage by a broken link somewhere in the website.
 
 ### Background
 Broken links discovered during the Front End content build currently prevent content releases.  Notifications are sent to Slack #CMS-Team.  This pulls CMS team members away from their planned work, to triage the situation.  This includes:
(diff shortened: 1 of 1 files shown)


platform/engineering/design-docs/2020-09-27-broken-links.md
```diff
@@ -0,0 +1,123 @@
...
+     2. link text  {'click here'}
+     3. link destination {'path/to/page-not-found`}
+     4. uuid - a hash of page path, link text, link destination
+     5. First appeared (timestamp) - The time that this particular UUID first appeared in the CSV.
+     6. Drupal Page (boolean) - TRUE if it is a CMS based page
+     7. Escalated to Section (timestamp) - Will be set when section owners have been notified.
+     8. Escalated to CMS Team (timestamp)  - Will be set when the CMS team has been notified.
+     9. Resolved (timestamp) - Will be set when this no longer appears in a broken link CSV.
+
+   b.  The CMS will look up the page or menu item with the broken link.  If it is determined to be a CMS page or menu item, it will send a notification within the CMS to the last person who edited that page to that user's dashboard and send an email to that user using the notification queue.  If the same link text and broken destination appears on multiple pages, only one notification will be sent, to avoid sending a barrage of identical notifications (as would be the case of a broken menu item).
```
kaise-lafrai, line 69: This should start with c.

platform/engineering/design-docs/2020-09-27-broken-links.md
```diff
@@ -0,0 +1,123 @@
...
+   c. If the same page has a broken link for 3 separate content releases  _(~15 min span of time)_ notify other content owners responsible for the same section.
```
kaise-lafrai, line 70: This should start with d.

platform/engineering/design-docs/2020-09-27-broken-links.md
```diff
@@ -0,0 +1,123 @@
...
+    d. If the same page has a broken link for 5 separate content releases _(~25 min span of time)_, or there are no other editors for that section, alert Slack #CMS-team so the team can escalate the fix.
```
kaise-lafrai, line 71: This should start with e.

kaise-lafrai (commented):
Just a couple small ordering issues.

platform/engineering/design-docs/2020-09-27-broken-links.md
```diff
@@ -0,0 +1,122 @@
+# FE Broken Link Notifications -> CMS -> Empowered People
+
+**Author(s):** Steve Wirt
+**Last Updated:** September 27, 2020
+**Status:** **Draft** | In Review | Approved
+**Approvers:**
+   - [ ] Rian Fowler
+   - [ ] Andrew Gunsch
```
rianfowler, line 8:
```suggestion
   - [ ] Michael Fleet
```
Michael Fleet has taken over for Andrew as the VSP engineering lead

platform/engineering/design-docs/2020-09-27-broken-links.md
```diff
@@ -0,0 +1,122 @@
...
+    2. If the same page has a broken link for 5 separate content releases _(~25 min span of time)_, or there are no other editors for that section, alert Slack #CMS-team so the team can escalate the fix.
+3. If it is not a CMS based page, notify the appropriate Slack channel based on the location of the broken link.
+4. Sections that are under construction (hidden behind basic auth) could be disregarded so broken links in that section would be ignored by the CMS to avoid creating extra noise.
+
+## Specifics
+
+### Detailed Design
+There are three phases of changes that would need to be implemented across the two systems in the following order.
+
+1. Front end build makes broken link CSV public.  The CSV already gets built, it just needs to be saved as a public file.  (https://www.va.gov/data/vagov-broken-links.csv)
```
rianfowler, line 54: I think if this process is not going to block content deployments because of broken links, then the build process doesn't have to be part of this validation. The current production html can be synced and run against an independent link checker like [broken-link-checker](https://www.npmjs.com/package/broken-link-checker) or [link-checker](https://packagist.org/packages/glicer/link-checker) and the validation can be triggered by successful deployment or on a schedule. 

What do you think?

mnorthuis:
Question from the non-developer ;-)  - why does a broken link in one node or page fail an entire build?  Isn't there a way to 1) identify the broken link at the time someone publishes a page and notifies them that their page cannot be moved to published until its fixed  2) just cause the offending page to fail and not the entire build?

platform/engineering/design-docs/2020-09-27-broken-links.md
swirtSJW: That's an interesting point.  Would this suggestion gain us speed improve the timing? It would have to be fast, and the broken link file that is exposed would have to match the current release.  Otherwise it could create wild goose chases for a broken link report that is out of sync with the actual files.
I'd rather not increase the complexity and change the current method if it doesn't gain us anything.    Right now the csv is produced, it just needs to be made public.  If platform wants to swap one link checker method for another that would pose no problem as long as that method could be in sync with the current release and have the csv ready when the front end release pings drupal to day the release is complete.

swirtSJW:
> Question from the non-developer ;-) - why does a broken link in one node or page fail an entire build? Isn't there a way to 1) identify the broken link at the time someone publishes a page and notify them that their page cannot be moved to published until its fixed 2) just cause the offending page to fail and not the entire build?

Good question @mnorthuis 
I think the best reason for not publishing is,  the broken link may not be the fault of that page.  It may have been the destination that was removed or changed path.  It may not be a cms based page (many of our more current broken link incidents are not related to the CMS.  We do the notification of broken links at time of saving the page, however in some cases we get false positives (false failures) because the CMS may not have access to certain domains. since it is behind the VA network.    Also it creates a chicken and an egg situation if two page are interlinked and neither are published.  One has to be published first and if a broken link prevents publishing the node, then the two pages are deadlocked.

The front end could elect not to push a page live if it contained a broken link... but that would create a cascade effect on all pages that point to it... and all pages that point to those page.... and all pages that point to those.  Pretty soon we'd have no pages that were published.

As long as we continue to make a broken links more important (blocking) than content going out on time, we end up with content that doesn't go out on time.

platform/engineering/design-docs/2020-09-27-broken-links.md
rianfowler: >It would have to be fast, and the broken link file that is exposed would have to match the current release.

Can we add more information about matching the current release to the design doc?

platform/engineering/design-docs/2020-09-27-broken-links.md
```diff
@@ -0,0 +1,122 @@
...
+
+
+### Guiding Principles
+1. Veterans should not get 404 errors. Content on VA.gov should not have broken links.
+2. As the CMS scales from ~20 editors and daily content releases, up to ~500 editors and continuous content releases, we can not allow a broken link to prevent content releases.  Knowing about a storm closure or active shooter [saves more souls](https://www.informationweek.com/government/dont-get-side-tracked-by-flashy-metrics-success-is-in-the-people/d/d-id/1331791) than not getting a 404 when following a link.
+3. Developers currently waste far too much time chasing down broken links.
+
+
+### High Level Design
+Currently the Front End link checker creates a CSV file and exposes that to Jenkins.  The CSV contains the location of the broken link, the url it is pointing to.  I propose publishing that file to the website as open data (perhaps located at /data/vagov-broken-links.csv or as json).  Currently the Front End build process notifies the CMS when a content release has succeeded by pinging the CMS  The CMS could act on that ping as a trigger and read the broken link data.  It could process it immediately and take the following actions:
```
billfienberg, line 41: I'm not certain, but we might have to find a different namespace other than `/data`. 

Currently, https://staging.va.gov/data redirects to https://www.data.va.gov/.

billfienberg:
@swirtSJW, here's a draft PR to make the broken link CSV publicly available:
https://github.com/department-of-veterans-affairs/vets-website/pull/15036

swirtSJW:
Machine readable json now available here https://s3-us-gov-west-1.amazonaws.com/vetsgov-website-builds-s3-upload/broken-link-reports/vagovprod-broken-links.json

PR #14616, opened 2020-10-08
Make proxy.pac compatible with Safari

ericboehs:
Safari didn't like the regex domain matching.

Also, it wanted `SOCKS` instead of `SOCKS5` (via [SO](https://apple.stackexchange.com/a/396773)).

### DIFF
--- a/scripts/socks/proxy.pac
+++ b/scripts/socks/proxy.pac
@@ -1,9 +1,14 @@
 function FindProxyForURL(url, host) {
-  PROXY = "SOCKS5 127.0.0.1:2001"
+  PROXY = "SOCKS 127.0.0.1:2001"
 
-  if (shExpMatch(host,"*.vfs.va.gov|*.vetsgov-internal")) {
+  if (shExpMatch(host,"*.vfs.va.gov")) {
     return PROXY;
   }
+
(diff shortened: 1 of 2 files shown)


scripts/socks/proxy.pac
```diff
@@ -1,9 +1,14 @@
 function FindProxyForURL(url, host) {
-  PROXY = "SOCKS5 127.0.0.1:2001"
+  PROXY = "SOCKS 127.0.0.1:2001"
```
ericboehs, line 2:
```suggestion
  PROXY = "SOCKS5 127.0.0.1:2001;SOCKS 127.0.0.1:2001"
```

karlbrown-va:
Have you confirmed this still works correctly with Chrome and Firefox?

Also, you'll need to update the readme, as it says that Safari doesn't work.

thanks!

ericboehs:
@karlbrown-va Looks like Chrome/Firefox needed `SOCKS5`. I've updated the `PROXY` to include both and from my testings on macOS 10.15.7, it works with Chrome 85 & 86, Firefox 81, and Safari 14.

karlbrown-va (approved):
ok, I tested this and it works for me as well.

PR #14627, opened 2020-10-08
Updates to socks.sh script and associated instructions

karlbrown-va:
* Update internal-tools doc to clarify that socks.sh and SwitchyOmega are alternatives
* Update socks.sh to avoid the use of *gov* which ended up causing issues with the script;
the problem is, it would try to load public keys as well, and fail. Thus, it
has been changed to reference id_rsa_vagov, which is the suggested keyname.
It is also possible to override the keyname, and instructions are given
in the script for these cases.

### DIFF
--- a/scripts/socks/README.md
+++ b/scripts/socks/README.md
@@ -19,7 +19,7 @@ Make sure to comment out any `IdentitiesOnly yes` directives set at the top leve
 Once this is saved, try the following commands:
 
 ```
-$ ssh-add -K ~/.ssh/*gov*
+$ ssh-add -K ~/.ssh/id_rsa_vagov
 $ ssh socks -D 2001 -N
 ```
 
@@ -31,7 +31,9 @@ $ ./socks.sh     # check status
 $ ./socks.sh off # turn off socks proxy
 ```
 
-The script assumes you have stored your key using "gov" somewhere within name (`~/.ssh/*gov*`). If you put it elsewhere or name it otherwise, you can set the VA_SOCKS_KEYFILE variable before invoking it. The script also sets up a proxy for the "Wi-Fi" network by default; if you are using a different network, you can override the SOCKS_NETWORK environment variable - for example:
+The script assumes you have stored your key at (`~/.ssh/id_rsa_vagov`). If you put it elsewhere or name it otherwise, you can set the VA_SOCKS_KEYFILE variable before invoking it (like this: `VA_SOCKS_KEYFILE=~/.ssh/my_vagov_privatekey`).
+
+The script also sets up a proxy for the "Wi-Fi" network by default; if you are using a different network, you can override the SOCKS_NETWORK environment variable - for example:
 
 ```
 $ SOCKS_NETWORK="USB 10/100/1000 LAN" ./socks.sh on
@@ -40,7 +42,6 @@ $ SOCKS_NETWORK="USB 10/100/1000 LAN" ./socks.sh off
 
 The socks script will start the socks proxy, configure the proxy settings on your computer to use a `proxy.pac` file, and start a small webserver to serve that file up. The proxy.pac file will send only a few URLs to the proxy, sending everything else directly to the internet.
 
-Finally, you'll need to configure your browser to use the system proxy if you use this script. This has been tested on OSX 10.14
-and 10.15, and works with Safari, Firefox, and Chrome.
+Finally, If you want to use a browser other than Chrome or Safari, you'll need to configure your browser to use the system proxy. This has been tested on OSX 10.14 and 10.15, and works with Chrome, Firefox and Safari. Chrome and Safari use the system proxy by default; you can set it up for Firefox by following [these instructions](https://support.mozilla.org/en-US/kb/connection-settings-firefox).
 
-If you don't want to use the proxy.pac route, another option is to install a proxy switcher into your browser - [details here](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/internal-tools.md#chrome--firefox).
+If you don't want to use this approach, another option is to install a proxy switcher into your browser - [details here](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/engineering/internal-tools.md#chrome--firefox).
--- a/scripts/socks/socks.sh
+++ b/scripts/socks/socks.sh
@@ -18,8 +18,8 @@ AUTOCONFIG_PORT=9500
 # from https://stackoverflow.com/a/246128/4907881
 PROXY_PAC_LOCATION="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
-socks_add_key_cmd="ssh-add -K ${VA_SOCKS_KEYFILE:-~/.ssh/*gov*}"
-socks_del_key_cmd="ssh-add -d ${VA_SOCKS_KEYFILE:-~/.ssh/*gov*}"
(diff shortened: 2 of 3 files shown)


scripts/socks/README.md
```diff
@@ -40,7 +42,6 @@ $ SOCKS_NETWORK="USB 10/100/1000 LAN" ./socks.sh off
 
 The socks script will start the socks proxy, configure the proxy settings on your computer to use a `proxy.pac` file, and start a small webserver to serve that file up. The proxy.pac file will send only a few URLs to the proxy, sending everything else directly to the internet.
 
-Finally, you'll need to configure your browser to use the system proxy if you use this script. This has been tested on OSX 10.14
-and 10.15, and works with Safari, Firefox, and Chrome.
+Finally, If you want to use a browser other than Chrome or Safari, you'll need to configure your browser to use the system proxy. This has been tested on OSX 10.14 and 10.15, and works with Chrome, Firefox and Safari. Chrome and Safari use the system proxy by default; you can set it up for Firefox by following [these instructions](https://support.mozilla.org/en-US/kb/connection-settings-firefox).
```
ericboehs, line 45: I didn't have to configure Firefox in a special way. It just worked like Chrome and Safari.

scripts/socks/README.md
```diff
@@ -31,7 +31,9 @@ $ ./socks.sh     # check status
 $ ./socks.sh off # turn off socks proxy
 ```
 
-The script assumes you have stored your key using "gov" somewhere within name (`~/.ssh/*gov*`). If you put it elsewhere or name it otherwise, you can set the VA_SOCKS_KEYFILE variable before invoking it. The script also sets up a proxy for the "Wi-Fi" network by default; if you are using a different network, you can override the SOCKS_NETWORK environment variable - for example:
+The script assumes you have stored your key at (`~/.ssh/id_rsa_vagov`). If you put it elsewhere or name it otherwise, you can set the VA_SOCKS_KEYFILE variable before invoking it (like this: `VA_SOCKS_KEYFILE=~/.ssh/my_vagov_privatekey`).
+
+The script also sets up a proxy for the "Wi-Fi" network by default; if you are using a different network, you can override the SOCKS_NETWORK environment variable - for example:
```
ericboehs, line 36: Instead of defaulting to `Wi-Fi`, I think we should default to `networksetup -listallnetworkservices | head -2 | tail -1`. This list is sorted by service order so it gets the primary connection. They could of course still override it using the `SOCKS_NETWORK` env var.

This is probably out of scope of the PR though.

scripts/socks/socks.sh
```diff
@@ -18,8 +18,8 @@ AUTOCONFIG_PORT=9500
 # from https://stackoverflow.com/a/246128/4907881
 PROXY_PAC_LOCATION="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
ericboehs, line 19: Also out of scope, but instead of having to host this locally, we could just directly reference the proxy.pac file on GitHub: https://raw.githubusercontent.com/department-of-veterans-affairs/va.gov-team/master/scripts/socks/proxy.pac.

scripts/socks/README.md
karlbrown-va: Hmm - ok that's interesting. This may vary based on machine, version of firefox, or other system settings. I think since Firefox (of the 3) is the only one that actually has it's own proxy settings in the control panel, it's worth having the link there nonetheless

scripts/socks/README.md
karlbrown-va: when I run that comment I end up with USB 10/100/1000 LAN, so that wouldn't work for my normal day to day. I think 'Wi-Fi" is probably the most common connection modality for people using laptops.

scripts/socks/socks.sh
```diff
@@ -18,8 +18,8 @@ AUTOCONFIG_PORT=9500
...
 
-socks_add_key_cmd="ssh-add -K ${VA_SOCKS_KEYFILE:-~/.ssh/*gov*}"
-socks_del_key_cmd="ssh-add -d ${VA_SOCKS_KEYFILE:-~/.ssh/*gov*}"
+socks_add_key_cmd="ssh-add -K ${VA_SOCKS_KEYFILE:-~/.ssh/id_rsa_vagov}"
+socks_del_key_cmd="ssh-add -d ${VA_SOCKS_KEYFILE:-~/.ssh/id_rsa_vagov}"
```
ericboehs, line 22: Since I've [setup ssh to use the keychain](https://apple.stackexchange.com/a/250572), I actually comment these lines out.

I wonder if we could edit [ssh/config](https://github.com/department-of-veterans-affairs/devops/raw/master/ssh/config) so that we wouldn't need this line.

My current setup doesn't require this script at all as I'm using [Core Tunnel](http://coretunnel.app) to establish the socks connection for me and I'm specifying the proxy.pac hosted on GitHub in my system proxy settings.

karlbrown-va: Hmm - if we could edit the ssh/config so we wouldn't need it that would be great; I haven't figured out the right incantations to do so however.

Core Tunnel looks like a cool solution as well. I'm a bit uncomfortable having proxy.pac hosted on github however; I know it's a bit of a stretch but I don't much like the idea of asking a file on the internet whether to proxy a given request. I know it's not that much of a difference however.

scripts/socks/README.md
ffafara-tw: `networksetup -listallnetworkservices` returns completely different order for me, too. So I don't think it's a valid option.

scripts/socks/socks.sh
ffafara-tw: proxy.pac file is used only during tunnel initialization to configure the proxy - it's not accessed on each request. It's an interesting idea to just grab it directly from github instead of running a local web server. There might be situations thought when folks might want to edit the rules in that file without pushing the changes where they would affect everyone (I've done it when needed access to more domains).

ericboehs: I'm hosting mine in a gist so if you want to diverge from repo, you can just fork/host else where.

I think proxy.pac is accessed more often than once but less often than every request. Maybe when the system boots or the interface goes down/up. I'm not certain of that though.

ffafara-tw (approved):

scripts/socks/README.md
ericboehs: `networksetup -listnetworkserviceorder` may be what we're looking for. But I'm fine with manually specifying as well.

ffafara-tw: This is what I get:
`networksetup -listnetworkserviceorder
An asterisk (*) denotes that a network service is disabled.
(1) Apple USB Ethernet Adapter
(Hardware Port: Apple USB Ethernet Adapter, Device: en7)

(2) USB 10/100/1000 LAN
(Hardware Port: USB 10/100/1000 LAN, Device: en8)

(3) Wi-Fi
(Hardware Port: Wi-Fi, Device: en0)

(4) Bluetooth PAN
(Hardware Port: Bluetooth PAN, Device: en6)

(5) Thunderbolt Bridge
(Hardware Port: Thunderbolt Bridge, Device: bridge0)`

ericboehs: Alright fine:
```
networksetup -listnetworkserviceorder | grep -B1 $(route get example.com | grep interface | awk '{print $2}') | head -1 | cut -f2- -d' '
```

:joy:

It's a little crazy. I'm good with skipping.

karlbrown-va: Let's consider this for another PR. Your approach may work and might make it more bulletproof - but probably needs more testing across different machines.

Issue #11513, opened 2020-07-22
Migrate GA Domo datasets to BigQuery


joanneesteban:
## Issue Description
How might we switch over GA Domo datasets to BigQuery?

---
## Tasks
- [x] Set up accounts
- [x] Use Domo connector
- [x] Write queries
- [x] Test queries
- [x] Write out BigQuery<>Domo schema for GA
- [x] Backup current Domo charts
- [ ] Processes are documented

- [x] Migrate charts from GA to BigQuery

## Acceptance Criteria
- [ ] Understand how to switch over GA Domo datasets to BigQuery
---
## How to configure this issue
- [ ] **Attached to a Milestone** (when will this be completed?)
- [ ] **Attached to an Epic** (what body of work is this a part of?)
- [ ] **Labeled with Team** (`product support`, `analytics-insights`, `operations`, `service-design`, `tools-be`, `tools-fe`)
- [ ] **Labeled with Practice Area** (`backend`, `frontend`, `devops`, `design`, `research`, `product`, `ia`, `qa`, `analytics`, `contact center`, `research`, `accessibility`, `content`)
- [ ] **Labeled with Type** (`bug`, `request`, `discovery`, `documentation`, etc.)

joanneesteban:
## Sprint 32 Tasks

In conjunction with [#6356](https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/6356)
- [ ] Rename the titles or work with Amy to clearly indicate maintenance of dashboards/pages
- [ ] Document testing plan/updates
- [ ] Document what ETL funnels might look like
- [ ] Draft template for ETL funnels
- [ ] Document how time to complete might look like
- [ ] Document any other change
- [ ] Test with recurring monthly reports
  - [ ] VEO
  - [ ] OMB
  - [ ] DMC
- [ ] Stretch: Continue testing with Covid dashboard (DO NOT do this on Thursday or Friday!)
- [ ] Further Stretch (Please reach out to the relevant PM & PO before testing on their KPI dashboards. Also put a maintenance flag before testing) Continue testing with:
  - [ ] Facility Locator
  - [ ] Chatbot
  - [ ] Medical Device Reordering
  - [ ] VAOS

jonwehausen:
### Update 8/12
- Maintained existing copies of VEO Dashboard using Google Analytics connector
[...]

PR #14708, opened 2020-10-12
BigQuery / DOMO documentation round 3

bsmartin-ep:
For Migrate GA Domo datasets to BigQuery #11513


jonwehausen:
Looks good here! Thanks @bsmartin-ep

PR #15591, opened 2020-10-29
Create 2020-10-29 update no-absolute-path eslint config

erikphansen:
Let's talk about disabling a low-value ESLint rule for the sake of improving how we use our Babel aliases


rianfowler (approved):

mpelzsherman:
It would be cool if there were some way to indicate that the import path maps to an alias. The leading slash, as you point out, is ambiguous and does not really achieve this. I wonder if we could use the `~` character instead, similar to how in UNIX, `~` is an alias for one's home directory? I'm not familiar with Babel aliases so this is just a naive suggestion.

platform/engineering/request-for-comment/2020-10-29 update no-absolute-path eslint config.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Design
+The proposed change is to update the `import/no-absolute-path` ESLint rule config to allow absolute imports for ES modules. This line in the `.eslintrc` should do the trick: `"import/no-absolute-path": [2, {"esmodule": false, "commonjs": true}]`
+
+Moving forward, whenever we have imports that take advantage of our `root` alias, we should make a point to prefix them with a leading slash. So `import startApp from 'platform/startup'`  becomes `import startApp from '/platform/startup'`. I am _not_ proposing making a massive change to update all existing imports with a leading slash when possible.
+
+## Risks
+With the rule turned off, it would be possible for someone to _actually_ import code from the root of their local disk. This risk would be mitigated by the fact that:
+1. That is a _seriously_ bad practice that a developer with any amount of experience would never do
```
micahchiang, line 25: :homer_disappear:

erikphansen:
@mpelzsherman Yeah `~` would be a nice stand in as well. That might be possible with some extra configuration. I will check it out!

Two points I'll make in support of using `/` to stand in for "the root of our project":
1. `/` already means "root" in filesystem-speak. We are just "taking back" the term to mean "the root of our projects" rather than "the root of our file system."
2. It's free! It just works without any extra config :) (I mean aside from disabling the ESLint rule that's prevent us from using it right now)

platform/engineering/request-for-comment/2020-10-29 update no-absolute-path eslint config.md
erikphansen: I actually tried to import something from my home folder `/Users/erikhansen/data.js` and either VSCode and/or our project setup did _not_ like that.

erikphansen:
@mpelzsherman @U-DON I'm kinda liking the `~` to mean `./src`:
```
import backendServices from '~/platform/user/profile/constants/backendServices';
import profileManifest from '~/applications/personalization/profile/manifest.json';
```

erikphansen:
I added a summary section at the top of the Markdown doc indicating we will not disable the ESLint rule and instead add a new `~` Babel alias.

PR #15629, opened 2020-10-30
Decision 4 - return uninflected VA-keys for camelCase

philipwritescode:
See PR for details, but a decision was made on how to move forward with VA inflections for vets-api.


annaswims:
Was there any consideration for renaming the file to match the class name?

docs/adr/0004-return-uninflected-va-keys.md
```diff
@@ -0,0 +1,25 @@
...
+Whom:
+- Philip Becker
+
+## Status
+
+Accepted
+
+## Context
+
+Using the Rails 6 zeitwerk file loader requires strict naming conventions of files.  So without a Rails inflection, `va_something` class files are expected to have the class name of `VaSomething`.  With an inflection, that file would be the expected `VASomething`.  Using this inflection has a downstream affect on json responses transformed into camelCase by the [olive_branch](https://github.com/vigetlabs/olive_branch) gem.
```
LindseySaari, line 14:
```suggestion
Using this inflection has a downstream affect on the named attributes in serializes and thus, json responses transformed into camelCase by the [olive_branch](https://github.com/vigetlabs/olive_branch) gem.
```

docs/adr/0004-return-uninflected-va-keys.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Decision
+
+To create a deprecation path, there will be a patch implemented for the olive_branch middleware that returns an uninflected version of keys containing `VA`.
+
+## Consequences
+
+- We will be returing additional keys in JSON responses using a `VA`-named key, so this will make payload sizes slightly larger.
```
LindseySaari, line 22:
```suggestion
- We will be returning additional keys in JSON responses using a `VA`-named key, so this will make payload sizes slightly larger, supporting both naming schemes (`VA`-named key and `Va`-named key)
```

docs/adr/0004-return-uninflected-va-keys.md
erluti: how can i not take this opportunity to include the word `thus`!?

erluti:
> Was there any consideration for renaming the file to match the class name?

As far as using the inflection goes, yes it was considered.  But vets-api has a mix of the two throughout, and it was thought to be better to use VA as an inflection since it is one and even impacted some document creation when `humanize` lower-cased `VA` when it's not an inflection.

rianfowler (approved):

erluti:
This PR is when this change will be implemented, so these should be merged in parallel https://github.com/department-of-veterans-affairs/vets-api/pull/5247
That PR needs to be changed to reflect what was decided today in the mob programming session

PR #16109, opened 2020-11-12
15581 update pact documentation

LindseySaari:
This updates the Pact documentation with the new broker url in our AWS infrastructure. Some information is included about the old url


U-DON (approved):
I think we can now remove the part about the broker being in the processs of moving, since it's officially moved. 🎉

PR #16184, opened 2020-11-13
Myva research la

lalexanderson-dsva:
Made some suggestions on both docs. 

What's not clear to me from this RP: 
- A/B testing is usually for a small feature or a specific set of content, not entirely different designs. If you think about the scientific method - you're changing ALL of the variables here, not just one or two. How are you planning to evaluate task success between the two designs? 
- Are you planning to show every user BOTH designs and run through the tasks on both, or are you showing half of your participants A and the other half B? 
- If the former, I'd recommend 90 mins - 2 hours for each of these studies. 
- If the latter, you have a really small sample size, which you're dividing further, which is going to make it hard to be confident that you got equivalent representation for both of your samples. A/B testing is usually done with large sample sizes. At minimum, I would revise your recruitment strategy to articulate that you need a minimum of 2 of every key role you're looking to recruit. You need 2 women, 2 veterans with disabilities, etc. etc. So you have one looking at A and one looking at B. And if one drops out from one side, you need to specifically recruit to fill that spot with someone of that profile. But know that also means you're making design decisions based on 1 woman's opinion, 1 disabled veterans' opinion. 

A last note - va.gov spells ```health care``` as two words, not one.

### DIFF
--- a/products/identity-personalization/my-va/2.0-redesign/product/Research/MyVA Research Plan.md
+++ b/products/identity-personalization/my-va/2.0-redesign/product/Research/MyVA Research Plan.md
@@ -4,14 +4,14 @@
 This research is to test two mobile versions of the My VA dashboard re-design and gain feedback on the preferred design direction (A vs. B). This includes:
 
 - overall, is there enough relevant information on the dashboard to be valuable
-- can the user find the information that they are looking for effectively
+- can the user quickly find the information that they are looking 
 - does the page structure (scrolling vs navigation) affect ease of visibility
-- can the user easily access and understand any personal updates (claims, healthcare, or forms)
+- can the user easily access and understand any personal updates (claims, health care, or forms)
 - how easily can the user navigate to the sections they are looking for
 - is the new dashboard layout more effective than the current MyVA page
 
 ## Background
-Our goal is to create a My VA dashboard design that will improve the online experience for Veterans, by better integrating Veteran-provided data onto the My VA dashboard. Our MVP will focus on surfacing their healthcare and claims information into widgets and notifications to help Veterans access their critical information and to take action on status updates.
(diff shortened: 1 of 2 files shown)


Samara-Strauss:
@lalexanderson-dsva 

> A/B testing is usually for a small feature or a specific set of content, not entirely different designs. If you think about the scientific method - you're changing ALL of the variables here, not just one or two. How are you planning to evaluate task success between the two designs?

I think the term "A/B testing" has been used incorrectly on the design team (though, I believe this has been cleared up with Shawna and the team). People are using it to mean "showing people two designs and seeing what they like better" when, as you mentioned, that's not what it means. I'm sorry I didn't catch the mention of "A/B testing" when I was reviewing these documents myself, or I would have had it taken out. 

**TL; DR**: We are not doing A/B testing. We are doing user interviews around two designs.

> Are you planning to show every user BOTH designs and run through the tasks on both, or are you showing half of your participants A and the other half B? If the former, I'd recommend 90 mins - 2 hours for each of these studies.

Yes, we are planning to show both designs, **but** for whichever design is second, there are far fewer questions we're asking. This is a little hard to follow on paper, but in the convo guide, you'll notice Tressa/Cassandra designated that some questions won't be asked for the prototype that is shown second. Therefore, the interview for for the second prototype should maybe be half that or less of the first.

Personally, I still think an hour is fine, but we might want to run a practice round first to confirm. If a mock session went a full 60 minutes, I would say we definitely would want to schedule folks for 90 minutes.

> If the latter, you have a really small sample size, which you're dividing further, which is going to make it hard to be confident that you got equivalent representation for both of your samples. A/B testing is usually done with large sample sizes.

Everyone will see both designs, so 8 people should be enough. And again, we're **not** doing A/B testing, so sorry about the confusion around that.

PR #16208, opened 2020-11-16
Update websites-apply-wizard.md

andaleliz:
Adding language to specify that the browser back button should behave as such when going back to the wizard from an intro page


andaleliz:
Ack, I also started to add a link to the prototype but got distracted before I actually added it.  Will add on the master branch if you are okay with these edits, @mnorthuis

mnorthuis (approved):

andaleliz:
@mnorthuis I've looked this over and I'm still not sure 😆   I can't find where this document is linked or referenced for teams to use, so I think it's okay to scrap my additions (adding mock-ups) and disregard this.

PR #16291, opened 2020-11-17
Create accessibility-for-localization.md

1Copenut:
Capturing technical notes and guidance for teams looking to localize their content. These notes and examples assume Spanish for the moment, but can absolutely be expanded to include other languages as warranted.

### DIFF
--- a/platform/accessibility/resources/accessibility-for-localization.md
+++ b/platform/accessibility/resources/accessibility-for-localization.md
@@ -1,0 +1,2 @@
+# Accessibility for Localization
+Adding a placeholder to collect notes from early discussion with DEPO Content Lead
--- a/platform/accessibility/resources/accessibility-for-localization.md
+++ b/platform/accessibility/resources/accessibility-for-localization.md
@@ -23,12 +23,19 @@ Adding a placeholder to collect notes from early discussion with DEPO Content Le
   </a>
   ```
   
-* We'll also want to make sure that `aria-labels` contain full phrases so that the text translates accurately. For example, if a button's visible text is "Enroll now," the aria label text should include "Enroll now in VA health care."
+* We'll also want to make sure that `aria-labels` contain full phrases so that the text translates accurately. For example, if a button's visible text is "Submit my application", the aria label text might be "Submit my application for education benefits".
 * Accessibility experts can provide more guidance for how to make translated content successful for screen readers.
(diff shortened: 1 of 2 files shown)


platform/accessibility/resources/accessibility-for-localization.md
```diff
@@ -1,2 +1,35 @@
...
+  ! Screen reader only text must be updated with the visual text to ensure good localization
+  
+  <a href="#">
+    Presentar mi solicitud
+  - <span class="vads-u-visibility--screen-reader"> for education benefits</span>
+  + <span class="vads-u-visibility--screen-reader"> de beneficios educativos</span>
+  </a>
+  ```
+  
+* We'll also want to make sure that `aria-labels` contain full phrases so that the text translates accurately. For example, if a button's visible text is "Enroll now," the aria label text should include "Enroll now in VA health care."
```
jenstrickland, line 26: We should not have any buttons that simply say "Enroll now" to meet [WCAG 2.0 Level A - 2.4.4](https://www.wuhcag.com/link-purpose-in-context/), and the aspirational [ WCAG 2.0 Level AAA - 2.4.9,](https://www.wuhcag.com/link-purpose-link-only/).

platform/accessibility/resources/accessibility-for-localization.md
```diff
@@ -1,2 +1,35 @@
...
+* Accessibility experts can provide more guidance for how to make translated content successful for screen readers.
+* We can also wrap individual pieces of translated content in `"lang-es"` so that screen readers will use the right pronunciation to read the translated text. We are going to try this in a test environment for the block of helper text on [va.gov/health-care/how-to-apply](va.gov/health-care/how-to-apply).
```
jenstrickland, line 28: The `"lang-es"` may not be correct. Recommend providing the `<span lang="es">Jugador</span>` example.

jenstrickland (changes requested):

platform/accessibility/resources/accessibility-for-localization.md
1Copenut: Ack, that one's an oversight. Will correct.

platform/accessibility/resources/accessibility-for-localization.md
1Copenut: Updated this text.

jenstrickland (approved):

1Copenut:
@peggygannon or @gmaciolek Would either of you have time to review this PR this week? Thank you!

PR #17056, opened 2020-12-07
VAgov SSO Brain Dump Part 2

ericbuckley:
Documentation on the VAgov SSOe project and its rollout over 2020.


ericbuckley:
@joeniquette more to come, but I wanted to open this up to you a little early so you can provide some feedback while I'm still writing.

PR #17265, opened 2020-12-10
Update and rename helpdesk_reporting_process to helpdesk_reporting_pr…

erogray:
…ocess.md

Also removed section with email addresses, so as not to show this to the public


VanessaLuxen (approved):

VanessaLuxen:
I approve this change.

PR #17282, opened 2020-12-10
Update collab-cycle-feedback.md

allison0034:
Added directions for VSP team


shiragoodman (commented):
Can we add 1 more bullet?

- Add to the Feedback pipeline on the VSP Product Support board

Also, can we please move VSP directions to the bottom of the template? It should not be what VFS teams see first (its not relevant to them)

shiragoodman (approved):

PR #18421, opened 2021-01-12
change document title and tweak introduction language

pjhill:
* Change title of document to use upgrade instead of migration since that is already used in another document
* Tweaking the introductory language a bit for a layman explanation of the contents


ddzz (approved):
LGTM

PR #18662, opened 2021-01-15
Design Doc for Web Components

### LINKED ISSUE #18150: Write design doc for Web Component implementation
## Issue Description

We have done some initial discovery around Web Components and have decided to write a design doc to serve as a record before further development happens.

---
## Tasks
- [ ] Write a design doc

## Acceptance Criteria
- [ ] Design doc is approved
---

bkjohnson:
Closes https://github.com/department-of-veterans-affairs/va.gov-team/issues/18150

The doc can be viewed [here](https://github.com/department-of-veterans-affairs/va.gov-team/blob/wc-design-doc/platform/engineering/design-docs/2021-01-13-web-component-library.md).


platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+**Last Updated:** January 13, 2021  
+**Status:** **Draft** | In Review | Approved  
+**Approvers:** _Chris Valarida_ \[ \], _Michael Fleet_ \[ \]
+
+
+## Overview
+
+### Objective
+
+The goal is to establish a component library based on Web Components, which are a browser standard. This will supercede the current `formation-react` library and some of the liquid HTML templates currently in `vets-website`. A non-goal is creating an entirely new visual experience as part of the component library.
```
cvalarida, line 13: 1. Fun fact (which I just learned): `supercede` is apparently a "disputed variant spelling" of supersede.
2. Additionally, we can probably replace `formation-react` with `component-library` now.
3. Is the goal to completely replace the existing component library with web components?

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+
+The intended audience for this document is frontend engineers.
+
+### Background
+
+We currently have some "components" implemented in HTML/CSS (and sometimes JS) in the `formation` npm package. These can be used anywhere where `formation` is loaded. We also have `formation-react`, which is a package with some of those patterns implemented as React components. These can only be used in React apps. Lastly, in `vets-website` we have some Liquid templates which implement some of the components/patterns of the design system, and these are only used for content builds.
```
cvalarida, line 19: For clarity, prefer bullet points. (Source: [Google Technical Writing One](https://developers.google.com/tech-writing/one/lists-and-tables) course.)

Maybe something like:
```markdown
The problem this design document intends to solve is the re-implementation of common components across different repositories and libraries, using different technologies.

Currently, common components may be found in one or more of three places:
- The `formation` NPM package
  - More information here
- The `component-library` NPM package
  - More information here
- Liquid templates in `vets-website`
  - More information here
```

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+
+Maintaining multiple sources of truth is a burden for all teams involved and leads to wasted time.
+
+### High Level Design
+
+The goal is to establish a single codebase and documentation for the implementation of design system components. This will allow the components to be used in React apps, static HTML, or any other web framework.
+
+The Web Component library will live in its own repo and will be added to `vets-website` as a dependency, where the components will be loaded and defined for the browser to use. As more components are added to the Web Component version of the library, the older React-specific components will be phased out.
```
cvalarida, line 27: > As more components are added to the Web Component version of the library, the older React-specific components will be phased out.

Tell me more about this in the detailed design. How? What's the migration process look like? When will the component be removed from the library?

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+
+
+## Specifics
+
+### Detailed Design
+
+In the `component-library` repo, we will start a fresh orphan branch for the Web Component version of the library to live. If we reach a point where the React version of the library is no longer in use, we will rebase the orphan Web Component branch off of master (in order to cleanly preserve the history) and that will become the new default branch going forward.
```
cvalarida, line 34: Have you tried doing this to make sure it'll work like you want? I'm not super familiar with rebasing.

When you say "orphan branch," the web component branch will share a history with `master`, right?

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+
+Until we reach that point, the WC branch will be developed in the `component-library` repo alongside the main branch containing the React components. The WC branch will not be published to npm, and the `vets-website` dependency will instead use a Git url with a tag.
```
cvalarida, line 36: Tell me about this tag. When will it be created?

Link to the NPM documentation that says this is possible.

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+### High Level Design
+
+The goal is to establish a single codebase and documentation for the implementation of design system components. This will allow the components to be used in React apps, static HTML, or any other web framework.
+
+The Web Component library will live in its own repo and will be added to `vets-website` as a dependency, where the components will be loaded and defined for the browser to use. As more components are added to the Web Component version of the library, the older React-specific components will be phased out.
+
+
+## Specifics
+
+### Detailed Design
```
cvalarida, line 32: I recommend breaking this up into subsections. Maybe the following:
- Migration to web components
- Tooling
- Loading web components on VA.gov
- Styling

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+
+One of the additional benefits of Web Components is how they can be styled. Each Component can have its own styles scoped to its [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) (this means that if a Component renders a `<div>`, the CSS can broadly style the `div` node without worrying about it affecting all the other `<div>`s on the page). As a result, each component will have it's own CSS file which also makes use of global [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties).
+
+Once in `vets-website`, the setup code for the WC library will be in its own entry file that is loaded on each page. In addition, this script will be one of the first loaded in order to ensure that the Components will be defined in the browser before they are rendered on the page.
```
cvalarida, line 42: - What does the "setup code for the WC library" do?
- How will the script be one of the first loaded?
- Tell me about lazy loading

Don't forget to link to the appropriate Stencil docs so you don't have to write it all yourself. :wink:

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Privacy Concerns
+
+N/A
+
+### Open Questions and Risks
+
+Accessibility and performance are two things that should receive lots of attention.
+
+Do we want to remove the explicit dependency on [USWDS](https://designsystem.digital.gov/)? Rather than explicitly relying on `uswds`, the Component library could be an abstraction for `vets-website` that implements the appropriate styles from `uswds`. A Component library complete with global styling defaults might be able to accomplish everything that `uswds` is providing.
```
cvalarida, line 83: Two things come to mind:
1. Because we can independently style web components, is it feasible to adopt USWDS 2.0 styles in just our web components?
1. I believe the design system's component library should be an implementation of the visual design language that the design system defines. It should be opinionated about its styles and provide overrides only as needed. (To maintain consistency across all of VA.gov, those needs should be deliberate—more than just, "I don't like it.")

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+
+#### Risks
+
+Despite polyfills, there is some risk of certain Web Component features not working properly in all browsers (IE 11, older versions of Edge). Due to existing polyfill support combined with the trend of users switching to more modern browsers over time, this is a low risk. If we encounter this problem, the workaround will be to avoid using whatever feature is under consideration and instead implement a solution that will yield the same results.
```
cvalarida, line 87:
```suggestion
Despite polyfills, there is some risk of certain Web Component features not working properly in all browsers (IE 11, older versions of Edge). Due to existing polyfill support combined with the trend of users switching to more modern browsers over time, this is a low risk. To mitigate this risk, we will only use Web Component features which are supported with or without pollyfills on all [supported browsers](https://department-of-veterans-affairs.github.io/veteran-facing-services-tools/platform/front-end-standards/documented-decisions/browser-support/).
```

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+
+#### Develop an Accordion Web Component
+
+**Estimate**: ~4 days
+
+This will be the first WC to go live on va.gov.
+
+### Alternatives
+
+There isn't another way of achieving a proper (i.e., not copy-pasted HTML) component library that isn't specific to a single framework like React. However, there are other tools/frameworks besudes Stencil that we could use to build Web Components.
```
cvalarida, line 113:
```suggestion
There isn't another way of achieving a proper (i.e., not copy-pasted HTML) component library that isn't specific to a single framework like React. However, there are other tools/frameworks besides Stencil that we could use to build Web Components.
```

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+
+There are [dozens of tools/frameworks](https://webcomponents.dev/blog/all-the-ways-to-make-a-web-component/) that can be used to build Web Components, with [LitElement](https://lit-element.polymer-project.org/) being one of the bigger ones. Here are some reasons that Stencil was chosen over all of these alternatives:
+
+- JSX support
+  - Stencil is one of the few tools for building Web Components that lets the developer use JSX  syntax, as opposed to a template string or direct DOM manipulation in order to create the HTML  nodes. This will be very familiar for React developers.
+- Typescript
+  - The Component source will be written in TypeScript, and static typing can help eliminate bugs in code as well as allow developers to gain a clearer understanding.
```
cvalarida, line 120: This should be called out in the detailed design.

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+### Future Work
+
+Continue adding Components until the React version of `component-library` ic completely replaced. Web Components can co-exist fine with React components, but it will be easier if we only have to maintain a single library.
+
+#### Templating
+
+The library could be extended to include Components focused on overall site layout in addition to the more granular Components. This could allow us to use Web Components as a replacement for the Liquid templates in `vets-website`.
+
+### Revision History
+_The table below should record the major changes to this document. You don't need to add an entry for typo fixes, other small changes or changes before finishing the initial draft._
```
cvalarida, line 133:
```suggestion
```

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+- JSX support
+  - Stencil is one of the few tools for building Web Components that lets the developer use JSX  syntax, as opposed to a template string or direct DOM manipulation in order to create the HTML  nodes. This will be very familiar for React developers.
+- Typescript
+  - The Component source will be written in TypeScript, and static typing can help eliminate bugs in code as well as allow developers to gain a clearer understanding.
+- Auto-generated component docs
+  - Stencil is able to automatically generate a README for each component with documentation on its props, events, CSS variables, and more. This can be combined with Storybook.
+
+### Future Work
+
+Continue adding Components until the React version of `component-library` ic completely replaced. Web Components can co-exist fine with React components, but it will be easier if we only have to maintain a single library.
```
cvalarida, line 126:
```suggestion
Continue adding Components until the React version of `component-library` is completely replaced. Web Components can co-exist fine with React components, but it will be easier if we only have to maintain a single library.
```

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,139 @@
...
+
+#### Templating
+
+The library could be extended to include Components focused on overall site layout in addition to the more granular Components. This could allow us to use Web Components as a replacement for the Liquid templates in `vets-website`.
```
cvalarida, line 130: Maybe link to https://stenciljs.com/docs/static-site-generation?

I'd also recommend saying only that this is an _option_, not a thing we will definitely do as future work.

cvalarida (commented):
This is a good first draft, but I'd like to see more meat in the detailed design section.

Design documents are intended to be very technical. Feel free to get into the weeds with the implementation details! Tell me how we load the WC loader. Tell me how that loader works. Tell me about lazy loading the components. Tell me about the performance implications and how Stencil mitigates that risk. Tell me why I might be nervous about adding yet another technology to the stack. Tell me how our implementation addresses my concerns. Tell me about the relationship between React and web components. Can I treat them like any other blackbox React component?

I know we've talked about these things in Slack, so I know you've thought about them, but they should be captured here as well. Be sure to heavily reference the Stencil documentation.

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: 3. I would like that to be the goal.

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: I left out some detail here. I have done all of this locally except for the rebasing, but I'm pretty confident it will work like I expect. I created a branch off of `master` called `wiped` where I deleted everything but the `.git` directory, and then I created the orphan branch off of that. That gave me a branch with a clean slate in the same repo, but none of the history.

---

I just did the rebase as a sanity check, and the only conflicts were `.gitignore`, `package.json`, and the `.storybook` `main.js` and `preview.js` files, which is what I expected.

The result is [this branch](https://github.com/department-of-veterans-affairs/component-library/tree/wc-pilot-rebase-test). The commits are clean, but I have some more things to delete which isn't a big deal.

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: 1. The short answer is "probably". The longer answer would involve looking at the differences between USWDS 1.0 and 2.0 and seeing where they differ. However, we can make the CSS for each component be whatever we want. This could mean copy-pasting USWDS stuff (not ideal), or adding it as a dependency and `@import`ing the appropriate CSS files into each component and then making any additional modifications that we want.

With that being said, I think that adopting USWDS 2.0 as part of this design document would fall under non-goals:

> A non-goal is creating an entirely new visual experience as part of the component library.

2. I agree.

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,185 @@
...
+
+#### Styling
+
+One of the additional benefits of Web Components is how they can be styled. Each Component can have its own styles scoped to its [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) (this means that if a Component renders a `<div>`, the Component's CSS can broadly style the `div` node without worrying about it affecting all the other `<div>`s on the page). As a result, each component will have it's own CSS file which also makes use of global [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties). These variables will be modeled after `formation`/`uswds`.
+
+#### Loading Web Components on va.gov
+
+Once in `vets-website`, the setup code for the WC library will be in its own entry file that is loaded on each page. This code will be responsible for applying polyfills for older browsers as well as defining the Web Components.
```
cvalarida, line 69: I think this is what you mean?
```suggestion
The setup code for the WC library will be in its own `vets-website` entry file that is loaded on each page. This code will be responsible for [applying polyfills](https://stenciljs.com/docs/faq#what-polyfills-does-stencil-provide) for older browsers as well as [defining the Web Components](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define).
```
There may be better links for both of those. :man_shrugging:

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,185 @@
...
+
+In addition, this script will be one of the first loaded in order to ensure that the Components will be defined in the browser before they are rendered on the page. This will be done by adding a `<script>` entry towards the top of [the `<head>` tag that gets loaded on every page](https://github.com/department-of-veterans-affairs/vets-website/blob/dc5018065eccf073d172097a19922755126c98a3/src/site/includes/header.html#L16)
+
+For further optimization, the Components will be lazy-loaded. If our Web Component library ends up having hundreds of Components, this could cause a significant performance impact if we defined these every time a new user landed on a va.gov page. Instead, the Components will only _actually_ be defined if they appear on the page. [Stencil has a useful article that talks about their lazy loading approach](https://stenciljs.com/blog/how-lazy-loading-web-components-work).
```
cvalarida, line 73:
```suggestion
For further optimization, the Components will be lazy-loaded. Even if our Web Component library ends up having hundreds of Components, lazy loading will reduce the risk of a significant negative performance impact. The Components will only _actually_ be defined if they appear on the page. [Stencil has a useful article that talks about their lazy loading approach](https://stenciljs.com/blog/how-lazy-loading-web-components-work).
```

> The Components will only _actually_ be defined if they appear on the page.

Is this true? My understanding from that article is that it registers _wrappers_ for **all** the components, and then when they're rendered by the browser, the component wrapper renders the "skeleton" (not a real term) and makes the async call to fetch the actual component logic.

We might reference Ionic as an example of a really popular Web Component library using Stencil which has 100+ components. Bonus points for performance metrics.

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Code Location
+
+The code will live in the `component-library` repo on a dedicated orphan branch named `web-components`. This code will be used in `vets-website` via a dependency in `package.json`
+
+### Testing Plan
+
+Each Component will have:
+
+- Unit and E2E tests written using Jest
```
cvalarida, line 83: And whatever Stencil gives us? https://stenciljs.com/docs/testing-overview

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,185 @@
...
+- One or more stories in Storybook to validate appearance/behavior
+
+As of mid-January 2021, Stencil provides an experimental screenshot visual diff testing functionality. If this becomes stable, we will consider adopting it.
+
+### Logging
+
+N/A
+
+### Debugging
```
cvalarida, line 92: How will Sentry fit into this? (Maybe the answer is that it won't, but it's worth calling out explicitly.)

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,185 @@
...
+
+Storybook will be a useful tool for debugging and development. Stencil also comes with its own development server which can be used.
+
+There will be no special provisions for debugging provided for consumers of the WC library.
+
+### Caveats
+
+Even though Web Components are already a feature provided by the browser, this is a feature that the VA Design System hasn't yet taken advantage of. By introducing this "new" technology, there is a possibility of this just turning into technical debt if it exists alongside the current `component-library`, `formation` component styles, and Liquid templates for components. In order to avoid this, we want to be agressive about this migration and get out of the transitionary period as quickly and feasibly as we can.
```
cvalarida, line 100: From the template, this section is for
> Gotchas, differences between the design and implementation, other potential stumbling blocks for users or maintainers, and their implications and workarounds. Unless something is known to be tricky ahead of time, this section will probably start out empty.

What you've got here seems more like a risk to me. What do you think?

platform/engineering/design-docs/2021-01-13-web-component-library.md
cvalarida: My opinion: Let's say we're going to use USWDS 2.0 for our WC styling and if we run into any problems with that, we can reassess later. The design document is meant to be updated if we find the implementation needs to change.

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,185 @@
...
+
+**Estimate:** ~4 days
+
+This has already been done as a PoC - some additional work should be done to make sure that the built version of storybook can use the Web Components
+
+#### Create new Webpack entry for Web Component loader
+
+**Estimate:** ~3 days
+
+`vets-website` should create a `wc-loader.entry.js` file that will get loaded in the `<head>` of each page.
```
cvalarida, line 138: Bonus points for mentioning how cache busting will work.

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,185 @@
...
+- Typescript
+  - The Component source will be written in TypeScript, and static typing can help eliminate bugs in code as well as allow developers to gain a clearer understanding.
+- Auto-generated component docs
+  - Stencil is able to automatically generate a README for each component with documentation on its props, events, CSS variables, and more. This can be combined with Storybook.
+
+### Future Work
+
+Continue adding Components until the React version of `component-library` is completely replaced. Web Components can co-exist fine with React components, but it will be easier if we only have to maintain a single library.
+
+We could also significantly improve performance by using Web Components to replace React on pages that aren't React apps. Right now in `vets-website`, the "MegaMenu" is a React component and it is loaded in the header on each page. Having all of the React code necessary for that in the `static-pages` bundle bloats the bundle size and reduces performance. If this were a Web Component instead, less JS would have to be shipped in the bundle since the browser already knows what to do with Web Components and it doesn't need to load additional framework code.
```
cvalarida, line 163: I'm not recommending or asking for a change here right now since this is in `Future Work`, but food for thought: Are you saying the `MegaMenu` should be a component in the component library, then? If not, will there be two separate WC loaders—one from `component-library` and one from `vets-website`? Or is there a way we can use the one loader and register components from both places?

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,185 @@
...
+Currently, the design system has various implementations of "components" scattered in various places to serve various tools/libraries:
+
+- The `formation` npm package
+    - This contains static HTML/CSS (and sometimes JS), and is used by copy-pasting markup onto static pages
+-  The `component-library` package
+    - This is the rebranded `formation-react`, containing React implementations of some patterns defined in `formation`. Can only be used inside React apps
+- [Liquid templates in `vets-website`](https://github.com/department-of-veterans-affairs/vets-website/tree/958d89f18db29a1f3e1f1cbe4e304299eb387de7/src/site)
+    - This is similar to the static markup found in `formation`. These templates of components are only used in the content build for va.gov
+
+Maintaining multiple sources of truth is a burden for all teams involved and leads to wasted time.
```
cvalarida, line 28: Tell me how using Web Components is going to solve this problem.

cvalarida (commented):
I left a bunch of comments, but I hope it's less work than it looks! :sweat_smile: They're categorized roughly like so:
- High priority
  - The migration plan—this is the only one that's really stopping me from give this my stamp of approval right now
- Medium priority
  - Minor questions like in the `Testing Plan` and `Debugging` sections
- Low priority
  - Edits for grammar and clarity

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,185 @@
...
+
+Until we reach that point, the WC branch will be developed in the `component-library` repo alongside the main branch containing the React components. The WC branch will not be published to npm, and the `vets-website` dependency will instead [use a Git url with a tag](https://classic.yarnpkg.com/en/docs/cli/add/).
+
+> `yarn add <git remote url>#<branch/commit/tag>` installs a package from a remote git repository at specific git branch, git commit or git tag.
+
+We will use a [Github Release](https://github.com/department-of-veterans-affairs/component-library/releases/new) to create a new tag for the branch we want. The tags will follow [semantic versioning](https://semver.org/), and will be created after enough new development has landed in the branch (i.e. a new component, bugfix, etc. Documentation typo fixes won't get a release). A tag will have the pattern `wc-vX.Y.Z`, where `X`, `Y`, and `Z` represent major, minor, and patch versions respectively.
```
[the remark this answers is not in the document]
bkjohnson: I'll update `a` to `A`, but I think the period should remain after the parenthesis.

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: Or right! I meant to leave a comment on this since I had a question about this exact thing :grin: . I put it in Caveats because:

> other potential stumbling blocks for users or maintainers, and their implications

I'll go ahead and move it to Risks though.

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: From the design doc template:

> The background section should contain information the reader needs to know to understand the problem being solved. This can be a combination of text and links to other documents.
>
> Do NOT describe the solution here. That goes in High Level Design.

Web Components are part of the solution, so should I really mention that here?

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: I'm not saying we _should_ turn `MegaMenu` into a Web Component. There's a lot going on there and we'd probably have to break things down a ~~bit~~ lot. I'm just pointing out that on static pages, `MegaMenu` is the only reason why React is used at all. If we eliminated that, we could improve performance overall.

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: Yeah, I glossed over the wrapper part. I'll be a bit more specific.

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,185 @@
...
+
+If a new Web Component is created that replaces an existing React component, the Web Component dependency in `vets-website` will be updated to the new tag and VFS teams will be able to use this component in their code according the the associated documentation for how to use the component. After a sprint, the React component will be removed from `component-library` and a new npm version will be published _with a major version increment_. This means that any teams that haven't updated their code will cause the build to break on the newest `component-library` version. **It will be the VFS teams' responsibility to update their app code to use the appropriate components**. It will be the job of the Design System engineers to provide Web Components which achieve parity with their React counterparts.
```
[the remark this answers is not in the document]
bkjohnson: I've made some updates to the migration plan which follow a less aggressive approach! This is based off of some learnings from the following articles:

- https://lethain.com/migrations/
- https://increment.com/cloud/case-studies-in-cloud-migration/ (specifically the section on Pinterest)

platform/engineering/design-docs/2021-01-13-web-component-library.md
cvalarida: I think the log in / search component up at the top is also built using React and is not part of the `MegaMenu`. Mmm...I _think_ it's not a part of it, anyhow. Turns out I'm too lazy to look right now, though. :see_no_evil:

image

platform/engineering/design-docs/2021-01-13-web-component-library.md
cvalarida: No, you're right. I was looking for the first sentence of the `High Level Design` you've already got. (Literally the sentence after this one. :facepalm:) I completely skipped over it.

>Maintaining multiple sources of truth is a burden...

>The goal is to establish a single codebase and documentation for the implementation of design system components.

You've got it covered. :+1: 

Bonus points for using the same terminology, though. Maybe something like

> Maintaining multiple codebases and sources of documentation for effectively the same components is a burden...

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,221 @@
...
+
+#### Styling
+
+One of the additional benefits of Web Components is how they can be styled. Each Component can have its own styles scoped to its [shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) (this means that if a Component renders a `<div>`, the Component's CSS can broadly style the `div` node without worrying about it affecting all the other `<div>`s on the page). As a result, each component will have its own CSS file which also makes use of global [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties). These variables will be modeled after `formation`/`uswds`.
+
+For now each Component will have its own styles _implicitly_ based off of USWDS, which resemble what is currently available in the Design System. Potential future work would involve explicitly importing USWDS component stylesheets into each Component's CSS file.
```
cvalarida, line 68: :+1: 

Bonus points for citing the experiment you did to use USWDS 2.0 in here and the trouble you ran into. This experiment informed your decision, so it's worth calling out that you _did_ try. In case somebody asks later. :wink:

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,221 @@
...
+1. wrapper Component lazy-loads the _actual_ definition
+1. properly defined Component is rendered in the browser
+
+[Stencil has a useful article that talks about their lazy loading approach](https://stenciljs.com/blog/how-lazy-loading-web-components-work).
+
+> It doesn’t matter how the element came to be rendered in the DOM, either way, Stencil knows how to lazy-load itself using web-standards and native browser APIs. And by deferring most of the logic to the browser, rather than generating more JavaScript runtime, the components themselves have an optimal file size.
+
+#### Migration to Web Components
+
+Once a new release is made for the Web Component library, we will [configure a linting rule](https://github.com/AlexMost/eslint-plugin-deprecate/blob/master/docs/rules/import.md) as part of the `lint:js:changed` script in `vets-website` to throw an error if a React component is imported that has a released Web Component replacement. This will prevent the older React component from appearing in more places as we are actively making efforts to remove it.
```
cvalarida, line 89: :thinking: _Can_ we add this linting rule only for the `lint:js:changed` NPM script?

With this process, if an engineer makes _any_ changes to a file that imports the legacy component, they'll be forced to update the component usage before committing (or ignore the commit hook linting), right?

For the most part, forcing an update is probably fine if we're not changing the API of the component. If it's just like a `<TextInput>` -> `<va-text-input>` rename, great, but if it'll require more involved changes, that's probably not awesome.

How might we reduce the burden for engineers making unrelated changes? Is there any way to catch only _new_ imports for legacy components?

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,221 @@
...
+
+Next, the Design System team will reach out to a VFS team that we believe will get the gratest value by adopting the Component. We will briefly describe the new Component and how it can be used in place of the older components currently used in their codebase, and we will offer to set up a meeting to synchronously go over the Component and begin the migration. The focus will be on making the migration as smooth as possible for the team as well as bringing awareness to the new Component.
```
cvalarida, line 91:
```suggestion
Next, the Design System team will reach out to a VFS team that we believe will get the greatest value by adopting the Component. We will briefly describe the new Component and how it can be used in place of the older components currently used in their codebase, and we will offer to set up a meeting to synchronously go over the Component and begin the migration. The focus will be on making the migration as smooth as possible for the team as well as bringing awareness to the new Component.
```

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,221 @@
...
+
+When we have finished helping the team transition all of their code to use the new Component, we will announce this publicly in Slack, and move on to help a new team with the migration, repeating the steps. After the new Web Component has replaced all instances of its React/`formation`/liquid conterparts, the React component will be removed from `component-library` and a new version will be published. The associated styles will also be removed from `formation` if we are certain that this will not affect other styles on va.gov.
```
cvalarida, line 93: This approach will take a _lot_ of our very limited bandwidth. What I'd recommend instead is working with the first team directly _only_ if we believe the migration might be problematic. After that, we make the announcement and offer our normal support for any teams who need help making the migration.

The trouble with that approach, of course, is that we're not guaranteeing the sunsetting of legacy React components by directly being involved in their replacement.

The tradeoff here is our bandwidth vs having the legacy component stay in the codebase. For now, I'd say our time is more important to protect. We have other things that we need to be working on. I think the answer is that we won't provide support of the legacy components, but we will provide support for migrating to the new component if it's required. Down the road, we can re-evaluate based on the priority of this work compared to others, the time required, and whatever bandwidth we have at the time. (I'm still hoping for more team members. :sweat_smile: )

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,221 @@
...
+
+Additionally, the Storybook stories for the React component will be replaced by a simple document that points the user to a different Storybook host that is exclusively for Web Components. See [Chromatic hosting](#chromatic-hosting) for possible future work to improve on this.
+
+This process will repeat for each Component that has React/`formation`/liquid counterparts to replace.
+
+### Code Location
+
+The code will live in the `component-library` repo on a dedicated orphan branch named `web-components`. This code will be used in `vets-website` via a dependency in `package.json`
+
+### Testing Plan
```
cvalarida, line 103: How will this be tested in `vets-website`? Do we need to create any guidance? Tools?

cvalarida (commented):
I still have concerns about the migration, but offered a path forward. 30+ components * 20+ teams (no idea how accurate that is) is a _looooooooot_ of time spent helping folks migrate.

I understand not every team uses every component, and not every component is going to need a lot of help to migrate to, but committing to this is pretty aggressive.

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: > _Can_ we add this linting rule only for the lint:js:changed NPM script?

Yes, but that will likely involve some minor refactoring of the current config and/or the script itself. One option would be to change [the current command](https://github.com/department-of-veterans-affairs/vets-website/blob/9da6a50293000210257703ffef135c10fb6fe6d9/package.json#L38) from `eslint --quiet` to something like:

```
eslint --quiet -c component-library.eslint.js
```

Where `component-library.eslint.js` is [a custom configuration file](https://eslint.org/docs/user-guide/configuring#configuration-file-formats) that imports the existing `.eslintrc` and exports it as json with our new rule added along with the export.

---

> With this process, if an engineer makes any changes to a file that imports the legacy component, they'll be forced to update the component usage before committing (or ignore the commit hook linting), right?

I believe so - that is the limitation of our current pre-commit linting. It lints entire changed files, not just the changed chunks.

---

> How might we reduce the burden for engineers making unrelated changes? Is there any way to catch only new imports for legacy components?

I mean... yeah, it's a solvable problem. We could capture the output from the `eslint` command, and if there is an error we could:

1. See if it is related to a `component-library` import
1. Check the git log or blame and see if the commit is before or after `HEAD`

Do you think this is something we need to put the time into?

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: Some tests may need to be updated, i.e. the ones that are looking at the DOM for things specific to the React components. Some `skin-deep` tests use shallow rendering and may have a test that expects a `CollapsiblePanel` node to be rendered, for example. Other components may have specific classes that a test might look for, but with the Web Component implementation these classes may no longer be there or be within the shadow DOM.

I don't think we need to provide any tools specific to `vets-website`. If we find that there are common problems with a non-obvious solution related to testing, then anytime we make an announcement about a new WC I think we should include a blurb about that, like:

> Q: I'm getting [SPECIFC TEST ERROR], how do I fix it?
> 
> A: This has happened with other teams/components. We recommend XYZ.

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: I'll link to the section in Future Work where I talk about this.

platform/engineering/design-docs/2021-01-13-web-component-library.md
cvalarida: Cool. Let's work with the FE tools team on this when the time comes. We can iterate on the linting even after this doc gets approved.

cc @Clark-W-Griswold @meganhkelley for visibility

platform/engineering/design-docs/2021-01-13-web-component-library.md
cvalarida: I guess what I'm asking about is how web components will interact with the existing e2e and unit tests. Like, how will they get loaded?

So I know we'll need to change those `skin-deep` tests, but when we interact with the web component in the unit test—like clicking on the `<va-accordion>` to ensure that the expected content is there—how will the functionality be there? We're using JSDOM and—presumably—aren't adding the WC loader to it to register the components and provide functionality. Unless I'm wrong? (I know basically nothing about testing web components.)

The tool I was asking about is more along the lines of providing the web component functionality in downstream (`vets-website`) unit tests which have to interact with the components.

If I'm still not making sense (because words are hard :sweat_smile:), please reach out in Slack and we can talk there. :+1:

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,221 @@
+# Web Component library
+
+**Author(s):** Brooks Johnson  
+**Last Updated:** January 13, 2021  
+**Status:** **Draft** | In Review | Approved  
+**Approvers:** _Chris Valarida_ \[ \], _Michael Fleet_ \[ \]
```
cvalarida, line 6:
```suggestion
**Approvers:** _Chris Valarida_ [x], _Michael Fleet_ [ ]
```

cvalarida (approved):
Looks great! There are some solvable problems that we still have to work through:
- Linting rules (or something else entirely :shrug: ) to prevent new imports of deprecated components
- How to load the web components in `vets-website` unit tests
  - This is maybe already solved, but if so, we'll need to have the solution in writing here for posterity

All that said, I'm giving this the :+1: because I believe they're solvable. We'll just need to make sure to keep this document up-to-date if we make any changes down the road.

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,221 @@
...
+
+N/A
+
+### Open Questions and Risks
+
+How will we ensure that each Component meets accessibility standards?
+
+How will we measure performance of Web Components, lazy-loading, etc?
+
+#### Risks
```
cvalarida, line 140: Another good point @drorva mentioned to me is industry adoption. My understanding is it's been pretty slow going heretofore, so if the standard ends up becoming obsolete, that may affect the future of this work. I don't know what the risk of that is, but it's worth calling out.

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,221 @@
...
+**Last Updated:** January 13, 2021  
+**Status:** **Draft** | In Review | Approved  
+**Approvers:** _Chris Valarida_ [x], _Dror Matalon_ [ ], _Michael Fleet_ [ ]
+
+
+## Overview
+
+### Objective
+
+The goal is to establish a component library using the [Web Components standard](https://developer.mozilla.org/en-US/docs/Web/Web_Components). This will eventually supercede the current `component-library` code and some of the liquid HTML templates currently in `vets-website`. A non-goal is creating an entirely new visual experience as part of the component library.
```
f1337, line 13: Why do we want a Web Components library? Please add one sentence in this paragraph, probably right after the first, summarizing the value of web components vs our existing approach.

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,221 @@
...
+    - Can only be used inside React apps
+- [Liquid templates in `vets-website`](https://github.com/department-of-veterans-affairs/vets-website/tree/958d89f18db29a1f3e1f1cbe4e304299eb387de7/src/site)
+    - This is similar to the static markup found in `formation`
+    - These templates of components are only used in the content build for va.gov
+
+Maintaining multiple codebases and sources of documentation for design system components is a burden for all teams involved and leads to an inconsistent user experience and wasted engineering time.
+
+### High Level Design
+
+The goal is to establish a single codebase and documentation for the implementation of design system components. This will allow the components to be used in React apps, static HTML, or any other web framework.
```
f1337, line 35: This sentence is probably what I'm asking for in the Objective paragraph. Consider moving it up?

f1337 (commented):
I made two inline comments re: clarifying the Why? in the opening Objective section.

I'd also like to see some example code, maybe at the end of the doc, addressing the following:
* uber simple Web Component
* using WC in React
* using WC in plain ole HTML

bkjohnson:
I've added the "why" to the Objective section, and added some example code as well. I was going to include images, but I couldn't get them to resize properly, and including the rendered page along with the inspector made for a large and awkwardly placed screenshot in the document.

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: Included that point in this commit: https://github.com/department-of-veterans-affairs/va.gov-team/pull/18662/commits/86772e69b3b2d04821684374714f1df6c16135d6

bkjohnson:
@drorva and @f1337, do you have any thoughts on the changes?

f1337 (approved):
Excellent! :metal:
Bill & Ted playing air guitar, saying 'Excellent!'

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,319 @@
...
+- The `formation` npm package
+    - Built using `uswds` 1.6
+    - This contains static HTML, CSS, and vanilla JavaScript
+    - Used by copy-pasting markup onto static pages
+-  The `component-library` package
+    - This is the rebranded `formation-react`, containing React implementations of some patterns defined in `formation`
+    - Can only be used inside React apps
+- [Liquid templates in `vets-website`](https://github.com/department-of-veterans-affairs/vets-website/tree/958d89f18db29a1f3e1f1cbe4e304299eb387de7/src/site)
+    - This is similar to the static markup found in `formation`
+    - These templates of components are only used in the content build for va.gov
```
bkjohnson, line 29: @drorva , does this answer your question:

> I would add a section in the beginning of the doc that describes the current state. As I understand it, we forked uswds 1.0, use react components (based on some library?), etc. This provides context.

I recently added the bullet point:

`     - Built using `uswds` 1.6 `

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,319 @@
...
+
+Next, if we believe a particular component migration will be problematic the Design System team will reach out to a VFS team to work with them on the migration. Once the first team has completed their migration to the new Component, we will announce the success and encourage other teams to do the same. At this point we will provide support for the migration if requested, but we will no longer provide support for the legacy React component.
+
+After the new Web Component has replaced all instances of its React/`formation`/liquid conterparts, the React component will be removed from `component-library` and a new version will be published. The associated styles will also be removed from `formation` if we are certain that this will not affect other styles on va.gov.
```
bkjohnson, line 93: Does this answer the question about the frontend?

> Similar to the previous point, what does the full front-end stack look like with Web components on va.gov? Do developers still need to use Formation, react, etc. It's fine to point to info on https://design.va.gov/.

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,319 @@
...
+
+For now each Component will have its own styles _implicitly_ based off of USWDS, which resemble what is currently available in the Design System. Potential future work would involve explicitly importing USWDS component stylesheets into each Component's CSS file. See [USWDS styles](#uswds-styles) for ideas on explicitly using USWDS 2.0 styles, and some obstacles that were discovered during an experiment.
+
+#### Loading Web Components on va.gov
+
+The setup code for the WC library will be in its own `vets-website` entry file that is loaded on each page. This code will be responsible for [applying polyfills](https://stenciljs.com/docs/faq#what-polyfills-does-stencil-provide) for older browsers as well as [defining the Web Components](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define). We will be able to achieve [full browser support with polyfills](https://stenciljs.com/docs/browser-support#browser-support).
```
bkjohnson, line 72: Does this answer your question? I just added the last sentence beginning with "We will be able to achieve ..."

> https://developer.mozilla.org/en-US/docs/Web/Web_Components#browser_compatibility indicates limits in Edge and Safari beyond the ones that we're aware of in IE. Are we comfortable that browser shims solve for the feature set we need?

platform/engineering/design-docs/2021-01-13-web-component-library.md
```diff
@@ -0,0 +1,337 @@
...
+
+Web Components are a feature that is already built into modern browsers, but they don't have as wide of an industry adoption as a framework like React. Despite this, it is unlikely that the Web Component standard will be completely abandoned. There are many large companies which are already developing products using Web Components:
+
+- Youtube
+- GitHub
+- Xfinity (Comcast)
+- [Apple Music](https://hackernoon.com/apple-cements-the-unlikely-rise-of-web-components-6b1d3g1t)
+- [Adobe](https://opensource.adobe.com/spectrum-web-components/)
+
+There aren't any published articles detailing Web Component adoption by the companies without links, but Web Components can be found on their websites.
```
bkjohnson, line 164: This is a new section added in response to your comment:

> While I, like many other open source and gov people, like the idea of platform Web Components, I am not clear that I see industry adoption of this approach. I worry that down the line we'll need to pivot again because this will be a dead-end. What would reassure me is finding out that some major organization are using it.

bkjohnson (commented):
@drorva , I've made some changes in response to your questions/comments in Slack.

platform/engineering/design-docs/2021-01-13-web-component-library.md
bkjohnson: I added a bit more detail to this: https://github.com/department-of-veterans-affairs/va.gov-team/pull/18662/commits/ce991f7eb3a463e1456c6e4bbaee8e06d0aef957

f1337 (approved):
AFAIC, we're good to move forward. I'm very excited to see where the experiment leads!

More Cowbell

PR #19211, opened 2021-01-28
updated new-team-vsp-orientation-epic template

shiragoodman:
These changes were made in support of t[his issue ticket.](https://app.zenhub.com/workspaces/vsp---product-support-5f85b91c14d8df0018fac414/issues/department-of-veterans-affairs/va.gov-team/19080)


alexpappasoddball (commented):
Is there a way to not have Travis Hilton called out, but instead either the Utility Dev who is assigned to that sprint, or the tech lead, or even just the PM?

shiragoodman:
From what I've been told, we had individuals names included to make it more personal. I don't think it's necessary though. So are you proposing something like this:
`Back-end Engineers: Request Back-end orientation with @ Alex Pappas in Slack`
Or would you not one anyone's name listed?

alexpappasoddball:
I would be an acceptable name for sure.  Travis Hilton will conduct the interviews ONLY when he is on the utility dev role, so it makes more sense to funnel these requests through me or the tech lead (PM is fine!) to assign out to the appropriate team member as they are on the support role.

shiragoodman:
sounds good @alexpappasoddball . I'll update it to your name. @meganhkelley do you want to do the same?

meganhkelley:
@Clark-W-Griswold how do you want to handle the FE engineer orientation requests — have them Slack you directly? something else?

PR #19818, opened 2021-02-10
Reviewer comments on convo guide

lalexanderson-dsva:
Made a few comments in here to improve clarity. 

Of note, I think you need to confirm if you really want to test or evaluate "use my location" services in this study. You're giving participants a specific location, that is likely NOT where the participant will be when they are talking with you. Yet, you're also looking to see if they will use the "use my location" feature. If you're telling a participant who is joining you from Ohio to look for optometrists in San Francisco, and they use the "use my location" feature, it will locate them in Ohio, and they may be very confused. 

There's a lot of focus on the differences between the list view and map view. It seems like you're looking to test whether each of those views is useful/helpful to Veteran in finding a VA location, but the way you're phrasing questions, you'll likely get a lot of "well one is a map, and one is a list". If there is some specific difference in the content that shows in each view, recommend updating your questions or simply observing and listening as they interact with each view, asking what questions they have, what works what doesn't,  and noting for yourself behavioral patterns about whether they are aware of and affected by the differences.

### DIFF
--- a/products/facilities/facility-locator/mobile-experience/conversation-guide-v2.md
+++ b/products/facilities/facility-locator/mobile-experience/conversation-guide-v2.md
@@ -73,8 +73,11 @@ _For Android troubleshooting, try here: https://support.zoom.us/hc/en-us/article
 
 - Can you tell me a little about yourself or your service history?
 - Have you used the VA website before?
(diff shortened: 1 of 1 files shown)


laurellawrence (approved):

laurellawrence (commented):
Adding Lauren's feedback

laurellawrence (approved):

PR #20042, opened 2021-02-16
reorganize Demian's copy in ADR-0005

f1337:
@dginther wrote the original copy in a Slack conversation between @drorva, @jhouse-solvd and myself. Another related conversation preceded this one, involving Demian, Jesse, myself, @karlbrown-va, and @joeniquette.

This PR is just me re-organizing Demian's brain dump. The thinking is all from him & the Ops team, who have been working hard on this problem.

Let me also preface with: We really didn't want to abandon CircleCI for our own internal (VSP Platform) uses. But the security + maintainability impasse we arrived at does not scale, and is therefore untenable.

Questions & feedback are welcome. Please feel free to comment on this PR.

### DIFF
--- a/docs/adr/0005-use-github-actions-for-vsp-platform-ci.md
+++ b/docs/adr/0005-use-github-actions-for-vsp-platform-ci.md
@@ -1,40 +1,50 @@
-# Context
+# 5. Use GitHub Actions For VSP Platform CI
 
-https://dsva.slack.com/archives/C01CJV0L9PS/p1612651359029600?thread_ts=1612651359.029600&cid=C01CJV0L9PS
+Date: 2021-02-16
 
-TL;DR: CircleCI as implemented with VA restrictions does not allow easy management of permissions on ‘Contexts’ which contain credentials which are available to the entire ‘department-of-veterans-affairs’ Github Org. Although they are not displayed in plaintext in the CircleCI UI, they are obtainable in plaintext by running a build in CircleCI and using SSH to connect to your in-progress build, and outputting ENV variables. If the entire Org has access to the context, there is no way to stop anyone in the Org from doing so.
+Whom:
+- Demian Ginther
+- Jesse House
+- Michael Fleet
+- Dror Matalon
+## Status
+
+Accepted
+
+## Context
+
+[Slack discussion thread](https://dsva.slack.com/archives/C01CJV0L9PS/p1612651359029600?thread_ts=1612651359.029600&cid=C01CJV0L9PS).
+
+### CircleCI + VA Security
+
+CircleCI as implemented with VA restrictions does not allow easy management of permissions on ‘Contexts’ which contain credentials which are available to the entire ‘department-of-veterans-affairs’ Github Org. Although they are not displayed in plaintext in the CircleCI UI, they are obtainable in plaintext by running a build in CircleCI and using SSH to connect to your in-progress build, and outputting ENV variables. If the entire Org has access to the context, there is no way to stop anyone in the Org from doing so.
 
 In addition, CircleCI as implemented with VA restrictions does not allow use of ‘third-party’ Orbs (reusable code) which means that we cannot publish our own Orbs either.
 
-These two items are not preventing the use of CircleCI completely, as it is still possible to manage per-project secrets (only available to those with Github permissions on that repo) and you are free to cut and paste code from one workflow to another to re-use code, however, it makes managing multiple repos that have similar functions very tedious, and error prone.
+These two items are not preventing the use of CircleCI completely, as it is still possible to manage per-project secrets (only available to those with Github permissions on that repo) and you are free to cut and paste code from one workflow to another to re-use code, however, _it makes managing multiple repos that have similar functions very tedious, and error prone_.
 
 In addition, managing the secrets per-repo means that if the Operations team wanted to rotate the credentials for that repo, they would have to use the CircleCI UI to change all of those credentials, one project at a time.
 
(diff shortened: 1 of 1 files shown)


docs/adr/0005-use-github-actions-for-vsp-platform-ci.md
```diff
@@ -1,40 +1,50 @@
...
+
+Accepted
+
+## Context
+
+[Slack discussion thread](https://dsva.slack.com/archives/C01CJV0L9PS/p1612651359029600?thread_ts=1612651359.029600&cid=C01CJV0L9PS).
+
+### CircleCI + VA Security
+
+CircleCI as implemented with VA restrictions does not allow easy management of permissions on ‘Contexts’ which contain credentials which are available to the entire ‘department-of-veterans-affairs’ Github Org. Although they are not displayed in plaintext in the CircleCI UI, they are obtainable in plaintext by running a build in CircleCI and using SSH to connect to your in-progress build, and outputting ENV variables. If the entire Org has access to the context, there is no way to stop anyone in the Org from doing so.
```
karlbrown-va, line 20: I would add here that CircleCI contexts can be restricted based on github teams. However:
1) Only github organizational admins can add such restrictions - thus it must be explicitly explicitly requested
2) There isn't as yet a process for this; as such, it would be possible that even a restricted context could have an inappropriate team added to it, due to uncertainty around who 'owns' a context.

docs/adr/0005-use-github-actions-for-vsp-platform-ci.md
```diff
@@ -1,40 +1,50 @@
...
 
 In addition, CircleCI as implemented with VA restrictions does not allow use of ‘third-party’ Orbs (reusable code) which means that we cannot publish our own Orbs either.
```
karlbrown-va, line 22: I *believe* that the VA could decide to allow 3rd party orbs. It's still not clear to me who is blocking this.

docs/adr/0005-use-github-actions-for-vsp-platform-ci.md
```diff
@@ -44,24 +54,22 @@ what I’m saying is that if they love Circle and want to use it, we can support
 
 as long as they manage their own project-level credentials
 
-the problems with CircleCI for the Operations team is that we have many repos and managing those credentials at the repo level across all of those, as well as managing them all as separate pieces of code (when in reality they are 99% similar) is tedious, error prone, and makes Circle a bad fit (as long as the VA maintains those limitations)
-Github Actions is, I think, the best option.
+## Decision
 
- 
+Use GitHub Actions for VSP Platform CI.
```
karlbrown-va, line 59: This whole thing seems to be more about 'why we don't like CircleCI'. It doesn't necessarily address the weaknesses of Github actions - including difficulty in debugging complicated workflows which we've seen in other projects at TW.

docs/adr/0005-use-github-actions-for-vsp-platform-ci.md
dginther: I agree that they _could_ decide to do this, but I am not sure that they will at this point, and the functionality we need is already included in Github Actions re: creating reusable workflows/actions

docs/adr/0005-use-github-actions-for-vsp-platform-ci.md
dginther: Say more about this. For the most part our workflows will be exceedingly simple; i.e.: 
- Build a container or mirror a container from quay/dockerhub to ECR
- update a Kubernetes Manifest
- Make a release in Github

Interested to know what issues you have seen.

karlbrown-va: Don't you plan to build a VSP/VA.gov release pipeline? Similar for other adjacent products?

I'll find some stories for you and share them here re: complex pipelines.

dginther: We do not plan to use any of the ansible Build/Release/Deploy system that is currently in Jenkins. What would a VSP/VA.gov release pipeline look like in your mind?

dginther: Keep in mind, we have settled on a CD product, with ArgoCD, we are not interested in complexity in the CI pipelines, they should _just_ be for testing, building containers, producing GH releases, and updating k8s manifests.

dginther: > This whole thing seems to be more about 'why we don't like CircleCI'.

This is absolutely true, this was the impetus for making this decision. In fact, in my initial conversation, I said "I think the ADR for this would be an ADR to NOT use Circle for our own use, and the factors leading up to that, which is how I have been approaching it."

karlbrown-va: "they should just be for testing, building containers, producing GH releases, and updating k8s manifests." - got it, understood, that makes sense. The testing part might be more or less complex esp if you want to have some way of ensuring certain tests are run across a broad variety of players - like security scans for example - but I take your point that you want this to just focus on CI.

U-DON:
> Each application has it’s own repository, it’s own ECR repo, and it’s own AWS user, with limited permission scope

As `vet-website` is currently hosted with S3 and routed with nginx, what would it look like to deploy it in this new framework?   I imagine the actual build would continue to be hosted in S3, but the nginx server would be containerized and deployed with ArgoCD. Is that accurate?

Just trying to understand what impact this might have on FE Tools.

docs/adr/0005-use-github-actions-for-vsp-platform-ci.md
f1337: > I think the ADR for this would be an ADR to NOT use Circle for our own use…

This confusion is my doing, I honestly missed that distinction and mis-titled the reformatted ADR. I have since corrected it to, I hope, reflect Demian's intent better.

docs/adr/0005-use-github-actions-for-vsp-platform-ci.md
drorva: I would add:
* Github is already ATOed
* Github is fully supported and paid for by the VA, so there should be no licensing issues.

PR #20803, opened 2021-03-03
Update claims-and-appeals-status.md

erikphansen:
Please confirm that this new case is accurate


tressaellen:
Yes this new case is accurate.

Samara-Strauss (approved):

PR #21760, opened 2021-03-19
updated to include user create drush command

erogray:
I changed the drush command to be the create user, but also included the password change as an option


xiongjaneg (approved):

erogray:
No longer needed since change was already made

PR #22336, opened 2021-03-29
Update design-intent-collaboration.md

allison0034:
Added Ryan's name


shiragoodman (changes requested):
Thurlwell*

shiragoodman:
@allison0034 can you make this same change in the Touchpoint documentation?

PR #23059, opened 2021-04-12
update to broken-link-process.md

joshworkscivicactions:
Modified with additional descriptive language, with an example of a broken image alert, and removed redundant information, and some grammatical changes.

### DIFF
--- a/platform/cms/helpdesk/broken-link-process.md
+++ b/platform/cms/helpdesk/broken-link-process.md
@@ -1,10 +1,18 @@
-# Help Desk Process for Broken Link Resolution
+# Help Desk Process for Broken Link Ticket Resolution
 
-## Broken Link Alerts
+## Overview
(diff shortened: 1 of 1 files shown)


erogray (changes requested):
I don't believe that lines 26-34 are clear/accurate. `null` in this context means that if the link alert shows "x broken links found in the vagovprpod build on null" it's about a content release (a content-only deployment) instead of a code deployment. It's true that this is usually because of publishing of content. But the words "content release" are more useful. and I haven't seen "cannot read property of `null`" show up in these before.

erogray (changes requested):
Nitpicky, but on line 98/99, I think "See a suggested email template below:" is actually clearer than "An example of what the email communication to the User can look like:"

erogray (changes requested):
in line 31/52, " The first responsibility of the Tier 1 Help Desk is to analyze what the root cause of the issue could be, whether content or code related.  " is incorrect. Our first responsibility is to assess whether we need Tier 2. This may or may not have anything to do with whether it's a code deployment or content release. I think my previous wording is more accurate.

erogray:
Closing since there are a lot of questions and/or changes needed. Let's revisit in a coworking session.

PR #23477, opened 2021-04-19
Update product-outline.md

Dene-human:
@ala-yna I created a branch because I started to wonder if you really wanted me to write over the existing.


Dene-human (commented):
please check updated outline @ala-yna

PR #23524, opened 2021-04-20
Update overview_wip.md

joshworkscivicactions:
added Daily Shift Update section


erogray (commented):
in line 520, the mentioned slack channels are mis-identified; `cms-team` is on the VA slack instance, and `proj-vagov-helpdesk` is on the A6 slack instance (but shared with CivicActions). This isn't super important to get right but I thought I'd note it.

Issue #23762, opened 2021-04-26
Update the Liquid Template Testing Framework documentation with information about how to use the DOM Testing Library


holdenhinkle:
## Issue Description
Update the Liquid Template Testing Framework with information about how to use the [DOM Testing Library](https://testing-library.com/). 

This ticket is derived from [ticket #21651](https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/21651)

---
## Tasks
- [x] Update the Liquid Template Testing Framework documentation 

## Acceptance Criteria
- [x] The Liquid Template Testing Framework documentation is updated

holdenhinkle:
PR - https://github.com/department-of-veterans-affairs/va.gov-team/pull/23764

This ticket is complete.

pjhill:
Nicely done -- thanks for the quick review, @billfienberg !

@holdenhinkle, Have we notified the platform personnel / teams that were interested in this capability?

holdenhinkle:
Yes, I notified Michael Pelz-Sherman and he's spreading the word.

pjhill:
Merged and requesting team members notified

PR #23764, opened 2021-04-26
add information about using DOM Testing Library

holdenhinkle:
Resolves [ticket #23762](https://app.zenhub.com/workspaces/vsp-5cedc9cce6e3335dc5a49fc4/issues/department-of-veterans-affairs/va.gov-team/).

Adds information about using the DOM Testing Library with the Liquid Template Testing Framework to the framework documentation.

### DIFF
--- a/platform/testing/liquid-templates/README.md
+++ b/platform/testing/liquid-templates/README.md
@@ -48,6 +48,21 @@ This technique can be used to generate tests of varying complexity, ranging from
 ### `axeCheck(container)`
 `axeCheck()` takes the `HTML` document returned by `renderHTML()` and returns an array of accessibility violations.
 
+### DOM Testing Library
+
+To perform queries using the [DOM Testing Library](https://testing-library.com/), simply import the queries or functions you want to use. Example:
+```
+import { getByText } from '@testing-library/dom';
+```
+
+From the [Dom Testing Library docs](https://testing-library.com/docs/queries/about): "All of the queries exported by DOM Testing Library accept a container as the first argument."
+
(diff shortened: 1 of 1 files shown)


platform/testing/liquid-templates/README.md
```diff
@@ -164,6 +164,21 @@ The liquid template for the smoke test is located at `src/site/layouts/liquid_te
...
+```
+
+From the [Dom Testing Library docs](https://testing-library.com/docs/queries/about): "All of the queries exported by DOM Testing Library accept a container as the first argument."
+
+In the Liquid Template Testing Framework examples found below, the `HTML` document generated is assigned to a variable called `container` in the `spec.js` files. Use this variable as the first argument to the imported DOM Testing Library functions, like this:
+
+```
+const node = getByText(container, '3500 Ludington Street');
+```
+
```
billfienberg, line 181: It might be worth adding a line or two about the difference between the `getBy...` and `queryBy...` queries. 

When I first learned about them, I lost hours trying to assert that something was absent because I was using the wrong query. 

https://testing-library.com/docs/queries/about#types-of-queries

>- `getBy...`: Returns the matching node for a query, and `throw` a descriptive error if no elements match or if more than one match is found (use `getAllBy` instead if more than one element is expected).
>- `queryBy...`: Returns the matching node for a query, and return `null` if no elements match. **This is useful for asserting an element that is not present.** Throws an error if more than one match is found (use `queryAllBy` instead if this is OK). 
```

billfienberg (approved):
LGTM

platform/testing/liquid-templates/README.md
holdenhinkle: Thanks for the comment. I don’t think I want to include that information in this doc. Perhaps we can create a doc about the Testing Library itself at some point which would be a great place for tips and tricks, etc…

PR #24096, opened 2021-05-03
Update password-policy.md

joshworkscivicactions:
Submitting for review and feedback.  Rachel this might be something we move to User Administration? Or if you have a document similar this may be redundant?


erogray (changes requested):
I recommend that we not include any mention of temporary passwords; on this project we technically don't share temporary passwords with anyone (except perhaps for each other); we create non-usable passwords, then we _don't_ share them, and ask people to use their PIV to login on prod, and then re-set pw on their own.

PR #24140, opened 2021-05-03
Create customer-support-time-tracking.md

jonwehausen:
Let me know y'alls thoughts on the Toggl tags, if there's any you think that need to be added / removed / renamed

Please also checkout out the [DOMO connector dataset](https://va-gov.domo.com/datasources/419ddcb3-5a3c-4316-aab0-437cf691215f/details/overview) and [ETL])(https://va-gov.domo.com/datasources/365fe473-d9d6-458d-ae1e-d59989107988/details/overview)

The ETL currently does two things: 
1) Filters for only our time entries (myself, Brian, Lauren) 
2) Does a substring method for the first 5 characters of the description in attempt to capture the GH ticket number
   - _We'll need to be diligent about preceding with the GH issue number and delimiter to return these, open to other simpler solutions you all may have_

### DIFF
--- a/platform/analytics/customer-support-time-tracking.md
+++ b/platform/analytics/customer-support-time-tracking.md
@@ -1,0 +1,49 @@
+## Toggl Customer Support Tracking  
+
+### Purpose
+- To enable the Analytics & Insights team to track time + efforts spent supporting customer support requests over time, we would like 
+to standardize the way we track our customer support 
+
+### Background
+- Toggl is the current solution Empirical Path contractors utilize for time tracking on the VSP contract
+- Toggl has **three fields** to classify for time tracking:<br>
+      - _Description_: Free-text field description of work performed<br>
+      - _Project_: Pre-determine field to identify the project<br>
+      - _Tag_: Customizable field to provide additional context / categorization to work performed
+ - Toggl includes a start + stop button to signal time spent
+ 
+ 
+ ### Customer Support Time Tracking Process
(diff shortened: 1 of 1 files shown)


jonwehausen:
cc: @joanneesteban for visibility

laurensimpson14:
I think it depends on our goals for tracking time. Is the goal to see what tickets take the most time? To see what internal teams take up the most time w/ticket requests? If so then the ticket number makes a lot of sense to me. 

Are we also interested in tracking how much time is spent w/meetings, slack support, dashboards, etc.? If so then I think we should add the following "Tasks" within Toggl to the VSP Ongoing Support Project: 
- Slack Support & short one-off questions (including office hours)
- Implementation
- QA 
- Internal meetings
- Onboarding/Offboarding/Access requests
- Domo 
- GA
- Misc (as this grows we'll know what other big buckets need a "Task" created)

Both approaches would also be good for Joanne to easily see how each of us are trending on time to be sure we're staying at the right pace - which would help her with assigning tickets.

I think your plan @jonwehausen is perfect if our goal is to know time spent within internal teams & on specific tickets, but may not work as well if the goal is just to get an idea for general time spent on various tasks. Just depends on the goal here - I'm good with whatever we land on. What was the goal that Joanne had in mind when this topic first cropped up?

jonwehausen:
Thanks for y'alls input @laurensimpson14 + @joanneesteban 

This has merged to master and should begin being implemented as early as Sprint 52.

PR #24495, opened 2021-05-10
Update topline-summary-template.md

laurylzenobi:
updating template


sshein (commented):
Is there a reason we only list 5 things? I feel like I usually have around 7-8 themes in a topline. Maybe we can change it to 5-10 findings

PR #24908, opened 2021-05-18
Forms library design document

cvalarida:
You can view the latest rendered markdown [here](https://github.com/department-of-veterans-affairs/va.gov-team/tree/forms-library-design-doc-cv/platform/engineering/design-docs/2021-05-18-forms-library.md).


platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+    it some default props for `initialValues`, `validate`, and `onSubmit`
+  - If the save-in-progress React context is available, `Page` will use it to
+    auto-save the form
+- Render the children, passing them the extra props for form state
+  - For an example usage, see [below](#functionality-array-pages)
+  - **Caveat:** This means when the children of `Page` are a collection of
+    `<Field>`s (or similar) instead of the functional component in the example
+    below, they'll all have "extra" props for the form context
+    - This is the simplest way to support both simple and complex form pages,
+      but may be subject to change later if we can figure out a better way
```
bkjohnson, line 206: I'm not sure which "extra" props you're talking about - can you be a bit more specific?

This is also confusing in general, because the example below has a `<Field*>` inside of a `<Page>`, but this text reads as if there is a _different_ use of `<Field*>` inside of `<Page>` where these extra props won't be used.

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+- Render the form navigation buttons "Back" and "Continue"
+- Prevent navigation if there are validation errors
+
+##### `Route` Component
+**Purpose:**
+The `Route` will be re-exported from `react-router-dom`. The purpose of
+re-exporting is twofold:
+- Simplifies imports
+- Ensures only one version of `react-router-dom` is used
```
bkjohnson, line 215: What is going to keep someone from importing from `react-router-dom`, since it would be in `node_modules` as a dependency of `formulate`, if I'm understanding this correctly.

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+```
+
+For array pages to work we need a couple core things:
+- The path for the page
+- The data for the current item in the array
+
+To get these pieces, `Page` will accept the `path` and `pathParams` props.
+
+`path` will be the URL within the application and must (for array pages) contain
+URL parameters (term pending). URL parameters must be the only token between
```
bkjohnson, line 258: Would "slug" get the point across?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+of a way to automatically build the review page and meet all of the following
+requirements:
+- The chapters are divided correctly
+- The pages are divided correctly
+- The edit mode for each page is identical to the page in the form flow
+- The review mode for each page displays each field on the page in a data table
+  with the appropriate label and value
+- Each field in the review mode may be overridden when the label or value need
+  to be changed for a better user experience
+- Each page may or may not have a review mode
```
bkjohnson, line 280: This may not be the right place, but I'm curious to learn more about this even though I've known about it for a while. Do solutions for some of these requirements complicate the solution for others?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+- [Re-usable `FormSaved` component](#component-formsaved)
+  - To be used in a `<Route path="/form-saved">`
+- [Data migration functionality](#data-migrations)
+
+##### Component: `SaveInProgress`
+**Purpose:** This is the primary component for enabling the save in progress
+functionality for a form.
+
+**Usage requirements:**
+This component must be wrap the `Router` or `Formik` components so we can pass a
```
bkjohnson, line 310:
```suggestion
This component must wrap the `Router` or `Formik` components so we can pass a
```

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+    - Save the form when `saveForm` is called
+  - `loadForm` function
+    - Call the supplied endpoint with the supplied arguments
+    - Set the form data to the form data in the result
+    - Perform any data migrations applicable
+    - Navigate to the page supplied by the result
+- Accept a [`dataMigrations`](#data-migrations) prop to be used with `loadForm`
+
+##### Component: `SiPIndicator`
+**TODO:** Rename this thing. Ideas?
```
bkjohnson, line 327: `<SaveStatus>`?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+```
+
+**Single validate function for the page:**
+```js
+import fromSchema from '../schema-validation';
+
+const validate = values => {
+  const errors = {};
+  
+  const passwordErrorMessage = fromSchema('password'), // Path to schema
```
bkjohnson, line 441: Is that trailing comma necessary, or am I misunderstanding what this syntax is doing?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+Implementing such a mistake would look something like this using the
+`fromSchema` validation:
+```jsx
+<StringField
+  name="email-confirmation"
+  validation={fromSchema('email.confirmation')} />
+```
+
+**Discussion:** Is not ensuring the shape of the data matches the schema a
+reasonable boundary for the library to stop at?
```
bkjohnson, line 518: I think so, partially because providing that kind of developer-facing validation seems challenging based on the structure I've seen so far. The return value of `fromSchema('email.confirmation')` would have to include its own argument in order for us to do a comparison. It looks like `validate` is expecting that return call to be a function, so returning multiple values might allow this kind of validation to work, but I'm not sure if it's worth the complexity.

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+
+**Discussion:** What tools might Formulate give to developers to help catch
+these kinds of data shape mismatch mistakes?
```
bkjohnson, line 521: A linting rule, perhaps?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+  - Getting started
+  - Building a multi-page form
+- [How-to guides](https://diataxis.fr/how-to-guides/)
+  - How to build a review page
+  - How to make array pages
+  - How to make single-page list loops
+  - How to build a custom field
+  - How to write a data migration
+- [Reference documentation](https://diataxis.fr/reference/) tooling
+  - See[Reference documentation](#reference-documentation) below
```
bkjohnson, line 541:
```suggestion
  - See [Reference documentation](#reference-documentation) below
```

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+- [Explanations](https://diataxis.fr/explanation/)
+  - Why keep form state in a React context for multi-page forms?
+
+All documentation will be located within the Formulate repo in the `docs/`
+directory. It is essential that documentation and code live side-by-side so the
+documentation may be updated at the same time as the code. Additionally, keeping
+all the documentation in `docs/` allows us to serve it simply using GitHub Pages.
```
bkjohnson, line 548: Will we be serving it on design.va.gov?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+[`api-documenter`](https://github.com/microsoft/rushstack/tree/master/apps/api-documenter).
+See the
+[`rushstack.io-website`](https://github.com/microsoft/rushstack.io-website) repo
+and [corresponding website](https://rushstack.io/pages/api/).
+
+The built reference documentation will live in `docs/reference/`.
+
+#### Explanations
+Architecture decisions and other explanations for decisions made for the
+library will live in `docs/explanations/`.
```
bkjohnson, line 581: I'm assuming we will link back to this document as well?

bkjohnson (commented):
Once I got to the Work Estimates I began skimming, but I didn't notice anything in the doc about getting the data out of each of the `<Form>` components and into something that is ready to submit with fetch. Should that be covered?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+forms library will be called **Formulate**.
+
+#### Problems
+
+##### Library bloat
+The current forms library uses a single form configuration object to create the
+entire form application. Modifying this `formConfig` object is the only way to
+change anything in the form. When new features for individual forms are needed,
+a change to the underlying library is required. This creates significant bloat
+in the library and makes it difficult to modify.
```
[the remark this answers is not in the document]
timwright12: Looking for more system-level autonomy in individual forms

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+
+The form configuration approach is not one that is familiar to front end
+engineers to create applications and requires _much_ prior knowledge to use
+effectively. When learning the basics of the library, its complexity is
+relatively minimal, but quickly grows as the form becomes more complex. This is
+largely due to the library bloat outlined above.
+
+##### And more!
+There are many other problems introduced by the existing forms library, but for
+now, I digress.
```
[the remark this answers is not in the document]
timwright12: I would list them if possible, that way you'll have a more concrete problem statement

cvalarida:
> Once I got to the Work Estimates I began skimming, but I didn't notice anything in the doc about getting the data out of each of the <Form> components and into something that is ready to submit with fetch. Should that be covered?

Yeah, that's a good call. For routed forms, when the user goes forward or backward, it'll merge the data gathered on the page with the form state in the React context. Submission then just takes that data from the context and is up to the engineers to make whatever transformations they need (ideally none) and submit it to whatever endpoint they need it to go to.

Non-routed forms will just use the [normal Formik submission function](https://formik.org/docs/guides/form-submission).

I'll write that down in the doc. :+1:

platform/engineering/design-docs/2021-05-18-forms-library.md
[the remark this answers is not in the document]
cvalarida: Can you think of examples of risky changes, @bkjohnson? The only ones I can think of right now were just bloat, not risky.

@timwright12 I think that's what I'm talking about in the next problem, `Inflexibility`. I like your phrasing of the problem better. :+1:

platform/engineering/design-docs/2021-05-18-forms-library.md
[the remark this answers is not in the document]
cvalarida: Yeah, that's a good call. Only trouble is it's been so long since I've used the library, I can't remember them all, so this is where I phone a friend or...several. :sweat_smile: 

What other problems do you run into, @jbalboni, @Mottie, @LinnJS (and whomever else)?

I know there are some underlying assumptions that aren't always true for all forms these days:
- Must have an introduction page
- Must have a review page
- Must have a schema
- ...what else?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+The following sections outline the various parts of the system.
+
+#### Form page builder
+Formulate will leverage Formik for much of the form page building. It will
+**re-export Formik components** where it makes sense and **provide light
+wrappers for Formik input fields** using design system components.
+
+In addition to simple wrappers, Formulate will use Formik to provide more
+complex fields such as file upload which may be imported and used in
+applications.
```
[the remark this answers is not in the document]
cvalarida: Ideally, all fields offered by Formulate will be wrappers around design system components, but I hesitate to say that there will be no Formulate field that will _not_ be a wrapper for a design system component. (Not sure if that answers your question. :sweat_smile: )

But yeah, I don't see a place for the concept of a form widget as it exists in the current library.

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+
+##### Why not just use Formik directly?
+It should be possible to use Formik directly without any issues. The form page
+builder sub-module intends to make using the design system with Formik simpler.
```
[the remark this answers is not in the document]
cvalarida: So there are two answers to your question.

**Philosophically**
It will not be the responsibility of Formulate nor the design system team to enforce the usage of only Formulate fields, if that's your question. We can highly recommend it, however, because of the benefits teams will enjoy in the form of:
- Simple usage (better DX)
  - I see now I need to add a bit in here about wrapping the label and error message as well; this is where the simpler usage comes in
- "Free" bug fixes, a11y improvements, and design updates (when applicable)

At the end of the day, if a team wants to totally deviate from the design system, they own their product and are responsible and accountable for it. :man_shrugging: We simply provide the tools to make design system adoption easy and desirable.

**Technically**
Also no. :sweat_smile: There's not really a way I can think of to stop folks from just using `<Field>`.

platform/engineering/design-docs/2021-05-18-forms-library.md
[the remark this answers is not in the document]
Mottie: - The path must always start on the introduction page (wizards may need their own path eventually)
- Schema in vets-json-schema isn't the same schema you'll use to build the form
- Form validation errors based on schema do not match visual representation
- List loop (ArrayField) component isn't customizable - I've copy/pasted and modified the same component into multiple forms now
- Checkbox error messaging not working as expected
- Checkbox group pattern does not handle error messages

And these are just the ones I've encountered recently...

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+- Manage the form data
+  - Each individual page is essentially its own mini-form; the router stitches
+    all the data together into a single cohesive experience
+
+To accomplish this, there will be four primary pieces in the routing
+sub-module:
+- [`Router` component](#router-component)
+- [`Chapter` component](#chapter-component)
+- [`Page` component](#page-component)
+- [`Route` component](#route-component)
```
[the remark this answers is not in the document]
cvalarida: :thinking: It might make sense to call it out earlier in the document, but it probably makes more sense to call it out right here. :man_shrugging: 

I'll make it so. :+1:

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+##### `Router` Component
+**Purpose:**
+The `Router` is at the heart of the routing sub-module. It will hold the routes
+and components to render at each route.
+
+**Usage requirements:**
+_None_
+
+**Responsibilities:**
+1. Provide a React context with the form state
```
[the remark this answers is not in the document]
cvalarida: Ah yeah. It's a big "C" Context. I'll make it a link. :+1: 

And yes, the plan is to not use redux. When I first started writing this document, I was going to use redux, but as I built the system out in my mind, I discovered we really don't need it. In fact, I expect it'd be more bloat than we need since we'll be using Formik.

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+    - `Page` will use this as the page's
+      [`initialValues`](https://formik.org/docs/api/formik#initialvalues-values)
+      - This is because each page is a separate Formik form, so the `Router` is
+        responsible for stitching all these individual forms together into one
+        cohesive experience
```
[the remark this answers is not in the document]
cvalarida: Good call out. The React Context will contain the unified form data, so stitching it together (deep merging page data with the whole form data) will happen between page navigation. Submission will simply take the form data from the context. I'll make sure to talk about merging the page data into the form state when navigating between pages.

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+      ```
+      This will result in a path definition of `/personal-information/name`
+1. Re-compute the list of routes before each page navigation (below)
+    - This ensures any data changes that might affect the routes are captured
+1. Provide callbacks in the React context to navigate forward and backward
+   through the list of computed routes
+
+##### `Chapter` Component
+**Purpose:**
+The **optional** `Chapter` component will be used to render the VA.gov "chrome"
```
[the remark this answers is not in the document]
cvalarida: Good call. :+1:

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+for multi-page forms which have multiple chapters. This "chrome" includes the
+form title, current chapter name, and progress bar.
```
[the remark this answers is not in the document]
cvalarida: Yeah, that's the struggle I had. :sweat_smile: I'm open to suggestions!

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+
+##### `Router` Component
+**Purpose:**
+The `Router` is at the heart of the routing sub-module. It will hold the routes
+and components to render at each route.
+
+**Usage requirements:**
+_None_
+
+**Responsibilities:**
```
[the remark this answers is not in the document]
cvalarida: I guess `Purpose` is more like `Description`. It's meant to be a brief description of what the component does. Its `Responsibilities` are a list of things it will do.

I'm open to changing up the format, but the idea here was to split out answering the "What is it?" question and the "How does it work?" question. I decided not to go with `What is it?` and `How does it work?` because I didn't want to get _too_ low-level. I felt like it was going to be wordy enough as it was. :sweat_smile: :grimacing:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: > I'm not sure which "extra" props you're talking about - can you be a bit more specific?

I can call out the two that I know for sure we'll need, but I want to leave it open-ended for adding more in later because I don't know what _all_ we'll need until we get to implementing it.

> This is also confusing in general, because the example below has a `<Field*>` inside of a `<Page>`, but this text reads as if there is a different use of `<Field*>` inside of `<Page>` where these extra props won't be used.

Yeah, that's fair. This section went through a few iterations and I may not have cleaned it up very well. Basically, it goes like this:
1. You can pass children to `<Page>` just like you would for any other component
    ```jsx
    <Page>
      <StringField />
      <NumberField />
    </Page>
    ```
2. But if you needed to know some information about the form state, like, for example, what array page we're on, you can get that data from `<Page>` by passing a custom component that accepts the props describing the form state
    ```jsx
    <Page>
      {({ formData, pathParams }) => ( /* Render your fields using the data from the form state /* )}
    </Page>
    ```

The alternative to this is importing the React Context that `Router` provides and doing something like
```jsx
import { useContext } from 'React';
import { routerContext } from 'formulate`;

export default () => (
  <Page>
    {() => {
      const { formData, pathParams } = useContext(routerContext);
      return (/* Render fields using the form state */);
    }}
  </Page>
);
```
I opted to go with the former approach because I thought it'd be a better developer experience. I'd love to hear from VFS engineers, though!

That said, I'll go back over this and try to clean it up to make it clearer. :+1:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Nothing. :man_shrugging: But if they import the component from the same dependency as Formulate, it'll be the same version, so that's fine.

The problem I foresee is having a peer dependency of `react-router-dom` that doesn't match the version Formulate uses. So if we try to use the `<Route>` from the `react-router-dom@1.0.0` specified in the application's `package.json` in the `<Router>` from Formulate, which uses the `Router` from `react-router-dom@2.0.0`, we may have a problem.

This library won't attempt to prevent engineers from doing anything. It'll attempt to provide engineers with the tools they need to do what they need the right way. In this case, it means ensuring the `Route` is using the same version as the `Router`.

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Isn't a slug a different concept? I thought that was the last portion of the URL's path.

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: :thinking: Maybe?

In the context of this document, the struggle I had was doing this automatically when there is no easily digestible single source of truth for all pages. In the schema-based forms library, we had access to all the configuration needed to build all the fields on all the pages. With this new approach, that data lives in the props passed to the fields, so we'd have to somehow start at `<Router>` and iterate through all its children to get all the chapters and pages, all their fields, and build out the review mode for each page, making lots of assumptions along the way, _and_ then provide hooks in for engineers to override the automatically-built stuff when it's wrong for whatever reason.

So at least here, the problem wasn't that the solution to some of the problems complicated the solution to other problems (thought that's probably true). The problem was more just doing it at all. :sweat_smile:

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,852 @@
...
+Instead, Formulate will supply components to aid engineers in rapidly building
+review pages. These components will include:
+
+- `FormReview`
+- `ChapterReview`
+- `PageReview`
+- `FieldReview`
+
+**TODO:** Figure out how submission validation plays in here. How do we run all
+validation on all pages?
```
cvalarida, line 291: I'm still not sure how to do this. The follow-up here is also displaying all the error messages.

I might be able to do something clever like pass a `validateAllPages` function to the React Context the `Router` provides which iterates through all the `Router`'s descendants, looking for `Formik` components and.....somehow grabs its validation function...slaps them all together in an array or something, runs them all, and associates all the errors with the individual pages. I'm not sure. I'm dubious of that approach—I suspect array pages make this even more difficult—but maybe?

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Yeah, I thought about it, but it also displays the `Save your progress` link. (Or whatever the text is.) So it's not _just_ status.

Probably still better than `SiPIndicator`, so if I don't get any better suggestions, I'll change it to that, but I'll hold out for a while for better suggestions. :sweat_smile:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Oh, nope. That's a typo. Good catch. :male_detective:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Oh, that would be ideal. I'm not sure if it's feasible, though, since that schema may (probably will) come from an external package like `vets-json-schema`. It's worth exploring, though. :+1:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: No, probably not. I think design.va.gov is probably not the right place for all this developer-facing documentation. For simplicity, we'll stick to GitHub pages and link to that wherever it makes sense (like the project README).

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Yeah, that's a good idea! :raised_hands:

platform/engineering/design-docs/2021-05-18-forms-library.md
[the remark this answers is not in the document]
bkjohnson: I think I understand the difference now with a bit more noodling.

platform/engineering/design-docs/2021-05-18-forms-library.md
bkjohnson: So this re-exporting doesn't necessarily ensure that only one version of `react-router-dom` is used? I don't think we need to try to make that a thing, but I was curious about the statement itself given some past experiments.

There may be an assumption here that's interfering with things. For the build of `formulate` that gets published, will it already have the relevant stuff from `react-router-fom` in the bundle? When I made the original comment I was thinking that that wouldn't be the case.

platform/engineering/design-docs/2021-05-18-forms-library.md
bkjohnson: I think a slug is the entire portion of a URL after the domain.

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Assuming tree shaking is working properly in the application using Formulate, if it doesn't import `Router` or any of the routing components, the application bundle will not include `react-router-dom` because of Formulate. When an application imports `Router` or any of the routing components, the application bundle will include (parts of? :thinking:) `react-router-dom` as a dependency of the Formulate package.

platform/engineering/design-docs/2021-05-18-forms-library.md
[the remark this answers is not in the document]
cvalarida: (Bah, apparently I can't resolve this whole thread? :confused:)

platform/engineering/design-docs/2021-05-18-forms-library.md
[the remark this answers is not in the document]
cvalarida: I added a bunch of bullet points based on your feedback. Thanks kindly! If there are any more you'd like to call out, please leave a comment in your review. :grin:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: I added some more clarity to this section. If it's still confusing, feel free to leave another comment on the new part. :smile:

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,927 @@
...
+  path="/marriages/:marriageIndex/foo"
+  pathParams={{
+    marriageIndex: 'marriages'
+  }}>
+  {({ formData, pathParams }) => (
+    <FieldArray
+      name="marriages"
+      render={arrayHelpers => (
+        <Field name={`marriages.${pathParams.marriageIndex}.name.first`} />
+      )}
```
Mottie, line 320: It isn't clear what `arrayHelpers` is doing in this function.

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,927 @@
...
+To get these pieces, `Page` will accept the `path` and `pathParams` props.
+
+`path` will be the URL within the application and must (for array pages) contain
+URL parameters (term pending). URL parameters must be the only token between
+`/`s or at the end of the URL. For example, `/:index/foo` and `/:index` are
+valid path parameters. `/my-:index-item` is an invalid path parameter.
+
+`pathParams` will be an object whose keys map to the path parameters and whose
+values map to the path to the form data. In the above example, the `pathParams`
+will assign `:marriageIndex` in the URL a number based in the current item in
```
Mottie, line 339: can the index be named anything? Is the only requirement the `:` at the beginning? Or does the word "index" have to be included?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,927 @@
...
+validation on all pages?
+
+#### Save-in-progress
+Pieces of the save-in-progress sub-module are:
+- [`SaveInProgress` component](#component-saveinprogress)
+  - Adds `SiPIndicator` to each page
+  - Provides functions to save and load the form
+- [`SiPIndicator` component](#component-sipindicator)
+  - Allows users to manually save the form
+  - Shows the form save state (saved, saving, error, etc.)
```
Mottie, line 375: Can we also include & show the SiP/application ID. So the users can provide a non-PII ID to the help desk allowing engineers an easy method to look up the SiP data.

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,927 @@
...
+- The page URL when the form was saved
+
+Each function must return the new form data and metadata. If multiple migration
+functions must be run, the result of an earlier migration will be passed to the
+next migration in the array.
+
+Migrations are run during the `loadForm` function. It will check the form
+version in the metadata against the length of the `dataMigrations` array. If the
+latter is greater, all migrations after the saved form version will be called in
+order.
```
Mottie, line 465: For form 526, the list of rated disabilities is reloaded upon each form start, and if the loaded list doesn't match the SiP list, we redirect them back to the rated disabilities page so that they know a change happened and that they need to review their options. So, maybe this `dataMigrations` can allow a way to reroute, or change the `page URL` metadata

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,927 @@
...
+**Single validate function for the page:**
+```js
+import fromSchema from '../schema-validation';
+
+const validate = values => {
+  const errors = {};
+  
+  const passwordErrorMessage = fromSchema('password') // Path to schema
+    .minLength("This, uh, isn't very safe...") // Optional error message overrides
+    .pattern('Please follow the pattern, will you?') // Chaned message override functions
```
Mottie, line 518: chained?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,927 @@
...
+**[Validate a single field:](https://formik.org/docs/guides/validation#field-level-validation)**
+```jsx
+import fromSchema from '../schema-validation';
+import { StringField } from 'formulate';
+
+const ThisPage = () => (
+  <>
+    <StringField
+      name="name.first"
+      validate={fromSchema('name.first').pattern("Please enter your _real_ name.")} />
```
Mottie, line 533: Will the `pattern` function accept markdown? Can we pass it JSX?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,927 @@
...
+**Data**
+```js
+{
+  email: {
+    original: "foo@bar.com",
+    confirmation: "foo@bar.com"
+  }
+}
+```
+Formulate will not deliberately catch mistakes like this, but it should be
```
Mottie, line 579: Maybe showing a "correct" json schema that matches this data would be helpful?

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: This [comes from Formik](https://formik.org/docs/api/fieldarray#fieldarray-helpers). In this function, it isn't doing anything. I opted to keep the example simple and not use it. :man_shrugging: 

I'll either add a link to the `FieldArray` Formik documentation to clarify the purpose, or simplify the example. :+1:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Yeah, whatever you want to name it. The only requirement is that the URL segment start with a `:`. I'll clarify in the document. :+1:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: If this is data that's returned by the API, we 100% can show it. :+1:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Oooh, this is interesting. That isn't a data migration since the form itself didn't change. The data from an external source did. It sounds like you need the ability to provide your own `loadForm` function (using the one provided by Formulate, likely). One option is to consume the form state React Context and override the `loadForm` function, but that seems cumbersome. I'll see if I can't figure out a cleaner way for you to supply your own `loadForm`. :+1: 

This is a great call out, thanks! :raised_hands:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: All the cool kidz are chaning these days. :homer-disappear:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Oh, right. I meant to write this part down but forgot. Yeah, the error message parameter for all these functions will accept a string, React component, or function which returns a string or React component.

:thinking: JSX may or may not have to be returned from a function if I can't figure out how to distinguish between a vanilla function or React component, but that's an experiment I need to run.

Mottie:
One that I just thought about... can we also preload & include the feature toggles data along with the form data?

cvalarida:
> can we also preload & include the feature toggles data along with the form data?

This likely isn't something the library will do, no. You should be able to do that in your application code, however, and use it as you see fit.

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: After thinking about it some more, I think the best developer experience will be to provide a `postLoad` function that occurs between data migration and routing to the `metadata.returnUrl`. This can be an `async` function in which you can make an external service call (fetching the latest rated disabilities, for example) and modify either the form data or metadata (re-routing back to the rated disabilities selection page, for example).

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,942 @@
...
+builder sub-module intends to make using the design system with Formik simpler.
+
+#### Routing
+This sub-module will be responsible for the user navigating through a multi-page
+form. In particular, it will:
+- Present the form broken down into chapters and pages
+- Manage routes for each page
+- Render each page wrapped in the chrome appropriate for navigation
+  - Form title, page title, progress bar, back / continue buttons, etc.
+- Prevent progress through the form if there are validation errors on the
```
jbalboni, line 153: This is something that the form page builder module probably needs. Or at least, anyone using the form page builder on its own will need this functionality.

platform/engineering/design-docs/2021-05-18-forms-library.md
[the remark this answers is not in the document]
jbalboni: "Form-wide layout elements" might work?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,942 @@
...
+parameter. The name of the path parameter may be anything, but it must start
+with a `:`.
+
+`pathParams` will be an object whose keys map to the path parameters and whose
+values map to the path to the form data. In the above example, the `pathParams`
+will assign `:marriageIndex` in the URL a number based in the current item in
+the `marriages` array.
+
+##### Review page
+To keep Formulate simple, it will not build the review page automatically. It
```
jbalboni, line 340: I think this is going to be hard to swallow if it's not accompanied by a recommendation from the design system group of a simpler pattern for form review pages. Even with the components listed below, building a review page like the current one is going to be difficult and time consuming, particularly for large forms.

platform/engineering/design-docs/2021-05-18-forms-library.md
jbalboni: This isn't helping you, but: a couple years ago we did validate through Sentry errors that not having this validation at the end of the form was putting users in a position where they were stuck at the end.

f1337 (approved):
I thought I reviewed & approved this a while back. Mea culpa folks. It looks good!

Jonny Lee Miller yelling 'Hack the planet' from inside a police car, in the movie Hackers

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,942 @@
...
+- [`api-extractor`](https://api-extractor.com/pages/overview/intro/)
+- [React router](https://reactrouter.com/)
+- [Formik](https://formik.org/)
+
+The following sections outline the various parts of the system.
+
+#### Form page builder
+Formulate will leverage Formik for much of the form page building. It will
+**re-export Formik components** where it makes sense and **provide light
+wrappers for Formik input fields** using design system components.
```
jbalboni, line 136: I would like to see more about how validation is handled for form page components. The current form system implements a particular pattern where validation is run instantly and shown depending on the visited/blurred state of the field. And when you submit the form, all those errors are revealed and the first one is focused on. Formik provides the hooks needed to do all that (I think), but in this doc I'm not sure where that would live (some of it would live in the form field wrappers, but not all of it can go in there). I'd also be interested to know how we should write validation rules, for both regular form fields and also dynamic fields. Should we use Yup? Something else? I know jsonschema is mentioned below, but it's in the context of the final submission.

Making validation "just work" is I think one of the big strengths of the current form system, since it's very easy to mess up and create inaccessible flows. So I think Formulate should make that priority as well.

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,942 @@
...
+  schema, which means data transformation functions are required
+  - The 21-526EZ form is a prime example of this because of its complexity
+- Again, the [learning curve](#steep-learning-curve) gets in the way of rapid
+  progress on more complicated forms
+  - Engineers have to learn the interface for making dynamic changes through
+    hooks provided in the `uiSchema` instead of using standard React conventions
+- Using external services to determine which fields to render is cumbersome
+  - Rather than conditionally rendering the fields, engineers must generate a
+    schema
+- Paring `schema` and `uiSchema` makes sharing reusable patterns awkward
```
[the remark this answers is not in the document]
cvalarida: Blast. Spell checker didn't catch this because `paring` is a real word. :sweat_smile: 

Thanks. :+1:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: This is specifically in the routing section because it prevents routing through the form. [Formik handles validation when the form is submitted](https://formik.org/docs/guides/validation#when-does-validation-run), so for the form page builder, my thought is applications would just use Formik's submission result to determine what it should do.

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Agreed. The review page is the part I'm the least comfortable about for this whole design document. I just don't know how to automate the creation of it without some pretty nasty blows to the developer experience when using the rest of the library. :disappointed: 

The root cause of many of the problems with the current forms library—as I see it—is the centralized configuration for building form elements. That's the main problem I'm trying to solve. But without that centralized configuration, it's really tough to build all the pages _and_ the review mode for each page all at once automatically. So barring that, my next best option I could come up with was to simplify the creation—and customization—of the review page.

If you or anybody else has ideas for how to make that happen, I'd love to hear them. :sweat_smile:

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Ah, great call out. I'll go back over and strengthen that section for handling validation, outlining the specifics for when validation will run and how focus will be handled. :+1:

dat-gcio:
Not sure if this is the right place to ask this, but are there any migration strategies for existing forms? Is the plan to only use this for new forms? If the old form system is deprecated, would we have to migrate when we want new features?

cvalarida:
@dat-gcio Great question. I'm still figuring out what a migration strategy might look like, but the idea is that my team would eventually stop supporting the old forms library since it's a pretty big drain on our already-limited capacity. New forms would use the new library, and we'd figure out how to port the old forms to the new library.

One idea I had was to open up the ability in the current forms library to totally bypass using `schema` and `uiSchema` at the page level and instead let engineers supply a React component for the "normal" page and the view state on the review page. There are some complexities there that I haven't worked through yet, but it seems feasible. If it works out, that would allow teams to adopt the new library incrementally, page-by-page, until all pages are ready. From there, it should be pretty straightforward to migrate wholesale to the new library using `Router` and `SaveInProgress`.

To avoid clutter in this PR, though, let's continue that discussion in Slack. :+1:

platform/engineering/design-docs/2021-05-18-forms-library.md
bkjohnson: I know with Web Components we can have our DOM element and do something like `pageNode.validate()`, provided that the page has a `validate` method. Can we do the same with Formik?

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,1025 @@
...
+
+**Tasks:**
+- Re-export Formik components such as `Formik` and `Form`
+- Build input fields
+  - Text box
+  - Text area
+  - Checkbox
+  - Radio group
+  - Select box
+  - File input
```
bkjohnson, line 850: Is this just referring to the Formik wrappers, or for the Web Components that they will be wrapping?

bkjohnson (commented):
I feel like we should have a more concrete plan for piecing data from all of the separate `<form>` elements into a single object for validation & also submission.

bkjohnson:
Oops, looks like I overlooked this line:

> When the user navigates to a new page, the page state will be deeply merged with the form state in the React Context

platform/engineering/design-docs/2021-05-18-forms-library.md
cvalarida: Just the Formik wrappers. I didn't include the building the Web Components it'll use since that's work adjacent to the forms library itself.

platform/engineering/design-docs/2021-05-18-forms-library.md
```diff
@@ -0,0 +1,1025 @@
+# Forms library
+
+**Author(s):** Christopher Valarida  
+**Last Updated:** May 18, 2021  
+**Status:** **Draft** | In Review | Approved  
+**Approvers:** Brooks Johnson [ ], Michael Fleet [ ]
```
bkjohnson, line 6:
```suggestion
**Approvers:** Brooks Johnson [x], Michael Fleet [ ]  
```

bkjohnson (approved):

cvalarida:
Awesome! Thanks everybody! I'm gonna go ahead and merge this with the understanding that there are still some outstanding questions about the review page. Specifically, I still need to work out validation. I have some ideas, but I think it'll be more efficient to try them out when we have a working library, and I don't expect it'll greatly affect the overall architecture. When we settle on a specific approach, I'll update the design doc to reflect the current state.

As always, if there are any questions after this has been merged, feel free to either leave them on this PR or reach out to the design system team in Slack.

PR #25943, opened 2021-06-11
Component docs formatting + accordion dataLayer

bsmartin-ep:
Some missing `'` characters in some code blocks.

Accordion dataLayer was using older, non-component dataLayer.


joanneesteban (approved):
LGTM! Thanks, Brian!

Issue #25940, opened 2021-06-10
Add preview support for “child” pages like /*health-care/locations/


mpelzsherman:
Currently, the preview server only works on "top-level" pages that have a nodeId, e.g. pittsburgh-health-care.

Some "child" pages such as pittsburgh-health-care/locations are computed from the parent and are not preview-able.

We need to create a general mechanism for previewing such pages, e.g. adding the URL as a query parameter:

`/preview?nodeId=XXX&childPath=locations`

mpelzsherman:
Turns out for these pages, there is still a node in Drupal - we just aren't displaying it in the header for some reason. You can find the node by going to the URL in Drupal and editing it - the node will appear in the Drupal page URL. We should add a ticket to display the missing entityId's in the page header.

For the /*health-care/locations/  page, we also needed to add a graphql query to `GetLatestPageById.graphql.js`. This will be handled in the PR for #25479.

Issue #25793, opened 2021-06-08
Generate headerFooter.json on demand


mpelzsherman:
# Description

When running `yarn preview`, if the file `../vets-website/build/localhost/generated/headerFooter.json` does not exist, the server throws an error and dies. The only soluion is to run `yarn build`, which kind of defeats the purpose of using the preview server.

`headerFooter.json` is only 21K. It can probably be generated on demand quickly as needed.

# Acceptance Criteria

The preview server does not throw an error if `headerFooter.json` doesn't exist, and the header and footers render correctly.

alexandec:
Just catching up on how the preview server setup works. Let me see if I have it right. The process is:

1. Run `yarn build` in the vets-website repo (only need to do this once)
2. Run `yarn build` in the content-build repo (only need to do this once)
3. Run `yarn preview` in the content-build repo
4. Visit a url like `http://localhost:3002/preview?nodeId=15595` in your browser

On step 4, the preview server fetches the latest data for the provided nodeId and runs a metalsmith build just for that node. You can reload or change the nodeId anytime to rerun this single-node build to see an up-to-date preview of that node.

When you say that running `yarn build` defeats the purpose of using the preview server, could you please explain further? Doesn't the preview server allow you to see the latest content for the supplied node, without rerunning `yarn build`?

theodur:
I made a [PR](https://github.com/department-of-veterans-affairs/content-build/pull/283) with an update for this. The changes in my PR create the `headerFooter.json` file in the preview server's Metalsmith pipeline rather than just fetching it from the `generated` directory. Also, fetching `file-manifest.json` from the `generated` directory will fail if you don't run a full content build prior, since it makes the symlink to `vets-website/build/localhost/generated`. I updated the preview script to make the symlink if it doesn't exist so that there's no need to run the content build for that as well. @mpelzsherman whenever you get a chance, feel free to checkout the branch and let me know if everything is working properly.

mpelzsherman:
> Just catching up on how the preview server setup works. Let me see if I have it right. The process is:
> 
> 1. Run `yarn build` in the vets-website repo (only need to do this once)
> 2. Run `yarn build` in the content-build repo (only need to do this once)
> 3. Run `yarn preview` in the content-build repo
> 4. Visit a url like `http://localhost:3002/preview?nodeId=15595` in your browser
> 
[...]

PR #25956, opened 2021-06-11
Design Doc Review for Small Batch Publishing

indytechcook:
*Edit* https://github.com/department-of-veterans-affairs/va.gov-team/blob/nh-small-batch-publishing/platform/engineering/design-docs/2021-06-11-small-batch-publishing.md


platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,350 @@
...
+* [Revision History](#revision-history)
+
+## **Overview**
+
+
+### **Objective**
+
+An editor on the CMS should be able to publish content to va.gov when the "Publish" button is pressed in the CMS. This document outlines an approach utilizing the preview server’s code to render HTML which is directly deployed to S3.
+
+This design doc only covers publishing pages which are represented in Drupal as a node. The full content build still runs hourly.
```
timwright12, line 56:
```suggestion
This design doc only covers publishing pages which are represented in Drupal as a node. The full content-only build still runs 9am-12pm Hourly, 1:45pm, 4pm, 5pm ET M-F.
```

timwright12: What does the content-only deploy contain if all the drupal content goes out immediately? The vagov-content stuff? If that's the case, I feel like we can just cancel the content-only deploys, that repo is barely ever updated

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,350 @@
...
+* [Accelerated Publishing Discovery](https://app.zenhub.com/workspaces/vagov-cms-team-5c0e7b864b5806bc2bfc2087/issues/department-of-veterans-affairs/va.gov-cms/4748)
+* [Single Page Preview Discovery](https://app.zenhub.com/workspaces/vagov-cms-team-5c0e7b864b5806bc2bfc2087/issues/department-of-veterans-affairs/va.gov-cms/5095)
+* [Accelerated Publishing Documentation](https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/platform/cms/accelerated_publishing/)
+* [Accelerated Publhsing Discovery Finding](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/cms/accelerated_publishing/findings_recommendations.md)
+
+### **Background**
+
+The current content-build for va.gov runs once an hour and generates all HTML pages. The current build takes approximatly 15 minutes for page to show up on va.gov. When the amount of content in Drupal reaches 48K nodes, the build will take close to 45 minutes to publish content to va.gov.  Editors are often confused about when the content they published in the CMS will be live on va.gov.
```
timwright12, line 71:
```suggestion
The current content-only build for va.gov runs once an hour and generates all HTML pages. The current build takes approximately 15 minutes for page to show up on va.gov. When the amount of content in Drupal reaches 48K nodes, the build will take close to 1 hour to publish content to va.gov.  Editors are often confused about when the content they published in the CMS will be live on va.gov.
```

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,350 @@
...
+
+Editors coming from Team Sites expect publication of content within 5 minutes.  As editors migrate from Team Sits to Drupal, editors often do not understand why the content they publish does not show up on va.gov in a timely mannter.
```
timwright12, line 73:
```suggestion
Editors coming from Team Sites expect publication of content within 5 minutes.  As editors migrate from Team Sites to Drupal, editors often do not understand why the content they publish does not show up on va.gov in a timely mannter.
```

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,350 @@
...
+
+
+### **High Level Design**
+
+The same code used to generate the HTML for the preview server can be used to generate the HTML for va.gov.  
+
+The following diagram shows a high level design of how a node in Drupal would be published “immediately” on va.gov.
+
+Drupal is in charge of determining when and which pages need to be published based upon the action by an editor.  When Drupal determines a page is to be published, a Publish Batch Record will be sent to Amazon Simple Queue Service (SQS). A Lambda job will pick up the Publish Batch Record to build the HTML. After the HTML has been created, it’s sent to an S3 bucket which serves the html for va.gov.
```
timwright12, line 82:
```suggestion
Drupal is in charge of determining when and which pages need to be published based upon the action by an editor.  When Drupal determines a page is to be published, a Publish Batch Record will be sent to Amazon Simple Queue Service (SQS). A Lambda job will pick up the Publish Batch Record to build the HTML. After the HTML has been created, it’s sent to an [S3 bucket](http://content.www.va.gov.s3-website-us-gov-west-1.amazonaws.com/) which serves the html for va.gov.  
```

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,350 @@
...
+
+## **Specifics**
+
+
+### **Detailed Design**
+
+
+#### Drupal Triggers Content Publishing
+
+When an editor performs an action in Drupal (ie publishing a node, edits a menu item), Drupal will determine which pages need to be published. [The logic will be encapsulated by a Drupal Service.](https://app.zenhub.com/workspaces/vagov-cms-team-5c0e7b864b5806bc2bfc2087/issues/department-of-veterans-affairs/va.gov-cms/5509) The rules around the dependency graph are being tracked on [here](https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/platform/cms/accelerated_publishing/dependency_graph) and are currently a work in progress. Which rules are addressed will be determined by the content type being worked on to support Small Batch Publishing.  Drupal will also have logic to not allow a Small Batch Publishing if a full build is in progress.
```
timwright12, line 100: > Drupal will also have logic to not allow a Small Batch Publishing if a full build is in progress.

I'm curious about this implementation and the user affects. If we ever push that 1 hour build time the assumption is that not content will be published in that time period or maybe queued? and released after?

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,350 @@
...
+
+Drupal will track the queue progress using the [advancedqueue](https://www.drupal.org/project/advancedqueue) and the Public Batch API described below. The advancedqueue module allows for observability and reporting.
+
+
+#### Preview Server Changes
```
timwright12, line 105: Just food for thought that I think engineers use the preview functionality locally as well, so another point of QA

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,350 @@
...
+#### Facility Banner Alerts
+
+The first type of data to be updated using the Small Batch Publishing will be the facility banner alerts and alerts. These types of content currently trigger a deploy from Drupal to the jenkins job. The trigger from Drupal to Jenkins will be replaced with Small Batch Publishing.  By removing the need for Drupal to trigger the Jenkins job, we eliminate logic around Drupal bulk publish events and when an editor is allowed to trigger a Small Batch Publish.
+
+1. Banner Alert or Alert updated in Drupal
+2. Drupal sends Publish Batch Record to SQS
+3. Lambda picks up Publish Batch record
+4. Metalsmith build runs for alerts and generates a JSON file instead of an HTML file.
+5. JSON file is pushed to S3
+6. A React Widget looks for the JSON file on S3 to populate the alert.
```
timwright12, line 126: If this is a react widget on the page do we need metalsmith to run at all? Can we just update an endpoint and hit it live?

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,350 @@
...
+| Plug the content into the templates                       | X              | X                          | X                       |                                                                                                                                                         |
+| Translate the markdown to html                            | X              | X                          |                         | This probably isn’t needed on SBP since the markdown pages will not be processed.                                                                       |
+| Add permalinks and change foo.md to foo/index.html        | X              | X                          |                         |                                                                                                                                                         |
+| Create header and footer                                  | X              |                            |                         |                                                                                                                                                         |
```
timwright12, line 154: There's an open ticket to add this to the preview functionality

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,350 @@
...
+If the current branch is `master`, also get the version that the `master` alias points to. If the master alias doesn't match the version of the checksum alias:
+
+*   [Update the master alias](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/lambda/update-alias.html) to the version currently aliased with the checksum.
+
+A second Github Action will be created that is [manually triggered](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). The action will [Update and publish the function code](https://docs.aws.amazon.com/lambda/latest/dg/API_UpdateFunctionCode.html) the production lambda function.
+
+
+#### Assets
+
+Assets (images, pdfs) which are generated in Drupal will be uploaded to S3.  The content build will be updated to use the new S3 location as the `src` for the assets.  The details of this implementation will be in a different design doc and implemented outside of Small Batch Publishing.
```
timwright12, line 245: > The content build will be updated to use the new S3 location as the `src` for the assets.

Ideally we'd hope that the `src` are already set to S3 before the content build starts so we don't have to parse it. Is that possible?

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,350 @@
...
+*   GitHub Actions: `.github/workflows/single-page-build.yml`
+
+The Lambda functions will be provisioned in the devops repo.  The POC code will be updated to support dev/stage/prod.
+
+The Drupal code will live in the `va.gov-cms` repository.
+
+
+### **Testing Plan**
+
+SBP and full build need to be tested and the output validated per content type.  The preview server currently has a `/diff` endpoint which generates a visual difference between the preview server HTML and production.  Example: [http://preview-dev.vfs.va.gov/diff?nodeId=2578](http://preview-dev.vfs.va.gov/diff?nodeId=2578). The preview will manually check for a subset of each content type before we turn on instant publishing for the content type.
```
timwright12, line 264: > SBP and full build need to be tested 
Also local workflows

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,350 @@
...
+GitHub Actions will output logs for the build and versioning of the function. Log output from the function (from calling console methods) will be stored in CloudWatch.
+
+The lambda function will also communicate back to Druapl using the [Publish Batch API](#publish-batch-api) where Drupal will generate an Audit Trail of what was build by who and when.
+
+### **Debugging**
+
+*   One can start diagnosing any issues using the logs mentioned above.
+*   The Webpack build of the function can be run locally to debug issues with the build.
+*   The code should be abstracted in a way where the code can be executed outside of the lambda function.  For example, the preview server itself should be running the same code.
+*   Debugging the HTML produced and using the `/diff` endpoint which shows the HTML from the code compared to production.
```
timwright12, line 277: We use timestamps on assets as a cache buster, so you likely won't be able to get a true diff comparison here: https://github.com/department-of-veterans-affairs/content-build/blob/master/src/site/stages/build/plugins/modify-dom/process-entry-names.js

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: We are planning on rolling out SBP by content type over a period of months.  After that, there are the pages from the markdown content repo and finally, pages (like listing and location pages) that use Drupal content to build a page but are not represented in Drupal.  Though https://github.com/department-of-veterans-affairs/va.gov-team/issues/25940 from @melissajschaff  might bring a solution.  Either a dependency graph in the build or moving the pages to Drupal would help with this use case.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: This is a great question and something we have been discussing without a great solution.  

 Initially, the lambda will requeue the job if there is a build in-progress.  It should also set the Publish Batch Records `state` to `ON HOLD` or something.  Then when the lambda job runs, it can look at the revision that is live and compare it to the revision it's publishing to not publish an older revision.

Longer-term, hopefully, we can find a way to remove content types from the full build and only do them in the SBP.  Actually, thinking about this now, it's probably easier to do short-term.  To just cut over from full build to SBP.  @timwright12 thoughts about this?

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: The idea would that the react widget would load the data directly from a JSON file on S3 so no metasmith build required.  The JSON file would be updated as part of the SBP pipeline.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: Would this requre the https://github.com/department-of-veterans-affairs/vagov-content repo?

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: Yeah.  Initially, we were thinking of adding a new field to the Drupal file entities which contained the full S3 path.  Then we update the build/templates to use the new field.  This would remove any processing in the build.  We discussed this idea with @alexandec earlier this week.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: Yeah, those and the ids added by the analytics middleware.  When we manually validate, it's easy to see them on the diff.  We don't have an automated method in mind yet.

Thanks for the heads up.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
timwright12: I think we should try and push for more unification of the content so we actually turn off the "hourly" content release. Seems like there would be a ton of times throughout the day were a build were happening that would block instant publishing making it a little more annoying rather than helpful

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
timwright12: For content, yeah, I'd love to just move it into drupal so metalsmith doesn't really even need to worry and can just focus on templates

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
timwright12: Gotcha, ok, #3 confused me there

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
timwright12: TBD on that I think @theodur is picking up that ticket next week

indytechcook: Is this the issue? https://github.com/department-of-veterans-affairs/va.gov-team/issues/25793

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: I added more detail here and linked to the epic.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: I separated out the creation of the react widget and linked to the epic with more detail.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: I updated the plan to cutover content types from full build to SPB.  This will make reduce lots of complexity.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: I updated the plan to cutover content types from full build to SPB. This will make reduce lots of complexity.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: I added notes about validating local build.

mpelzsherman:
How (if at all) would this impact local development?

Under this model, will the preview server still be used to view draft content?

indytechcook:
> How (if at all) would this impact local development?
> 
> Under this model, will the preview server still be used to view draft content?

Ideally, there would be no change to local development.  The preview server will still run the same.  I have a note in the testing section of the document to make sure the preview server still functions as expected.  There will be a query param added to specify which revision of the node to load which will add flexibility to load the draft or published version.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,353 @@
...
+| Check for CMS URLs                                        | X              |                            |                         |                                                                                                                                                         |
+| Parse a virtual DOM from every .html file                 | X              | X                          | X                       |                                                                                                                                                         |
+| Ignore assets for build                                   | X              |                            |                         |                                                                                                                                                         |
+| Rewrite AWS URLs                                          |                | X                          | X                       |                                                                                                                                                         |
+
+
+
+#### Publish Batch API
```
indytechcook, line 175: The implementation here might shift a bit to a more generic Event Bus using SNS and SQS.  I have experience with RabbitMQ and it looks like the SNS/SQS AWS pattern most closely matches RabbitMQ functionality: https://stackoverflow.com/questions/46880229/migrate-from-amqp-to-amazon-sns-sqs-need-to-understand-concepts

indytechcook: I've updated the architecture diagram and add a note to this section regarding SNS and SQS patter.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
timwright12: @indytechcook yeah that's the one

mpelzsherman:
If we’re going to roll our own "instant publishing" solution (which appears likely), I hope it will be a stepping stone to a move to a well-supported, open source/industry-standard like Gatsby or Next.js.

A key component of any successful migration will be solid automated tests, so we'll be able to tell if we're breaking things.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
indytechcook: Thanks, I linked it to the doc!

indytechcook:
@mpelzsherman please help me understand how this is "rolling our own" and what we would not have to do using Gatsby or Next.js.

mpelzsherman:
> @mpelzsherman please help me understand how this is "rolling our own" and what we would not have to do using Gatsby or Next.js.

We're already building on top of a custom solution (DrupalGraphQL + Metalsmith + Liquid + a ton of JS code).

My admittedly naive understanding is both Gatsby and NextJS already have support for most of what we do in our current solution, as well as incremental publishing: [Gatsby](https://www.gatsbyjs.com/blog/2020-04-22-announcing-incremental-builds/) [NextJS](https://www.smashingmagazine.com/2021/04/incremental-static-regeneration-nextjs/).

Based on what I'm seeing here, we're not really moving away from our custom solution. Instead we're adding more custom code to do things like computing rebuild dependencies. That's what I mean by "rolling our own".

This may be a reasonable approach given the lift involved in a complete migration to a new platform. OTOH what you've described here looks like a heavy lift as well, so maybe our efforts would be better spent moving to a new platform with the support of the open source community?

indytechcook:
Thanks @mpelzsherman 

Most of the work in this design doc is about setting up the infrastructure required to trigger small batches of publishing.  This work would be required no matter what the technology used to build the HTML.  

Gatsby and NextJS are frameworks that both would require a large amount of custom Javascript code to implement.  The change in frameworks also means it could require all VFS teams to rebuild all of their applications. 

The discovery work that was done in the Accelerated Publishing Initiate (thank you for participating!) showed the best direction is to be making Drupal public.  The industry standard for publishing content from a CMS is to have the CMS publish the HTML.  The discovery work also lead us to the conclusion that the effort and risk of moving to Drupal publishing the HTML is not something achievable in a 6 month time period.  

The approach in this design document was selected due to the amount of value we can achieve for the effort.  Don't think of this as an end goal, just another step along the journey.

mpelzsherman:
A few more questions:

1. Who will own the tasks of implementing and maintaining these changes?

1. This doc only covers publishing pages which are represented in Drupal as a node.

    - How will “derived” pages such as https://www.va.gov/pittsburgh-health-care/locations be published then? 
    - BTW, my team (facilities) has a [story](https://github.com/department-of-veterans-affairs/va.gov-team/issues/25940) in our backlog to support previewing “derived” pages.

1. Under Preview Server Changes, you wrote:

    > 	Code will be updated to use the json processing from here: src/site/stages/build/drupal/*.js

    Can you elaborate on this? Most of that code isn’t appropriate for the preview server because it operates on all of the files in the site. The Facilities team is currently in the process of removing as much of that code as possible, so we can use the preview server to render ALL of the pages we support. Local development using the full build is so slow that it’s almost impossible at this point, and getting worse every day.

1. In your Metalsmith Plugins chart, you have a “check” for Preview Server and SBP under “Parse a virtual DOM from every .html file”. I don’t think that’s accurate for the preview server - it only renders one page at a time.

attn @mmiddaugh

mpelzsherman:
Also:

>  The change in frameworks also means (moving to Gatsby or NextJS) could require all VFS teams to rebuild all of their applications

I'm pretty sure that's not the case. Most of the VA.gov applications are written in React, which works very seamlessly with Gatsby or NextJS. The static pages would need to be migrated from liquid templates to React/Web components, but that would be a welcome change - nobody _likes_ working with liquid! 😄

indytechcook:
One thing I realized I didn't make clear is the longer-term vision of where we see content publishing going.  I added a section to the document to expand on this area.  In short, Small Batch Publishing is a realistic solution we can implementing in Q3.  We can do discovery on making Drupal public in Q4 with a goal of migrating to a public Drupal API-based solution in 2022.  

The solution in this design doc is focused on publishing time and build scalability instead of developer experience.  Think of this work as a code refactor rather than new code.  Most of the code already exists, it will just moved around so it be used to publish smaller batches of content.  This solution allows editors to publish content within a minute and address the 1 hour build time we see coming at 48K nodes.

The discovery effort did look at Next.js/Gasby based solution.  We made the decision based upon the discovery, that several teams were a part, that the effort to refactor all templates would be better spent in migrating to a public Drupal based system (HTML publishing or API) rather than a different static site generation framework.  The actual solution will most likely be a mix between Drupal publishing HTML and javascript framework.  Something like https://druxtjs.org/.  Discovery will occur again before a solution is established and I hope you can be a part of the discussion.

More information about what we aren't doing yet is found in supporting documentation and summarized here: https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/platform/cms/accelerated_publishing/findings_recommendations.md 

> 1 Who will own the tasks of implementing and maintaining these changes?

CMS team will implement with support from FE tools.  

While the call to make the decision about code ownership is not my call, I see the CMS team owning the content publishing pipeline, FE tools owning the template/code build pipeline, and product teams owning their templates, apps.  But again, I can't make that call.

I assume what you actually want to know is how it will affect you and other teams.  The ask from others will be in advice, code review, and testing.  This shouldn't be more than the normal interactions we have already.  Many of the advice-based questions can be during the "Content Release Engineering Sync" meeting we already have.  Any formal involvement required will be made in tickets that other teams can prioritize.  I can't say for sure we won't ask for more help but at this time, we want to keep the impact on your day-to-day work minimal. 

> 2 This doc only covers publishing pages which are represented in Drupal as a node.

>How will “derived” pages such as https://www.va.gov/pittsburgh-health-care/locations be published then?
>BTW, my team (facilities) has a story in our backlog to support previewing “derived” pages.

I looked forward to seeing what you come up with!

> 3 Under Preview Server Changes, you wrote:
> 
> Code will be updated to use the json processing from here: src/site/stages/build/drupal/*.js
> 
> Can you elaborate on this? Most of that code isn’t appropriate for the preview server because it operates on all of the files in the site. The Facilities team is currently in the process of removing as much of that code as possible, so we can use the preview server to render ALL of the pages we support. Local development using the full build is so slow that it’s almost impossible at this point, and getting worse every day. 

How is this code not appropriate outside of the full build?  It just runs against the small subset of pages instead of all the pages.

The local build is slow because it runs against all pages.  Small Batch Publishing is more about refactoring the current code than writing more custom code with the primary goal being the preview server, SBP lambda job, and the full build all use the same code.

> 4. In your Metalsmith Plugins chart, you have a “check” for Preview Server and SBP under “Parse a virtual DOM from every .html file”. I don’t think that’s accurate for the preview server - it only renders one page at a time.

This step is already occurring in the preview server and will need to continue to occur for the HTML to output correctly.  The "Small Batch Publishing" will process groups of pages to publish based upon dependencies.  For example, if an event node is published, the event listing and event page will be processed and pushed to S3 at the same time.   The step will only process the change HTML files instead of all of them.

I plan on hosting office hours where more detail can be discussed and more questions can be answered.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,361 @@
...
+
+## **Specifics**
+
+
+### **Detailed Design**
+
+
+#### Drupal Triggers Content Publishing
+
+When an editor performs an action in Drupal (ie publishing a node, edits a menu item), Drupal will determine which pages need to be published. [The logic will be encapsulated by a Drupal Service.](https://app.zenhub.com/workspaces/vagov-cms-team-5c0e7b864b5806bc2bfc2087/issues/department-of-veterans-affairs/va.gov-cms/5509) The rules around the dependency graph are being tracked on [here](https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/platform/cms/accelerated_publishing/dependency_graph) and are currently a work in progress. Which rules are addressed will be determined by the content type being worked on to support Small Batch Publishing.  As a content type is add to Small Batch Publishing, it will be removed from the hourly content build.  Having support in only one build at a time will elimnte the risk of the full content build overriding the html from the Small Batch Publishing.
```
timwright12, line 102:
```suggestion
When an editor performs an action in Drupal (ie publishing a node, edits a menu item), Drupal will determine which pages need to be published. [The logic will be encapsulated by a Drupal Service.](https://app.zenhub.com/workspaces/vagov-cms-team-5c0e7b864b5806bc2bfc2087/issues/department-of-veterans-affairs/va.gov-cms/5509) The rules around the dependency graph are being tracked on [here](https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/platform/cms/accelerated_publishing/dependency_graph) and are currently a work in progress. Which rules are addressed will be determined by the content type being worked on to support Small Batch Publishing.  As a content type is add to Small Batch Publishing, it will be removed from the hourly content build.  Having support in only one build at a time will eliminate the risk of the full content build overriding the html from the Small Batch Publishing.
```

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,361 @@
...
+Editable link: [https://lucid.app/lucidchart/a8445048-94ce-4473-aafa-e8ed6cbfbad5/edit?page=0_0#](https://lucid.app/lucidchart/a8445048-94ce-4473-aafa-e8ed6cbfbad5/edit?page=0_0#)
+
+
+## **Specifics**
+
+
+### **Detailed Design**
+
+
+#### Drupal Triggers Content Publishing
```
timwright12, line 100: are/can we address the content that lives in [vagov-content](https://github.com/department-of-veterans-affairs/vagov-content)?

indytechcook: The content should be migrated to Drupal. I realize this is a loaded statement.

I had a brief blub in the `Mid Term Vision`

> This would require all pages in va.gov to have a Drupal representation, including the current markdown content

mpelzsherman:
> This solution allows editors to publish content within a minute 

I'd be careful about making promises like that until we have more data. 😄

mpelzsherman:
I'm still confused about the role of the preview server, and who is responsible for maintaining the preview server code.

I thought we were planning to leverage the preview server to render the "small batches" (which would be as small as a single page). Is that no longer the case?

indytechcook:
> I thought we were planning to leverage the preview server to render the "small batches" (which would be as small as a single page). Is that no longer the case?

That was the POC.  This document outlines the direction of making the code reusable between the content-build, preview server and lambda function.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
timwright12: excellent, 100% agree. thanks!

timwright12 (approved):

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,355 @@
...
+The first type of data to be updated using the Small Batch Publishing will be the facility banner alerts and alerts. These types of content currently trigger a deploy from Drupal to the jenkins job. The trigger from Drupal to Jenkins will be replaced with Small Batch Publishing.  By removing the need for Drupal to trigger the Jenkins job, we eliminate logic around Drupal bulk publish events and when an editor is allowed to trigger a Small Batch Publish.
+
+1. Banner Alert updated in Drupal
+2. Drupal sends Publish Batch Record to SQS
+3. Lambda picks up Publish Batch record
+4. Metalsmith build runs for alerts and generates a JSON file instead of an HTML file.
+5. JSON file is pushed to S3
+
+In a sepearte effort, the code which generates the Facility Banner Alert will be moved from the Metasmith Template to a React Widget.  The React widget will load the JSON file from S3 which was generated by Drupal in the process described above.
+
```
drorva, line 131: > Metasmith Template to a React Widget

Liquid, not metalsmith.

drorva (commented):
Overall, this is all  quite obvious in a good way.
There are two main areas in here that are somewhat buried and should be highlighted.
* Figuring out the rules around the dependency graph is a critical piece of this whole effort
* We generate a JSON file instead of an HTML and use a react widget.
Specifically the second one needs further discussion. Do all the liquid templates need to be rewritten? Are we going to use this across the board?

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,355 @@
...
+
+### **High Level Design**
+
+The same code used to generate the HTML for the preview server can be used to generate the HTML for va.gov.  
+
+The following diagram shows a high level design of how a node in Drupal would be published “immediately” on va.gov.
+
+Drupal is in charge of determining when and which pages need to be published based upon the action by an editor.  When Drupal determines a page is to be published, a Publish Batch Record will be sent to Amazon Simple Queue Service (SQS). A Lambda job will pick up the Publish Batch Record to build the HTML. After the HTML has been created, it’s sent to an [S3 bucket](http://content.www.va.gov.s3-website-us-gov-west-1.amazonaws.com/) which serves the html for va.gov.  
+
```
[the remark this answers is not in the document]
indytechcook: The Small Batch Publishing jobs are triggered from content changes in Drupal and not from code changes in a repo.  The use case for GH Actions seems to focus more on responding to events in GH.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,355 @@
...
+*   Code will be updated to use the json processing from here: `src/site/stages/build/drupal/*.js`
+*   `/diff` endpoint code will be refactored to use the same code as the lambda function.
+*   GraphQL Query building will be refactored to have Full and Small Batch use the same queries.
+*   The Get Drupal Content plugin should be broken apart to make the layers reusable with the different services.
+*   When changes are made, validate local develpment workflow using the Preview Server with the Product and QA teams.
+
+
+#### Facility Banner Alerts
+
+The first type of data to be updated using the Small Batch Publishing will be the facility banner alerts and alerts. These types of content currently trigger a deploy from Drupal to the jenkins job. The trigger from Drupal to Jenkins will be replaced with Small Batch Publishing.  By removing the need for Drupal to trigger the Jenkins job, we eliminate logic around Drupal bulk publish events and when an editor is allowed to trigger a Small Batch Publish.
```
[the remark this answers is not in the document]
indytechcook: Thanks, they are two different things but we really are focused on facility banner alerts.  I'll update.

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,361 @@
...
+
+1. Banner Alert updated in Drupal
+2. Drupal sends Publish Batch Record to SQS
+3. Lambda picks up Publish Batch record
+4. Metalsmith build runs for alerts and generates a JSON file instead of an HTML file.
+5. JSON file is pushed to S3
+
+In a sepearte effort, the code which generates the Facility Banner Alert will be moved from the Metasmith Template to a React Widget.  The React widget will load the JSON file from S3 which was generated by Drupal in the process described above.
```
indytechcook, line 130:
```suggestion
In a separate effort, the code which generates the Facility Banner Alert will be moved from the Liquid Template to a React Widget.  The React widget will load the JSON file from S3 which was generated by Drupal in the process described above.
```

platform/engineering/design-docs/2021-06-11-small-batch-publishing.md
```diff
@@ -0,0 +1,361 @@
...
+*   Code will be updated to use the json processing from here: `src/site/stages/build/drupal/*.js`
+*   `/diff` endpoint code will be refactored to use the same code as the lambda function.
+*   GraphQL Query building will be refactored to have Full and Small Batch use the same queries.
+*   The Get Drupal Content plugin should be broken apart to make the layers reusable with the different services.
+*   When changes are made, validate local develpment workflow using the Preview Server with the Product and QA teams.
+
+
+#### Facility Banner Alerts
+
+The first type of data to be updated using the Small Batch Publishing will be the facility banner alerts and alerts. These types of content currently trigger a deploy from Drupal to the jenkins job. The trigger from Drupal to Jenkins will be replaced with Small Batch Publishing.  By removing the need for Drupal to trigger the Jenkins job, we eliminate logic around Drupal bulk publish events and when an editor is allowed to trigger a Small Batch Publish.
```
indytechcook, line 122:
```suggestion
The first type of data to be updated using the Small Batch Publishing will be the facility banner alerts. These types of content currently trigger a deploy from Drupal to the jenkins job. The trigger from Drupal to Jenkins will be replaced with Small Batch Publishing.  By removing the need for Drupal to trigger the Jenkins job, we eliminate logic around Drupal bulk publish events and when an editor is allowed to trigger a Small Batch Publish.
```

indytechcook (commented):
> Figuring out the rules around the dependency graph is a critical piece of this whole effort

The dependency graph work is complicated and we need to be careful not to introduce more technical debt with duplicated logic in Drupal and javascript.  I've been talking with @mchelen about ways to reduce the risk.  One such way is to implement support for one type of content/template at a time.  As we implement the content type, we audit the code to remove implicit dependencies and create explicit ones in Drupal via entity references.  We can also remove some of the code which munges data after the GraphQL request and before metalsmith and liquid generate the HTML.

> We generate a JSON file instead of an HTML and use a react widget.
> Specifically the second one needs further discussion. Do all the liquid templates need to be rewritten? Are we going to use this across the board?

The actually build for the HTML of the pages will not change, it's still generated by the current Liquid templates.

The use case described in the Facility Banner Alerts is only to replace elements outside of the scope of the node/page.  For example, this pattern can also be used for menus, sidebars, and listing pages.  

Think of the JSON as a static representation of an API call that is stored on S3.  When the content updates in Drupal, the JSON file will be updated also via the same SQS/Lambda job.

We are planning on formatting the JSON to be the same as the Drupal JSON API so that in the future when we have a public Drupal API, the React widgets can be pointed on the https address instead of the JSON file on S3.

There currently is some logic in metalsmith which builds a JSON file and adds it to the build output and is consumed by a React widget.  This pattern is pretty much the same with the only difference being the file is generated by SBP instead of the full build.

indytechcook:
@mpelzsherman I added a section which adds more detail around the `metasmith-drupal` refactor.  https://github.com/department-of-veterans-affairs/va.gov-team/blob/34d977885efaff7dca6dc06c5aaeaaffe11ca967/platform/engineering/design-docs/2021-06-11-small-batch-publishing.md#refactoring-of-metalsmith-drupal-plugin

I dug into the code more and the derived page such as the facility location pages should be created without any effort by your team since all of the data needed to create it is contained within the single page logic.  This, of course, will be audit deeper on implementation and any code which acts on or filter the entire `pages` will be moved to Drupal.  One such example is the `addGetUpdatesFields` function.

alexandec:
@indytechcook thanks for this writeup! Some questions:

1. Will any work from FE Tools be required to implement small-batch publishing? I don't think so based on my read through thus far, but wanted to be sure.
2. Will it be possible to share dependency graph code between small-batch publishing and template-triggered publishing? I guess small-batch publishing is asking "Based on the content change in a single node, which set of nodes needs to be rebuilt?" and template-triggered publishing is asking "Based on the change in a single template, which set of nodes needs to be rebuilt?" Not exactly the same, but it seems like there are some connections 🤔 
3. You mention that "As a content type is add to Small Batch Publishing, it will be removed from the hourly content build." Will there still be a method to perform a full build of the entire site? Or will the size of the "full" build gradually be reduced until there's nothing left in the full build and small-batch publishing handles everything?

indytechcook:
Hi @alexandec thanks for the questions!

> Will any work from FE Tools be required to implement small-batch publishing? I don't think so based on my read through thus far, but wanted to be sure.

While I don't expect to directly assign tasks to the FE tools team, continued feedback and support in the form of Pull Requests and questions will occur.  Most of the questions will most likely be contained to our current sync call.  If it is determined we need more support, we will work with you and @meganhkelley.

> Will it be possible to share dependency graph code between small-batch publishing and template-triggered publishing? I guess small-batch publishing is asking "Based on the content change in a single node, which set of nodes needs to be rebuilt?" and template-triggered publishing is asking "Based on the change in a single template, which set of nodes needs to be rebuilt?" Not exactly the same, but it seems like there are some connections 

For some of them I think so.  The templates which are directly tied to a content type can be determined.   I can see us having an endpoint which says "Rebuild all content from content type X".  This won't work for the shared templates but you should be able to map those to the content type based templates based upon usage.  The filters though are a different question.  

> You mention that "As a content type is add to Small Batch Publishing, it will be removed from the hourly content build." Will there still be a method to perform a full build of the entire site? Or will the size of the "full" build gradually be reduced until there's nothing left in the full build and small-batch publishing handles everything?

The npm build will always have the ability to run a full build.  We will probably add a command line argument which takes a list of content types to skip.  This would just effect the initial queries.  Eventually, when SBP is proven out and supports all content from Drupal, the hourly build will no longer exist.

mchelen-gov:
For the "determine any dependencies" part, I'm ok if that is done through updates to content model to eliminate implicit dependencies and allow GQL queries to properly return all necessary data. Addressing this for each/any of the  [various examples](https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/platform/cms/accelerated_publishing/dependency_graph) is a shared responsibility between CMS Core, CMS Product support, and the VFS team that owns a given product such as PW or Facilities.

Is this functionally different from allowing a product to be supported by Preview Server?

PR #26918, opened 2021-07-02
Update README.md

allison0034:
Added the support table of trevor/noah teams


shiragoodman (changes requested):
Looks good! but I think we dropped, "Contact Center" from Trevor's list. Can you update here and in the Change Log?

PR #28239, opened 2021-08-03
Removes KPI dashboard pause

joanneesteban:
Removes tilda for KPI dashboard pause


shiragoodman (approved):
I will also apply these changes to the Getting Started page in Platform Website.

PR #28254, opened 2021-08-03
Update analytics-request-google-analytics-domo-access.md

laurensimpson14:
Re-ordered it to have the check boxes come first and then the email instructions. Do we actually use the info if they list what team they are on? If not, then we can delete step #3 to make it more concise.


joanneesteban (approved):
LGTM! Thanks for making these edits.

joanneesteban:
We can get rid of Step 3. We were using that to give Domo Group access, but that should be redundant with @chloedotbrown 's tracking.

PR #28788, opened 2021-08-16
add details of preliminary test selection data

pjhill:
Adding some preliminary data to the ADR based on a couple days' worth of data as to the potential savings involved in test selection.

### DIFF
--- a/docs/adr/0010-implement-cypress-test-selection.md
+++ b/docs/adr/0010-implement-cypress-test-selection.md
@@ -38,6 +38,7 @@ We expect:
 ### Preliminary Data
 We are collecting data about `master` builds and `branch` builds on our [dashboard](https://va-gov.domo.com/page/1232474697) including the below key takeaways:
 - About 80% of builds handled in GitHub Actions are `branch` builds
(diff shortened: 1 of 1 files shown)


holdenhinkle (approved):
This looks great!

One other thing you could mention if you want is the percent difference in the number of tests that are being selected on `branch` builds vs. the number that run on `master` builds. For example, here's a little blurb that I wrote to MPS when I was telling him about this feature:

> On Friday, tests ran 60% faster on push (against the branch build), **56% fewer tests ran**, and 61% few containers were need to run the tests.

pjhill:
> This looks great!
> 
> One other thing you could mention if you want is the percent difference in the number of tests that are being selected on `branch` builds vs. the number that run on `master` builds. For example, here's a little blurb that I wrote to MPS when I was telling him about this feature:
> 
> > On Friday, tests ran 60% faster on push (against the branch build), **56% fewer tests ran**, and 61% few containers were need to run the tests.

Good call -- quintessential to test selection is the subset of tests that were selected.

PR #29386, opened 2021-08-26
VA Event Stream -> Lighthouse: Open API spec and service description 

kreek:
The VA Mobile App is working with the Lighthouse team to expose an endpoint for ingesting events from VA systems outside of the mobile api. As part of the onboarding process Lighthouse would like an OpenAPI spec (engineering/openapi_spec.yaml) and service description (README.md and engineering/architecture.md).


products/va-event-stream/engineering/openapi_spec.yaml
```diff
@@ -0,0 +1,163 @@
...
+        payload:
+          type: object
+          description: A JSON object of data that is unique to this event type.
+          example:
+            messageId: 80337bc5-7215-4b22-a20d-84330631d8b5
+            title: Your prescription is ready
+        sourceSystem:
+          type: string
+          description: The VA system that is producing the event. Like the eventType field this should be constant and use the same naming convention.
+          example: MHV Secure Messaging
```
jjulian, line 93: example needs to be updated to uppercase style

jjulian:
```suggestion
          description: The VA system that is producing the event. Like the eventType field this should be constant and use the same naming convention.
          example: MHV_SECURE_MESSAGING
```

jjulian (approved):
This looks great. I appreciate the detailed background, and the planning for future needs, lik storing events for a week. Excited to see this taking shape!

PR #29832, opened 2021-09-10
Lighthouse openapi_spec.yaml feedback revisions

kreek:
Updates to openapi_spec.yaml based on the feedback from the Lighthouse team.

- Fixes some swagger parsing errors for multi-line text. 
- Changes the success response to 202 from 201.
- Updates 401 and 403 response descriptions.
- Adds a 429 error response for when the rate limit is exceeded.
- `Bearer` security definition defined in components.

### DIFF
--- a/products/va-event-stream/engineering/openapi_spec.yaml
+++ b/products/va-event-stream/engineering/openapi_spec.yaml
@@ -44,7 +44,7 @@ paths:
                     icn: 1008709396V637156
       responses:
         202:
-          description: Accepted. The event has been written to the stream but is not immediately available for retrieval.
+          description: Accepted. The event has been written to the stream but may not be immediately available for retrieval.
           content:
             application/json:
               schema:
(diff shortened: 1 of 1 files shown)


products/va-event-stream/engineering/openapi_spec.yaml
```diff
@@ -41,33 +43,53 @@ paths:
                   user:
                     icn: 1008709396V637156
       responses:
-        201:
-          description: Created
+        202:
+          description: Accepted. The event has been written to the stream but may not be immediately available for retrieval.
```
kreek, line 47: The Lighthouse team had commented:
> Should this return a “202 Accepted” since the event may or may not get processed? Returning a 201 is okay, but in the future if webhooks or a return is given for an actual delayed asynchronous response, providing a  “202 Accepted” now will allow the flexibility for that and not require a new version of the API to do that.

I agree with this, for a slightly different reason, while events are immediately written to a stream and return a stream id, it is not their final storage location, and they can not be retrieved in a RESTful manner as soon as they are written. Once a consumer has written an event to Postgres then they can be queried and retrieved. But an event would not be guaranteed to be there should a future hypothetical GET request follow the POST endpoint we're describing.

Issue #27159, opened 2021-07-09
Access requests - automate SOCKS PRs 


jhouse-solvd:
## Description
Creating, approving, and merging PRs to grant SOCKS access is a tedious/manual process. How might we improve this process through automation and speed up our ability to give users the support they need? 

## Background/context/resources
_Any additional context for the reader to know, if applicable_ 

## Technical notes
_Notes around work that is happening, if applicable_

---
## Tasks
- [ ] _What things need to happen?_

## Definition of Done
- [ ] _What is the expected outcome of this issue?_

---
### Reminders
- [ ] Please attach your team label and any other appropriate label(s) (operations, devops, and needs-grooming will automatically be applied as part of the template)
- [ ] Please connect to an epic

jhouse-solvd:
@omgitsbillryan - as time allows, could you write a quick summary of how this might work using GitHub actions? then, we can estimate the level of effort (LOE) and determine how to prioritize.

omgitsbillryan:
I added a Github Workflow that validates the public key for a SOCKS PR: https://github.com/department-of-veterans-affairs/va.gov-team/issues/29277 - not _exactly_ related but a close tangent.

omgitsbillryan:
I'm intending on adding a workflow to automate SOCKS only. The initial plan is

1. Trigger workflow on GH issue creation w/ label `access-request`
2. Exit early if _not_ requesting SOCKS access.
3. Validations:
  - GH user who created ticket is part of the `department-of-veterans-affairs` GH org
  - ssh key does not already exist w/ said email address
  - ssh key is valid (optional since this already happens in PR)
4. Create a GH issue comment summarizing:
- which validations passed & failed
- gathered ssh filename & key contents fetched from GH issue OP
- comment w/ instructions on how to automatically open a PR
5. Automatically create PR

I'm still debating step 5... we may not way to automatically generate the PR. Instead, we could have a 2nd workflow triggered by a slash-command comment on the GH issue, e.g. `/socks-bot create-pr foo@bar.com` that handles the PR creation. Step 5 of the initial workflow would simply comment on the GH issue w/ the validations that occurred + a comment on how to use the aforementioned slash-command comment.

omgitsbillryan:
Tweaking the design a bit after some considerations:

- "Is user in `department-of-veterans-affairs` GH org" & "ssh key does not already exist w/ said email" validations will occur in PR instead
[...]

PR #29971, opened 2021-09-15
Convert Access Request Template to Github Form

omgitsbillryan:
Convert our Access Request GH Issue template to be an issue form instead.

This connects with our effort to automate SOCKS access requests:
- https://github.com/department-of-veterans-affairs/va.gov-team/issues/27159

Using the form instead of markdown will more easily allow us to automatically parse user responses out. W/o using forms, we must use a nasty regex in order to tease out the user information.

### More Context
We were recently approved into the GH Forms beta https://dsva.slack.com/archives/CU1E4CX9U/p1631637698019500


mchelen-gov:
lgtm if you want to merge and give it a try?

PR #33810, opened 2021-12-08
Create draft-veteran-research-plan.md

thehastingsp:
Draft of Tele-Urgent Care Veteran Research Plan


thehastingsp (commented):
looks good

goldenmeanie (commented):
> We hypothesize that Veterans who utilize Tele-Urgent Care will reduce their need to visit the Emergency Room, and will increase the quality and efficiency of care they receive at the VA.

This looks like a longer term hypothesis that won't necessarily be directly tested by this research.

In the recruitment area. Are you looking for a range of ages or only 65+?
Same question for rural. Mix of rural and urban? Do you mean rural and urban or is that a proxy for internet speed?

PR #35362, opened 2022-01-13
Create 2022-01-13-switch-to-vaos-v2-appointments.md

kpethtel:
This is a proposal for using the VAOS V2 appointments endpoint in the VA mobile app.  

Ticket: https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2227

### DIFF
--- a/platform/engineering/request-for-comment/2022-01-13-switch-to-vaos-v2-appointments.md
+++ b/platform/engineering/request-for-comment/2022-01-13-switch-to-vaos-v2-appointments.md
@@ -1,0 +1,137 @@
+# RFC: Switch VA Mobile app to VAOS V2 appointments
+
+- Date: 2022-01-13
+- Related Issue: https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2227
+
+## Background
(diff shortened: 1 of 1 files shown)


platform/engineering/request-for-comment/2022-01-13-switch-to-vaos-v2-appointments.md
```diff
@@ -0,0 +1,137 @@
...
+
+This effort will require:
+- small changes to the proxy
+- rewriting the VA appointments adapter
+- removing or rewriting the CC appointments adapter
+- changes to the mobile appointments service to use the new url
+- possibly adding a new service for fetching clinic data
+- substantial effort to ensure that the data normalization process is correct and matches the current data sent to the mobile app. See the mapping sections below for an initial attempt to map incoming data to the existing mobile appointment payload.
+
+### VA Appointments Mapping:
```
kreek, line 28: Not sure why but the mapping sections are rendering as one big blob of text with no line breaks. The content here is good though, nice to have this level of detail about each field 👍

kreek (approved):
I think this LGTM 🚀 - the conclusion is not the one we were hoping for but it's good to know that the v2 endpoints aren't fully baked yet. Let's merge this and save it as guide on what to do once the v2 endpoints are ready.

platform/engineering/request-for-comment/2022-01-13-switch-to-vaos-v2-appointments.md
kpethtel: Huh. Strange. I don't see it that way. I'm using firefox. Not sure why it would matter though.

PR #35452, opened 2022-01-14
Create 2022-01-14-mobile-schedule-appointments-discovery.md

kpethtel:
https://app.zenhub.com/workspaces/va-mobile---backend-61d75d03a7505500147ec7f6/issues/department-of-veterans-affairs/va-mobile-app/2078

Discovery for adding appointment requests to mobile app.

### DIFF
--- a/platform/engineering/request-for-comment/2022-01-14-mobile-schedule-appointments-discovery.md
+++ b/platform/engineering/request-for-comment/2022-01-14-mobile-schedule-appointments-discovery.md
@@ -1,0 +1,95 @@
+# RFC: Schedule appointments in VA mobile
+
+- Date: 2022-01-14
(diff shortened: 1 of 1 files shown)


kpethtel (commented):
I don't have the ability to approve this PR, but my one thought is that Andrew's recent work probably makes this doc irrelevant. It looks like he did a lot of the same investigation but started from a place of greater understanding and spent more time on it (this was all done in a day). At the time I wrote this, I didn't realize that direct scheduling and appointment requests were intermingled into a single flow, and I didn't understand the eligibility check process. Since then, we've done a lot of appointment request work, product planning, and also had multiple talks with the vaos team and our collective understanding of how this works and what we want has evolved. Looking at this doc, it looks like I was scheduling for major VA facilities, which seems to result in fewer eligibility checks. I don't even see the word "eligibility" in this doc. I think that may be a fundamental flaw in how I approached this investigation. It may be better to have Andrew review this doc and fold anything of value into his own work.

PR #36189, opened 2022-01-28
Created va-pdf-generation readme

JaredMostudio:
Populated with content from research


JaredMostudio (commented):
My write access may have been removed while writing this document.

nprimak (approved):

PR #36468, opened 2022-02-04
updating with feedback from engineers

jhouse-solvd:
additional information and resources added, in addition to breaking out sections for increased clarity

@department-of-veterans-affairs/vsp-operations

### DIFF
--- a/.github/ISSUE_TEMPLATE/kubernetes-onboarding-template.md
+++ b/.github/ISSUE_TEMPLATE/kubernetes-onboarding-template.md
@@ -1,57 +1,85 @@
 ---
-name: Kubernetes Onboarding Template
-about: For onboarding a new app to the VSP Kubernetes clusters
-title: '[Insert your app name here] - Onboarding to Kubernetes (EKS)'
-labels: 'operations, kubernetes, devops, eks, infrastructure, infrastructure-request'
-assignees: ''
+name: Application Onboarding Template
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/kubernetes-onboarding-template.md
```diff
@@ -1,57 +1,85 @@
...
+- The Infrastructure Team will assist you with gathering requirements _and_ performing key setup and configuration tasks 
 
 ---
-## Background/Context/Resources
-_Please link to an initiative or other documentation that desribes the need and purpose of the application_
 
-## Technical Notes
-_Additional technical information that helps to explain the desired setup and configuration_
+## Description of application
+_Please provide a brief description of the application, ie What does it do? Who does it serve?_
```
rmtolmach, line 25: Your use of "ie" here is correct, but in the other instances below here where you use "ie", it should be "eg".

rmtolmach (approved):
LGTM! 

My comment about i.e. vs e.g. is  definitely non-blocking 😄 . At my old office, for some reason, there was a poster by my cubicle about the difference between ie and eg 😂  so now I always think about it (for better or for worse)!

.github/ISSUE_TEMPLATE/kubernetes-onboarding-template.md
jhouse-solvd: Cool. I will remember that on the next iteration. :)

PR #37509, opened 2022-02-23
Create 2022-02-23-remove-gender-fax-from-mobile-api

jperk51:
RFC for updating Mobile API in regards to removing gender and fax

### DIFF
--- a/platform/engineering/request-for-comment/2022-02-23-remove-gender-fax-from-mobile-api
+++ b/platform/engineering/request-for-comment/2022-02-23-remove-gender-fax-from-mobile-api
@@ -1,0 +1,23 @@
+RFC: Removing gender and fax number from mobile api
+Date: 2022-02-23
+Related Issue: department-of-veterans-affairs/va-mobile-app#2565
+Background
+Currently gender and fax are both returned from the mobile API and displayed in the app. 
+
+Motivation
+Profile is removing gender and fax number. More details: https://dsva.slack.com/archives/C018V2JCWRJ/p1643395405843779
+
+Design
+The proposed change for the API would be to add an accessor to the iam_user model to pull gender from the IAM identity or return null if it doesn't exist. This would remove the risk of the profile changes breaking the mobile API by making gender no longer accessible through the user model. Otherwise we propose doing nothing in the API until we are certain these fields have been completely removed from the upstream services and confident most users have updated the app beyond the version where the app no longer shows the fields.
+
+Risks
+  In versions of the app that still include the fields, once profile removes them from the upstream services
+    Gender: Users will see the field in the app, but will see "This information is not available right now" 
+    Fax: Similarly users will still see the field, but will see a button asking them to add a fax number. This could result in a user attempting to add a fax number then returning to the profile screen and still seeing the "add a fax number" button. This seems fairly low risk, because people don't use faxes anymore. LOBs said they haven't sent a fax in 25 years. 
+
+  More details and screens for how the app will show empty gender and fax: department-of-veterans-affairs/va-mobile-app#2445
+
+Alternatives
+The most extreme alternative is to remove the fields all together. This would result in all users immediately seeing the states described above while they might still see the info as is on .gov. We decided against this because we didn't want to cause confusion in the meantime, and it seemed safer to wait until most users have updated the app to a version where the fields are no longer shown to completely remove them   
+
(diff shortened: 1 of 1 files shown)


lexicalninja (approved):

platform/engineering/request-for-comment/2022-02-23-remove-gender-fax-from-mobile-api
```diff
@@ -0,0 +1,23 @@
+RFC: Removing gender and fax number from mobile api
+Date: 2022-02-23
+Related Issue: department-of-veterans-affairs/va-mobile-app#2565
+Background
+Currently gender and fax are both returned from the mobile API and displayed in the app. 
+
+Motivation
+Profile is removing gender and fax number. More details: https://dsva.slack.com/archives/C018V2JCWRJ/p1643395405843779
```
travis-newby, line 8: I'd like to see some discussion here on the impacts of removing these from Profile. Something along the lines of: The mobile API – a module inside vets-api – must support multiple versions of the app in the wild. We do this through API versioning. If gender and fax are removed from Profile without proper mitigation measures (making this data available from somewhere), the mobile API cannot properly support older versions of the app. The ultimate impact is that app users will see error messages when trying to view or update their personal information.

platform/engineering/request-for-comment/2022-02-23-remove-gender-fax-from-mobile-api
```diff
@@ -0,0 +1,23 @@
...
+
+Design
+The proposed change for the API would be to add an accessor to the iam_user model to pull gender from the IAM identity or return null if it doesn't exist. This would remove the risk of the profile changes breaking the mobile API by making gender no longer accessible through the user model. Otherwise we propose doing nothing in the API until we are certain these fields have been completely removed from the upstream services and confident most users have updated the app beyond the version where the app no longer shows the fields.
```
travis-newby, line 11: We can tie this back to the motivation. Our proposal is to version our api, removing gender and fax from the new version while, at the same time, supporting older versions of the app by getting that data (or at least gender) from somewhere else. That'll let Profile remove them.

platform/engineering/request-for-comment/2022-02-23-remove-gender-fax-from-mobile-api
```diff
@@ -0,0 +1,23 @@
...
+
```
travis-newby, line 12: I also think that – probably outside of this RFC – we should have an ultimate goal of providing these services inside vets-api through another layer of abstraction and versioning.

platform/engineering/request-for-comment/2022-02-23-remove-gender-fax-from-mobile-api
travis-newby: I'd like to at least provide some details here of when we'll be able to remove the fields entirely. I don't have a set # in mind, but once older versions of the app reach some small percentage, we'll be able to fully remove all references to those fields in vets-api. My guess is that small percentage will be reached in 6 to 8-ish months.

platform/engineering/request-for-comment/2022-02-23-remove-gender-fax-from-mobile-api
jperk51: Yeah I can updated that to be more verbose about the situation.

platform/engineering/request-for-comment/2022-02-23-remove-gender-fax-from-mobile-api
jperk51: The proposal in this doc doesn't actually require versioning the API. Yes users will see the empty state conditions, detailed [https://app.zenhub.com/workspaces/va-mobile-frontendqa-60f1a34998bc75000f2a489f/issues/department-of-veterans-affairs/va-mobile-app/2445](here), but my perspective is that's fine. The reason we want to add an accessor in iam_user is to insulate ourselves from how deeply gender would be removed. For example, if profile just removes it from the database then we don't actually need our own accessor, but if they delete the existing accessor in the base user in vets-api then that would break our existing code.

jperk51:
Selected solution:
1. Update API User v1 by removing fax and gender fields. This works because the app hasn't been updated to use v1 for login.gov yet. 
2. App updates to use v1 (login.gov updates and removing fax and gender fields completely) (let’s call it app version XX)
 3. Once usage for app versions less than XX drop to X% update v0 of user to return “null” for gender
 
 Patrick and Travis will work to determine the desired % for step 3.

PR #38527, opened 2022-03-14
2678 schedule pending appt doc

kreek:


platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Overview
+
+### Objective
+The goal is to add the feature to the mobile app to schedule pending appointments. This functionality is currently available in `vets-website` so much of the logic can be utilized in the mobile app.
+
+The intended audience is for the engineering team.
+
+### Background
+Currently, the `vets-website` scheduling feature is as follows:
```
kreek, line 16: Maybe a [sequence diagram](https://plantuml.com/sequence-diagram) could help illustrate the current state?

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
```diff
@@ -0,0 +1,156 @@
...
+  - `/facilities_api/v1/ccp/provider?{coordinate information}`
+- Choose desired date and times then schedule appointment request and post appointment request
+  - `/vaos/v2/appointments` 
+
+### High Level Design
+Implementation of this feature will largely follow the logic listed in the background with a key difference to improve user experience. all consumed APIS will be aggregated at the beginning of the process in order to create a single new endpoint to provide all information needed to schedule a request. It will only list types of care that are supported, along with the facilities that support that type of service. This will benefit the process in multiple ways:
+   - Move any business logic from the front end to the back end 
+   - Simplify overall backend and front complexity. 
+   - Create a better UX experience. Users will know of a service and facility eligibility before selecting it and waiting for a response.  
+
```
kreek, line 49: ☝️ These are great reasons 👍

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
```diff
@@ -0,0 +1,156 @@
...
+## Specifics
+
+### Detailed Design
+The backend business logic will mirror the process explained in the background above except rather then only querying the selected types of service and facilities, all registered facilities will be queried to create a comprehensive list. If load times get unacceptably long due to front loading all queries, we will explore pre-caching results. 
+
+New endpoint data structure:
+```json
+{
```
kreek, line 57: Could this be in [JSON-API format](https://jsonapi.org/examples/)?

kreek: This seems very simple... which is great! Just surprised it took multiple calls before to get this data (and I'm sure the backend will still be complex).

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
```diff
@@ -0,0 +1,156 @@
...
+      - Is primary facility? if not, ensure veteran has had visit within 12-24 months to proceed
+      - Is facility allow scheduled requests?, 
+      - Is request limit reached?
+- If Community Cares facility, choose provider
+  - `/facilities_api/v1/ccp/provider?{coordinate information}`
+- Choose desired date and times then schedule appointment request and post appointment request
+  - `/vaos/v2/appointments` 
+
+### High Level Design
+Implementation of this feature will largely follow the logic listed in the background with a key difference to improve user experience. all consumed APIS will be aggregated at the beginning of the process in order to create a single new endpoint to provide all information needed to schedule a request. It will only list types of care that are supported, along with the facilities that support that type of service. This will benefit the process in multiple ways:
```
kreek, line 45: Running it through my grammar app...
```suggestion
Implementation of this feature will largely follow the logic listed in the background, with a key difference in improving user experience. We will aggregate all consumed APIs at the beginning of the process in order to create a single new endpoint to provide all information needed to schedule a request. It will only list types of care that are supported, along with the facilities that support that type of service. This will benefit the process in multiple ways.:
```

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
```diff
@@ -0,0 +1,156 @@
...
+   - Move any business logic from the front end to the back end 
+   - Simplify overall backend and front complexity. 
+   - Create a better UX experience. Users will know of a service and facility eligibility before selecting it and waiting for a response.  
+
+## Specifics
+
+### Detailed Design
+The backend business logic will mirror the process explained in the background above except rather then only querying the selected types of service and facilities, all registered facilities will be queried to create a comprehensive list. If load times get unacceptably long due to front loading all queries, we will explore pre-caching results.
```
kreek, line 53:
```suggestion
The backend business logic will mirror the process explained in the background above except rather than only querying the selected service and facilities, we will query all registered facilities to create a comprehensive list. If load times get unacceptably long due to front loading all queries, we will explore pre-caching results.
```

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
```diff
@@ -0,0 +1,156 @@
...
+### Open Questions and Risks
+Will aggregating all external requests at the beginning of the process cause long load times? If so, we may want to precache results. 
+
+Are endpoints we'll be consuming reliable enough to have a good user experience?
+
+### Work Estimates
+TODO
+
+### Alternatives
+Rather then relying on other endpoints, we could create these features from the ground up to ensure we have full control of all parts of development and maintenance but it has been a long established objective of the Mobile API to reuse backend integrations wherever possible and to not add complexity to an already complex system. Creating redundant logic would not serve either of these objectives
```
kreek, line 147:
```suggestion
Rather than relying on other endpoints, we could create these features from the ground up to ensure we have full control of all parts of development and maintenance but it has been a long established aim of the Mobile API to reuse backend integrations wherever possible and to not add complexity to an already complex system. Creating redundant logic would not serve either of these objectives.
```

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
aherzberg: No problem. Created. Were you thinking the diagram and the current background or in place of?

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
aherzberg: Sounds good. Done. Let me know if I missed any aspect of the JSON-API format

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
```diff
@@ -0,0 +1,170 @@
...
+The goal is to add the feature to the mobile app to schedule pending appointments. This functionality is currently available in `vets-website` so much of the logic can be utilized in the mobile app.
+
+The intended audience is for the engineering team.
+
+### Background
+Currently, the `vets-website` scheduling feature is as follows:
+
+![XL9BReCm4DrpYhb0Ue0igaIehRfgrQZT4tYGQOrDFKCaRbz3BOXYb6mGiZtly-RjH0klVMkoJu6dpICTuC_mMyfTs_QMTQ2yy11wCjpmaSruye7V7g94cNKACgXrIs_M35IUKFu8rSw7NoUmf4C7SZL9mEKIsJy-RG9LeBnB99spWkmq4RbnSX8QnaxE2OI-zXJ8yfoudhv2c00C4alFHJd2CeHURP3O5ihEhhgP3-jbflQk](https://user-images.githubusercontent.com/1910447/158439763-fd97627e-1bb7-4053-bb30-86436b64a2a4.png)
+
+- Pick type of care to schedule an appointment for
```
kpethtel, line 20: Do we know if this is specific to the user? We know there are some eligibility requirements around "user is registered at facility and does not have more than two outstanding requests at facility". Do these endpoints do those checks? I assume they do, but if they don't we could set up an overnight cron job that collects this info for all facilities and caches it, which would substantially improve response time.

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
```diff
@@ -0,0 +1,170 @@
...
+- User is given list of facilities to choose from
+  - Get facility data 
+  - `/vaos/v2/facilities?children=true&ids[]={facility id}&ids[]={facility id}`
+  - Note: facility id's come from registered facilities in `/v0/user`
+  - Get additional facility scheduling information such as services provided and if available to schedule request (including child facilities)
+  - `/vaos/v2/scheduling/configurations?facility_ids[]={facility_id}&facility_ids[]={facility_id}`
+- User chooses facility
+  - Check if able to schedule request at given facility and service 
+  - `/vaos/v2/eligibility?facility_id={id}&clinical_service_id={service_type}&type=request`
+  - Multiple checks and responses are done here:
```
kpethtel, line 34: So it looks like this is where those user specific checks occur. So maybe it is possible to cache the above results and then the only external requests we have to make for eligibility are for user data and this.

kpethtel (approved):
This is a well written doc. It's easy to understand. As we discussed via slack, this is for the V2 version of the process, which seems less convoluted than the V0 process. Unclear if we can use it yet or not. Per my comments below, if it's possible to pre-cache data, I think we should do that up front rather than wait for performance issues. If we can do that, it will make our code cleaner, faster, and more reliable. Then it seems like our endpoint could be as simple as:
```
fetch user's facilities
filter out facilities that don't offer the service (based on cached data)
check those facilities for user specific eligibility checks
```
Which sounds very manageable. 

Also, I think the change to up-front eligibility checks is a really good call.

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Open Questions and Risks
+Will aggregating all external requests at the beginning of the process cause long load times? If so, we may want to precache results. 
+
+This uses new VAOS endpoints that have not yet been finalized. By the time we finalize our implementation, theirs should also be, but we will need to ensure that is the case, so consumed APIs don’t change out from underneath us. 
+
+Are endpoints we'll be consuming reliable enough to have a good user experience?
+
+### Work Estimates
+4 weeks.
```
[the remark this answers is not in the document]
aherzberg: I forgot to put this in the work estimates, but I was thinking we could divide some work up to be done in parallel. Separate the eligibility check endpoint with the `create`. I'm good with still padding it a bit more.

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
```diff
@@ -0,0 +1,172 @@
...
+## Specifics
+
+### Detailed Design
+The backend business logic will mirror the process explained in the background above except rather than only querying the selected service and facilities, we will query all registered facilities to create a comprehensive list. If load times get unacceptably long due to front loading all queries, we will explore pre-caching results.
+
+New endpoint data structure:
+```json
+{
+   "data": {
+      "type": "schedule-appointment",
```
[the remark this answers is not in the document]
aherzberg: Per our discussion on slack, we landed on `appointment-eligibility` instead.

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
aherzberg: yes, that'd be a good way of caching it.

platform/engineering/design-docs/2022-03-11-schedule-pending-appt.md
kpethtel: The v0 version of this process (as documented here: https://github.com/department-of-veterans-affairs/va.gov-team/pull/35452/files) also fetched the past two years of appointments, and I speculated that it was how we fetched the available facilities. I believe two years is the cut off we've heard for facility eligibility, so that would be consistent. I'm really not sure why the web app fetches past appointments here, but it would be kinda crazy to have the front end fetch past appointments and have its own logic to determine eligibility. That logic should definitely be in the back end.

kpethtel:
> Would like to get everyone's opinion on how to build out the controllers. The doc currently has "One for checking eligibility called `appointment_eligibility_controller` and one for creating the appointment requests called `appointment_request_controller` but I just realized that isn't consistent with how we're using the `appointments_controller`, which is wrapping appointment requests in with regular appointments. To be consistent with that, we'd want to put the `create` in `appointments_controller`. This gets more convoluted when we add in direct scheduling as there will need to be a parameter for if we're creating a request or normal appointment (or deduce it based on the parameters like we do for cancels). I'm leaning toward putting it all in the `appointments_controller` anyway to stay consistent.

The decision to conflate appointments and appointment requests in the index was driven by business requirements. It made it possible for us to keep pagination on the back end. To do that, we jammed both types into a single model, so it's consistent in the sense that it's an index of appointment objects as we defined them.

It might be worth thinking about what happens when we add direct scheduling. If we try to do both through the same controller action, we'll need query params and branching logic. Depending on the data we get back from VAMF, it may also not make sense to keep them in the same model. So I would argue that we may want to keep them in separate controllers.

jperk51:
Comment on eligibility: instead of making a controller for it we might want to create new policies for viewing and/or scheduling appointments and add those to the list in the user serializer. This way the authorizedServices object will get updated with these new services. 

Ex:   
Payment Info Policies: https://github.com/department-of-veterans-affairs/vets-api/blob/b7913b4668b7eb5b4b9fc7a13521f7769dcc4286/modules/mobile/app/controllers/mobile/v0/payment_information_controller.rb#L8-L11

User serializer Authorized Services: https://github.com/department-of-veterans-affairs/vets-api/blob/b7913b4668b7eb5b4b9fc7a13521f7769dcc4286/modules/mobile/app/serializers/mobile/v0/user_serializer.rb#L95-L101

aherzberg:
Could that make the user endpoint too susceptible to failure of upstream services? All we can reliably cache is facility name and address. Eligibility checks include limits on requested appointments at a facility which could change anytime so all eligibility checks couldn't be cached.

I was originally envisioning just including the scope of [2726](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2726 ) into the new policy. If we can find a good way of working in the rest of the logic, I'm open to it. That could keep it simple.

jperk51:
> Could that make the user endpoint too susceptible to failure of upstream services? All we can reliably cache is facility name and address. Eligibility checks include limits on requested appointments at a facility which could change anytime so all eligibility checks couldn't be cached.
> 
> I was originally envisioning just including the scope of [2726](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2726) into the new policy. If we can find a good way of working in the rest of the logic, I'm open to it. That could keep it simple.

Ah good catch. I was misunderstanding what was meant by eligibility. Makes sense to have something separate for checking all the different conditions we proposed during estimation yesterday. https://github.com/department-of-veterans-affairs/va-mobile-app/issues/2726 covers the policy for this I think. Sorry about that.

aherzberg:
Ended up needing to overhaul these designs into 5 endpoints. Reasons listed at the end of the alternatives section. Let me know if you have questions.

PR #38731, opened 2022-03-18
Create health-api-engine-refactor-decision.md

patrickvinograd:


jjulian:
+1 for option 2 or 3 (yes, refactor). In addition to providing a fresh clean workspace for new teams, the folks doing the refactor will garner understanding of the logic. Ideally the team would tackle this as their first task, but getting it all cleaned up for them first makes sense too.

ewilson-adhoc:
I’m in favor of refactoring the existing MHV-related APIs into engines as part of the general push to work towards modularizing vets-api. One of the goals that my team (IET) has in the medium term is to work towards identifying which teams have ownership over which sections of code. 

I think both options 2 and options 3 are viable approaches. It seems to me that it would make the most sense to begin with option 2 (refactoring it into a single engine). I think it is useful to have teams begin to split sets of related code into engines. 

Out of curiosity, what level of connection/dependencies do the different logical services (SM/Rx/Health Records) have with each other? I like the idea of approaching a problem consistently across the codebase (making the MHV engine more similar to VAOS), but before I would advocate for splitting it into multiple engines + the implementation overhead that would involve, I’d need to understand this code a little more deeply.

patrickvinograd:
Additional discussion in Slack: https://dsva.slack.com/archives/C0229UT3HSR/p1647562495984849

PR #39184, opened 2022-03-25
38276 update sequence diagrams for check-in

randomsync:
This PR adds sequence diagrams check-in and pre check-in experience.


randomsync:
Draft for early review. I've added the Sequence diagrams for check-in process. Will continue to work on pre check-in and other diagrams.

PR #39309, opened 2022-03-29
Update staging-review.md

jimuhh:
Added instructions on submitting non-launch-blocking accessibility issues for 508 review.


chaseakirby (approved):
I approve this change. I like the placement with the other VFS actions after the meeting.

shiragoodman:
@jimuhh I updated the language because it was not correct. We want the VFS team to communicate only the accessibility issues that are not addressed prior to launch, not ALL of them.

shiragoodman (approved):

PR #39355, opened 2022-03-30
Slack user removal request: Change icon to 🤖 

rmtolmach:
Instead of the Slack icon seen here:
image
This PR will change it to a 🤖 .  I asked Cory Trimm for his emoji opinion and he voted for the robot face. 

According to the [icon_emoji docs](https://api.slack.com/methods/chat.postMessage#arg_icon_emoji), we needed to explicitly set `as_user` to false (even though it defaults to false), so that's why the 👋  hadn't been showing up.


jbritt1 (approved):

jhouse-solvd:
🤖  ftw

PR #39843, opened 2022-04-12
Check In Experience - rearrange and connect CW with vista-api-service

stephenBarrs:
Clean up of CIE container diagram and connect clinician workflow to vista-api-service.  

Question:  Does CW use CHIP at all?  In the past CW called CHIP to access vista-api-service, but IIRC CW now calls vista-api-service directly. Is there any remaining dependency on CHIP from CW?


randomsync (approved):
lgtm

PR #39979, opened 2022-04-13
Mobile rx refill

aherzberg:
Have some open questions listed in the considerations sections that should be answered before merging this but should be a good starting place.


kpethtel (approved):
This makes sense and seems like a modest amount of BE effort. The one thing I don't understand (without digging into the code) is the how "active" works in the index. You say it's a separate endpoint, but it's also a status option for the filter param. So I'm not sure I understand why both params are needed.

aherzberg:
> This makes sense and seems like a modest amount of BE effort. The one thing I don't understand (without digging into the code) is the how "active" works in the index. You say it's a separate endpoint, but it's also a status option for the filter param. So I'm not sure I understand why both params are needed.

I'm not totally sure of the functional difference myself. The Prescription Controller that also uses this RX client (though I don't think the controller itself has been used for years) has a separate refill_status param that does this: 

`case params[:refill_status]
  when nil
client.get_history_rxs
when 'active'
client.get_active_rxs
end`

uses a separate MHV service all together even though there's a filter option for `active`. I'll connect with MHV team to see if there's actually a functional difference to `get_active_rxs` and just filtering out active status on `get_history_rxs`

jperk51 (approved):
LGTM

PR #40000, opened 2022-04-14
Add RFC 4 for PST-CIT team

dginther:
https://github.com/department-of-veterans-affairs/va.gov-team/blob/dg/pst-cit/RFC-4/teams/vsp/teams/pst-cloud-isolation-team/RFC/4-Removing-Kong-as-an-intermediary-between-vets-api-and-LH-in-cases-of-upstream-dependencies.md

### DIFF
--- a/teams/vsp/teams/pst-cloud-isolation-team/RFC/4-Removing-Kong-as-an-intermediary-between-vets-api-and-LH-in-cases-of-upstream-dependencies.md
+++ b/teams/vsp/teams/pst-cloud-isolation-team/RFC/4-Removing-Kong-as-an-intermediary-between-vets-api-and-LH-in-cases-of-upstream-dependencies.md
@@ -1,0 +1,131 @@
+### Discussed in https://github.com/department-of-veterans-affairs/va.gov-team/discussions/39753
+
+<div type='discussions-op-text'>
+
+<sup>Originally posted by **madebydna** April  7, 2022</sup>
+- Start Date: 2022-04-06
+
+# Summary
+
+The Kong API gateway is currently an integral part of the overall `dsvagovcloud` AWS architecture. The Lighthouse team is [planning to retire Kong](https://vfs.atlassian.net/wiki/spaces/PST4/pages/2138996933/Lighthouse+Migration+Plan) by migrating to a new Apigee gateway hosted on GCP. The idea is that the domain `api.va.gov` will be routed to Apigee, and a new domain `platform-api.va.gov` will expose the API endpoints of the `vets-api`. The major users of this new domain are the va.gov website and the mobile app and they will need to adjust their API calls accordingly.
+
+Our thinking is that in the first phase of the new `dsvagovcloud` architecture we will no longer need an API gateway, since Kong's main role was to funnel traffic to the vets-api and Lighthouse, respectively. Most requests to the **new** domain will need to go directly to `vets-api`. It is our opinion, therefore, that a reverse proxy would be sufficient to handle incoming traffic at this stage. (Further down the road, we may need to introduce an API gateway component to better support the migration of the `vets-api` from a monolithic Rails app to a more microservices-based system.)
+
+This RFC is about one of Kong's current functions of handling cases in which Lighthouse APIs are used as an upstream dependency of `vets-api`. These types of requests first go (either via Kong or via the current revproxy) to `vets-api`, and then back through Kong and a forward proxy to Lighthouse. Once Kong is removed, `vets-api` would need to make those requests _directly_ to Lighthouse via the forward proxy.
+
+This RFC outlines the hurdles and intricacies of removing Kong as an intermediary between the `vets-api` and Lighthouse and kicks off the discussion on our approach.
+
+# Motivation
+
+We're inviting comments on our initial investigation into what needs to be taken into account in removing Kong. The overall benefit of removing Kong is to achieve a simpler architecture with a more straightforward request flow. By investigating and exposing potential pitfalls in the removal of Kong, we can make the path to the new architecture smoother. Our initial research and your input will save our team development time and prevent us going down avenues that would have to be abandoned.
+
+**Use-cases:**
+- Website and mobile API calls to `vets-api` that make secondary API calls to Lighthouse under the hood
+
+**Expected outcome:**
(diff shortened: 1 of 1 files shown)


dginther:
From Patrick Vinograd:
I think there's a chicken and egg problem here. In the diagram above, step (4) shows the link from forward proxy to Lighthouse as being via 4. api.va.gov/services/va_facilities/ But that URL is still routing to Kong, until such time as Apigee is deployed and VA DNS is updated to route traffic for api.va.gov to Apigee. But perhaps we're not ready to do that cutover until Kong is decommissioned for these internal uses?

At minimum I think the more accurate thing would be to say that step (4) routes to Lighthouse at the configured upstream for the forward proxy, which is blue.qa.lighthouse.va.gov:443. In other words going straight from vets-api (via fwdproxy) to the upstream API application server, much like we do with other upstream dependencies for VA.gov. But, this carries with it some measure of risk because now there's no rate limiting or access control in front of that Lighthouse endpoint (because those capabilities were provided by Kong).

My question is - my understanding is that Apigee is intended as the public API gateway for external-to-VA Lighthouse clients. But vets-api is not really one of these - it's an internal-to-VA client. So what is the URL/route that internal Lighthouse clients are meant to use? Is such thing deployed yet, and can we point from vets-api->fwdproxy-> ? Otherwise, we need to be clear on the risk of bypassing Kong.

dginther:
From Patrick Vinograd:
I think there's a chicken and egg problem here. In the diagram above, step (4) shows the link from forward proxy to Lighthouse as being via 4. api.va.gov/services/va_facilities/ But that URL is still routing to Kong, until such time as Apigee is deployed and VA DNS is updated to route traffic for api.va.gov to Apigee. But perhaps we're not ready to do that cutover until Kong is decommissioned for these internal uses?

At minimum I think the more accurate thing would be to say that step (4) routes to Lighthouse at the configured upstream for the forward proxy, which is blue.qa.lighthouse.va.gov:443. In other words going straight from vets-api (via fwdproxy) to the upstream API application server, much like we do with other upstream dependencies for VA.gov. But, this carries with it some measure of risk because now there's no rate limiting or access control in front of that Lighthouse endpoint (because those capabilities were provided by Kong).

My question is - my understanding is that Apigee is intended as the public API gateway for external-to-VA Lighthouse clients. But vets-api is not really one of these - it's an internal-to-VA client. So what is the URL/route that internal Lighthouse clients are meant to use? Is such thing deployed yet, and can we point from vets-api->fwdproxy-> ? Otherwise, we need to be clear on the risk of bypassing Kong.

dginther:
From Andrea Singh: 
@patrickvinograd I've changed step 4 in the diagram to correctly use the url configured in the fwd proxy, which is blue.qa.lighthouse.va.gov:443. Also, really good question about coordinating the switch to Apigee and which endpoint to configure in the forward proxy for the transition!

dginther:
From Andrea Singh: 
@patrickvinograd I've changed step 4 in the diagram to correctly use the url configured in the fwd proxy, which is blue.qa.lighthouse.va.gov:443. Also, really good question about coordinating the switch to Apigee and which endpoint to configure in the forward proxy for the transition!

dginther:
From Beau Grantham:
In the new architecture diagram, routing directly to blue.[env].lighthouse.va.gov bypasses the Lighthouse gateway and its policies. All clients (including VA.gov / vets-api) should go through the Lighthouse gateway.

Unless, as Patrick points out, Lighthouse approves unrestricted backchannel traffic for very specific scenarios/sources.

dginther:
From Beau Grantham:
In the new architecture diagram, routing directly to blue.[env].lighthouse.va.gov bypasses the Lighthouse gateway and its policies. All clients (including VA.gov / vets-api) should go through the Lighthouse gateway.

Unless, as Patrick points out, Lighthouse approves unrestricted backchannel traffic for very specific scenarios/sources.

dginther:
> From Beau Grantham: In the new architecture diagram, routing directly to blue.[env].lighthouse.va.gov bypasses the Lighthouse gateway and its policies. All clients (including VA.gov / vets-api) should go through the Lighthouse gateway.
> 
> Unless, as Patrick points out, Lighthouse approves unrestricted backchannel traffic for very specific scenarios/sources.



> From Beau Grantham: In the new architecture diagram, routing directly to blue.[env].lighthouse.va.gov bypasses the Lighthouse gateway and its policies. All clients (including VA.gov / vets-api) should go through the Lighthouse gateway.
> 
> Unless, as Patrick points out, Lighthouse approves unrestricted backchannel traffic for very specific scenarios/sources.

From Andrea Singh:
Right, I will add a note that this is a temporary URL to bypass Kong and directly access the API application server. Or do you have any suggestions what domain to use instead to make the scenario clear, e.g. tbd-apigee-dns.va.gov or something like that?

dginther:
> From Beau Grantham: In the new architecture diagram, routing directly to blue.[env].lighthouse.va.gov bypasses the Lighthouse gateway and its policies. All clients (including VA.gov / vets-api) should go through the Lighthouse gateway.
> 
> Unless, as Patrick points out, Lighthouse approves unrestricted backchannel traffic for very specific scenarios/sources.



> From Beau Grantham: In the new architecture diagram, routing directly to blue.[env].lighthouse.va.gov bypasses the Lighthouse gateway and its policies. All clients (including VA.gov / vets-api) should go through the Lighthouse gateway.
> 
> Unless, as Patrick points out, Lighthouse approves unrestricted backchannel traffic for very specific scenarios/sources.

From Andrea Singh:
Right, I will add a note that this is a temporary URL to bypass Kong and directly access the API application server. Or do you have any suggestions what domain to use instead to make the scenario clear, e.g. tbd-apigee-dns.va.gov or something like that?

patrickvinograd:
Those most recent comments by Beau and Andrea are the crux of the matter for me.
* End state should be that VA.gov, like any other LH client, access APIs through the LH API gateway
* But if the intended LH API gateway is Apigee hosted at api.va.gov, then we have an order of operations problem.

So either:
1) There's a different **temporary** Apigee endpoint that VA.gov can use. 
2) There's a different **permanent** Apigee endpoint that VA.gov can use, since it's an internal client. 
3) There's a temporary non-Apigee endpoint that VA.gov can use (like set up a short-lived Kong instance in front of the DVP endpoints)
4) We accept the risk of **temporary** direct API access without an API gateway.

PR #40319, opened 2022-04-21
Updated commands for capture slo screen shots and generate service docs

RachalCassity:
The commands to capture screen shots in Grafana and to generate the SLO report were updated. The Watir default timeout was changed to 5 minutes to allow proper time for charts to load.


ericboehs:
Could you update the README here as well?

PR #41249, opened 2022-05-11
Update research-plan-template.md

jimuhh:
Added VA platform links.


claytonzook:
The changes from this ticket are reflected in the current Research Plan. Looks like at some point this review got missed, so I'm going to close without merging incase this merge would revert a more recent change

PR #41318, opened 2022-05-12
Update conversation-guide-template.md

jimuhh:
Added links and new headers to beginning. Minor re-ordering of steps and a few text edits.


claytonzook (approved):
I agree that the new headings clarify things. Looks good!

Issue #45256, opened 2022-08-03
[Application Hosting and Deployment] Automate the synchronization of ArgoCD GitHub teams using Infrastructure-as-Code


jhouse-solvd:
## Description
_What details are necessary for understanding the specific work or request tracked by this issue?_

## Background/context 
_Any additional context for the reader to know, if applicable (optional, please delete if unused)_ 

## Technical notes
_Notes around work that is happening, if applicable (optional, please delete if unused)_

---
## Tasks
- [ ] _What things need to happen?_

## Acceptance Criteria 
- [ ] _What are the specific things that must be true in order for this to be accepted?_

---
### Reminders
- [x] Please attach your team label and any other appropriate label(s) (operations, devops, and needs-grooming will automatically be applied as part of the template)
- [x] Please connect to an epic (this will typically be done by the Platform Operations PM or TL)

ph-One:
This one needs more discussion and possibly broken down into a ticket-per-workflow. Holding a call at 2pm ET today to go over this with the team.

oseasmoran73:
Linking [Create GitHub Teams that Grant ArgoCD Permissions](https://vfs.atlassian.net/wiki/spaces/OT/pages/2306408539/Create+GitHub+Teams+that+Grant+ArgoCD+Permissions)

jbritt1:
Other relevant commits: https://github.com/department-of-veterans-affairs/va.gov-team/commits/master/.github/workflows/ops-application-onboarding-form.yml

jhouse-solvd:
This is no longer relevant. Please see #44415

PR #45881, opened 2022-08-18
update onboarding and promotion automation

jbritt1:
### Description
first pass at breaking things out and making the automation for onboarding/promotion more modular and reusable

### Relations
- #45256

### Notes
there is a step that checks to see if the application exists or not, my changes will see if a specific env exists already for the given application name. it uses a `.select()` filter and if the application name is not found at all it would result in an empty string being set for the variable. i have accounted for that with a simple `if` statement making sure if the var is not set or is empty it is then forced to be `false` since the app name would have to not exist in this case and the rest of the logic could still work / build off the result.

image

also, i have ripped out hard-coding of the word `Onboarding` and set new outputs/inputs involving a `task`. this helps make things more dynamic and i have included some logic that will _hopefully_ work the way i think it will and detect if the `application_environment` is `dev` and set the `task` to `onboarding`. for all other envs besides dev, _if my changes work as expected_, the `task`  should get set to `promoting`.

image

in the steps to find out if the form submitted was for onboarding or promotion i use the same ID for both jobs. my hope here is that only one will actually get used / set and the other will be skipped due to what the filter finds and it won't be an issue. could be totally wrong on that and having two jobs with the same ID may break. figured it was worth a shot to at least test out the theory of only one getting set.

image

finally, i added a form for application promotion yesterday that uses a dropdown for setting an `application_environment` variable. since that var is used throughout the rest of the automation, i added the same dropdown to the application onboarding form and forced a single option of `dev` so we can reuse the existing pieces without too much more additional logic.

image

### Feedback
any


.github/workflows/ops-application-onboarding-or-promotion-form.yml
```diff
@@ -0,0 +1,60 @@
+name: Application onboarding or promotion
+
+on:
+  issues:
+    types:
+      - opened
```
rmtolmach, line 6: Is this workflow going to run on every issue opened within the `va.gov-team` repo?

.github/workflows/application-onboarding-or-promotion-automation.yml
```diff
@@ -0,0 +1,95 @@
...
+        type: string
+      maintainer_github_user:
+        description: Maintainer GitHub username
+        required: true
+        type: string
+      task:
+        description: Onboarding or promoting
+        required: true
+        type: string
+  workflow_call:
```
rmtolmach, line 22: Is it weird that we are using both `workflow_call` and `workflow_dispatch`? Wasn't exactly sure about the difference, so I pulled these from the docs:

> Use on.workflow_call to define the inputs and outputs for a reusable workflow

> When using the workflow_dispatch event, you can optionally specify inputs that are passed to the workflow.

.github/workflows/application-onboarding-or-promotion-automation.yml
```diff
@@ -0,0 +1,95 @@
...
+  workflow_dispatch:
+    inputs:
+      application_name:
+        description: Application name
+        required: true
+        type: string
+      application_environment:
+        description: Application environment
+        required: true
+        type: string
```
rmtolmach, line 13: Should we make this a dropdown with a list of `options`?

.github/workflows/application-onboarding-or-promotion-automation.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Check if application already exists
+        id: exists
+        run: |
+          application_exists=$(cat vsp-infra-argocd/chart/values.yaml | yq -o=json | jq -e '.projects[] | select(.name == "${{ inputs.application_name }}").environments | any(. == "${{ inputs.application_environment }}")' || true)
+          if [ -z "$application_exists" ]; then export application_exists=false; fi
+          echo "::set-output name=application_exists::$(echo ${application_exists})"
+        
+      - name: If application exists
+        if: steps.exists.outputs.application_exists == 'true'
+        uses: KeisukeYamashita/create-comment@v1
```
rmtolmach, line 61: [peter-evans/create-or-update-comment](https://github.com/peter-evans/create-or-update-comment) has more stars in GitHub than the one you're using. We're using it to comment on the SOCKS access requests here: https://github.com/department-of-veterans-affairs/devops/blob/master/.github/workflows/pst-amt-roster-check.yml#L105 Should we try to go with a more widely-used action?

.github/workflows/ops-application-onboarding-or-promotion-form.yml
```diff
@@ -0,0 +1,60 @@
...
+        id: issue-parser
+        with:
+          template-path: .github/ISSUE_TEMPLATE/application-onboarding.yml
+
+      - name: Promote application
+        if: contains(github.event.issue.title, 'Application to Promote')
+        uses: stefanbuck/github-issue-parser@v2
+        id: issue-parser
+        with:
+          template-path: .github/ISSUE_TEMPLATE/application-promotion.yml
```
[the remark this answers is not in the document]
jbritt1: Already exists https://github.com/department-of-veterans-affairs/va.gov-team/issues/new?assignees=va-vsp-bot&labels=application-promotion%2Coperations&template=application-promotion.yml&title=Application+to+Promote%3A+%5Bapplication-name%5D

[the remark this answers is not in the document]
jbritt1: image

.github/workflows/ops-application-onboarding-or-promotion-form.yml
jbritt1: It will run, check to see if the title includes one of the two options after this and skip the rest if it doesn't match. So pretty negligible on timing, resources, and run-minutes.

.github/workflows/application-onboarding-or-promotion-automation.yml
jbritt1: This was just copy/pasted from the existing work Matt had already done. I have no strong opinions and not married to one action over the other.  I could go either way and don't care much as long as it does what we need it to do.

.github/workflows/application-onboarding-or-promotion-automation.yml
jbritt1: This is an action, not the actual form. The form does have a dropdown with options.

image

The same will be true for the onboarding form once these changes are merged, although it only has one option of `dev`

image

mleclerc00 (approved):

.github/workflows/ops-application-onboarding-or-promotion-form.yml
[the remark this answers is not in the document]
rmtolmach: oh! nice 👍

.github/workflows/ops-application-onboarding-or-promotion-form.yml
```diff
@@ -0,0 +1,60 @@
...
+jobs:
+  onboard-or-promote:
+    runs-on: ubuntu-latest
+    name: Onboard or promote application
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Onboard application
+        if: contains(github.event.issue.title, 'Application to Onboard')
```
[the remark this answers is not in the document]
rmtolmach: 👍 It's a good start and we can always iterate on this if it actually becomes a problem.

.github/workflows/application-onboarding-or-promotion-automation.yml
rmtolmach: oh, okay. I see. thanks!

.github/workflows/application-onboarding-or-promotion-automation.yml
rmtolmach: I was thinking a dropdown could be in the `workflow_dispatch`:

image

jbritt1: Good call, will look into it!

.github/workflows/application-onboarding-or-promotion-automation.yml
rmtolmach: oh, I didn't realize Matt was already using this action. I had come across this guidance in regards to gh Actions: https://vfs.atlassian.net/wiki/spaces/OT/pages/1474595236/Guidance+for+Github+Actions+usage so I always try to at least see how active the action is. https://github.com/KeisukeYamashita/create-comment hasn't been modified in two years and is not used by many projects. using `create-comment` feels like more of a vulnerability than `create-or-update-comment`, but i haven't read through the code of either...

jbritt1: Fair. I will look into swapping it out.

PR #48554, opened 2022-10-21
update rules for CSP

joeniquette:
@bosawt or @bramleyjl Can you please change the values to the raw values in the table I added for address, email address? Also if we dont have a log for those differences can you please add one.

Ref: https://app.zenhub.com/workspaces/identity-5f5bab705a94c9001ba33734/issues/department-of-veterans-affairs/vets-api/10805


bosawt:
I think on documentation we should leave out the email difference for now. I'm finding it's extremely unreliable for the moment on MPI (I think we should still have a log but for internal use only for now as it'd be purely informative)

bosawt:
Also address isn't currently pulled from credentials and compared to MPI at the moment, so is functionality we need to add. When it is added the log will match the pattern of everything else though, which is:
`"Attribute mismatch, #{type} in credential does not match MPI attribute"`

PR #50134, opened 2022-11-29
Delete design-intent-collaboration.md

CherylEvans:
@it-harrison Deleting file.  We are deprecating this template. Hopefully deleting the file is the right path.  That is the instruction I was given in my ticket task.


it-harrison:
> @it-harrison Deleting file. We are deprecating this template. Hopefully deleting the file is the right path. That is the instruction I was given in my ticket task.

@CherylEvans sounds good I think you can merge the change.

it-harrison (approved):

Issue #857, opened 2019-08-07
[Call Center] Issue Triaging


joanneesteban:
## User Story
As a user of VA.gov, I need answers to my issues that I submit through IRIS and the sign in help form. 

## Goal
Provide answers to users and route issues to the triage team when appropriate. 

## Acceptance Criteria
- [ ] Get caught up on issues that need to be answered
- [ ] Route issues to triage team

## Definition of Done
- [ ] **Attached to a Milestone** (when will this be completed?)
- [ ] **Attached to an Epic** (what body of work is this a part of?)
- [ ] **Labeled with Team** (`product support`, `analytics-insights`, `operations`, `triage`, `tools-improvements`)
- [ ] **Labeled with Practice Area** (`backend`, `frontend`, `devops`, `design`, `research`, `product`, `ia`, `qa`, `analytics`, `call center`, `research`, `accessibility`, `content`)
- [ ] **Labeled with Type** (`bug`, `request`, `discovery`, `documentation`, etc.)

PR #51543, opened 2023-01-05
Update form-confirmation-page.md

taylojill:
@GnatalieH Hi, I used another image to try to reduce the blurriness. Also, should I add a caption and alt-text to the image? cc @humancompanion-usds


humancompanion-usds:
@taylojill - This is a fine update and we can merge it. But what we ultimately want to do is get the documentation out of here and into the Design System. That's what issue #857 is about.

taylojill:
Hey @humancompanion-usds I'll work with Natalie to get it figured out today. Thank you!

PR #53852, opened 2023-02-17
Create .devcontainer/devcontainer.json

arilivigni:
Add devcontainers config with larger host requirements to support the large files (git-lfs) in the repo


arilivigni:
@mchelen can you review when you get a chance? This would be so users can use codespaces to work directly to commit and push files.

PR #56117, opened 2023-03-30
Proposed runbook changes for managing dynamic content

jilladams:
@RLHecht I don't know if this format makes sense to your team, so I opted to make a Pull Request rather than edit the doc directly. But I think this captures the steps we talked about. 

Easier preview of the changes here: https://github.com/department-of-veterans-affairs/va.gov-team/pull/56117/files?short_path=08fdc78#diff-08fdc788e00b9066bd2fa255b4e00a7c2b294cbbd20cbef6359fcc0b27ed4471

Two pieces that're muddy for me: 
1. In a case like BTSSS, where this will be a React widget ongoing (afaik), does your team still eventually edit Drupal to move the widget content / CTA back into Drupal rich text fields? 
2. For a case like BTSSS, on launch we'll need to flip the Flipper, and you'll need to Publish the Drupal content. Is that simultaneous? 😬 

Happy to talk through what I did here if it doesn't make sense, and feel free to reject the change and modify the doc however makes more sense to you all, since you own the process.


RLHecht:
Thanks for working on this, @jilladams. i think how you've set it up makes a lot of sense. To answer your 2 questions:

1. If a widget needs to permanently stay on the page (like BTSSS) we wouldn't do anything with moving content back into Drupal. So the last steps would be publishing the Drupal page with the react widget and validating it's working properly on Prod.
2. We hadn't run into this Flipper/React widget timing situation before (a team had never used a flipper to hide the react widget entirely). I know you worked with Megan on the timing of this for BTSSS, but we should talk about this for future work and make sure we note the order of operations. With other product teams, they've just told us the day they are launching their tool to whatever the starting percentage is and we've published the Drupal page once we knew the tool was live. 

I made some small copy tweaks to your branch but also had one other comment/question.

1. I'm getting a little confused between when we're talking about when the react widget itself is 100% launched (in which case Sitewide CAIA can publish the Drupal page with the react widget added) and when the new tool is launching to 100% (in which case the widget would be removed). Or maybe you're only talking about the new tool launching to 100% and if so, then I think we need to reword some of these steps.

jilladams:
Might be easiest to try and find a time to talk it through. I don't think I get the difference between launching to 100% vs. 100% launched. Do you mean: turning up from 75>100, vs.: we are all done now and leaving it on for everyone?

jilladams:
@RLHecht I do think getting this merged would be wise. If you have a min to look it over again.

RLHecht:
@jilladams sorry this was stalled for so long but am merging this now!

RLHecht (approved):

jilladams:
better late than never! Merging. (Randi: made sure this takes into account your change yesterday to update team name.)

PR #62938, opened 2023-08-01
Added appointment terms

outerpress:
I added some common appointment terms. Some of these were refined in the recent appt list redesign and request flow content review. Others came from the 2021 appt list redesign, and others were before my time.


outerpress:
@DanielleThierryUSDSVA I just wanted to get this back on your radar if possible.

DanielleThierryUSDSVA:
Thanks, @outerpress! It's on my radar. It's just on a very long list of things to do/think through. And because this is a shared resource, I want to make sure that everything is 100% what we want before merging. Thanks so much for checking in!

outerpress:
Great @DanielleThierryUSDSVA thank you for the update!

outerpress:
FYI @DanielleThierryUSDSVA I started documenting [the internal terms we use to describe data displayed in appointments](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/appointments/va-online-scheduling/feature-reference/data-definitions.md). The starting goal is to align, esp when looking at our content docs (so when we reference placeholder data we know the difference between [Location] and [Physical Location]. Eventually we'd want to align these across teams and potentially show differences. 

I didn't add this to the content brief yet because it's not about Veteran-facing language, but we are thinking about where it might live at a higher level at some point. FYI also @laurwill @kristenmcconnell

laurwill:
Hi @outerpress , thanks for the reminder about this! I connected with Danielle, and we probably won't merge the pull request because not all of it is applicable to the content brief (like method of attendance/modality). But it's back on our radar, and we'll look at the brief to see which parts of the PR we should add there. We'll let you know when we have updates!

outerpress:
Thanks for following up @laurwill ! I'll add this topic for our next sync. Just FYI we're working on definitions for the data we display across appointment tools (very rough draft [here](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/appointments/appointments-reference/data-reference/data-definitions.md#instructions-prepare-for-visit)). I'm not sure what the line will be between that and the content brief, so we can chat more about it :)

outerpress:
@laurwill I'm going to close this pull request out, we'll keep the conversation going about other ways to track the shared definitions across teams 👍

Issue #62257, opened 2023-07-19
Add Delivery Manager to New VFS Team Member gh template


shiragoodman:
## User Story
As the Governance team PM, I want new VFS Delivery Managers to complete the same tasks that PMs are required to complete so that they sync with me and learn the same things that the PM learns.

Assignee: @erinrwhite 
Peer Reviewer: @CherylEvans 

## Description
With more and more Delivery Managers being included on VFS teams, it makes sense that they have role-specific tasks that they should complete in order to support their team on VA.gov. Since we are not prioritizing taks specifically aimed at the Delivery Manager, we should include them alongside the Product Manager.

## Impacted Artifacts
[New VFS Team Member template](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/.github/ISSUE_TEMPLATE/orientation-epic.md) (editable version)

## Tasks
- [x] Add "& Delivery Managers" alongside the Product Manager task list

## Peer Review
_To be completed by peer reviewer_
- [x] User story and acceptance criteria are met

## Acceptance Criteria
- [x] Delivery Managers have role specific tasks to complete as they partake in Platform Orientation

### Refinement
- [x] Ticket has user story, description, tasks, and acceptance criteria
- [x] Ticket has [additional appropriate labels](https://vfs.atlassian.net/wiki/spaces/PPST/pages/2220359751/WIP+Tech+Debt+Ticket+Organization), if applicable
- [x] Ticket has been sized 
- [x] Once all above items are checked, add the 'Ready' label.
### Planning
_If this ticket is picked up from the Backlog mid-sprint, connect with Shira to ensure the below items are completed correctly_
- [x] Ticket has been discussed as a team at Planning
- [x] Ticket has been assigned to appropriate initiative or quarterly epic
- [x] Ticket has been assigned to a Sprint
- [x] Ticket has been moved to the Planned pipeline in the Governance team board
- [x] Ticket as an assignee and peer reviewer

erinrwhite:
[Pull request ready for review](https://github.com/department-of-veterans-affairs/va.gov-team/pull/63116) @CherylEvans 

Also making a small update change "Preferred pronouns" to "Pronouns" to match style in the Design Standards and general best practices (discussed with @shiragoodman with a thumbs up!)

CherylEvans:
@erinrwhite Looks great! I like the other update you noted and changed; good catch! :-)

shiragoodman:
[...]

PR #63116, opened 2023-08-03
Update orientation-epic.md

erinrwhite:
Updating template to 

- add Delivery Managers per #62257 
- change "Preferred pronouns" to "Pronouns" to match style in the Design Standards and general best practices (discussed with @shiragoodman with a thumbs up!)


CherylEvans:
@erinrwhite Looks great!  I like the other update you noted and changed; good catch! :-)

PR #65439, opened 2023-09-12
Update va-notify README.md

mjones-oddball:
Overhauling our documentation and combining with the playbook.


bevnobev (commented):
Looks great!

PR #67067, opened 2023-10-06
Update vetsapi-flipper-ui-access.yaml

bdech:
VFS Team Roster has been deprecated and replaced with Atlas. This PR removes mention of VFS Team Roster and is replaced with Atlas


rmtolmach:
@bdech do you know if the Governance team and VFS teams are using Atlas now? Is the VFS spreadsheet not being updated anymore? I am not sure where that stands...

bdech:
I am not entirely sure. This thread sounds like Governance is now using Atlas. As far as VFS teams I don't know.

PR #67082, opened 2023-10-06
Replace VFS Team Roster with Atlas

bdech:
VFS Team Roster has been deprecated. 

This PR removes all mentions/links related to VFS Team Roster, and replaces with Atlas


rmtolmach (approved):
Looks good on my end.

PR #68214, opened 2023-10-24
Create technical-documentation.md

eselkin:
Created a general directory for facility-pages. Would love some feedback some of the specifics like which pages we call facility pages.

This is related to issue: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/15260


jilladams:
I am biasing for action and merging this, so the page is findable. We can always move it around as needed.

Issue #70171, opened 2023-11-17
Update Platform team labels in  the Platform Orientation gh templates


shiragoodman:
## User Story
As a Platform Support team member, I want to be able to find tickets my team is responsible for so that I can easily take the necessary action.

As an Analytics & Insights team member, I want my teams label to be removed from tickets we are no longer responsible for so that our backlog is not cluttered with other teams' tickets. 

As a Governance team member, I want my teams label to be removed from tickets we are no longer responsible for so that our backlog is not cluttered with other teams' tickets. 

Assignee: @shiragoodman 
Peer Reviewer: @allison0034 @rferguson13 

## Description
Platform Support team is going to be taking over the task of adding new VFS teams and new VFS Team members to the VFS Team Roster, soon to be Platform Atlas. In order to ease with the transition, we need to add the Platform Support team label (`platform-tech-team-support`) to both templates. In addition, we need to remove both Analytics and Insights team labels (`analytics-request` and `analytics-insights`) from both templates as they are no longer responsible for this work.

The `governance-team` label can be removed from the New VFS Team template, however, at this time, the label should remain in place on the New VFS Team Member template as Governance team still provides support adding new users to the vets.gov-write gh team. Platform Support will likely pick up this task at a later date.

Since we're updating labels, the urls will change, meaning the urls where these templates are linked out on Platform Website need to be updated. This should only impact the [Guidance for New teams and team members page](https://depo-platform-documentation.scrollhelp.site/getting-started/guidance-for-new-teams-and-team-members).

## Impacted Artifacts
[Guidance for new teams and team members](https://depo-platform-documentation.scrollhelp.site/getting-started/guidance-for-new-teams-and-team-members)
New VFS Team template in gh
New VFS Team Member template in gh

## Tasks
- [x] Confirm the correct gh label for Platform Support team (`platform-tech-team-support`) with @alyssagallion and/or @jknipes 
- [x] Create a pull request for the New VFS Team gh template and update the following:
   - [x] Add the Platform Support team label
   - [x] Remove the Governance team label
   - [x] Remove both Analytics & Insights team labels
- [x] Create a pull request for the New VFS Team Members gh template and update the following:
   - [x] Add the Platform Support team label
   - [x] Remove both Analytics & Insights team labels
- [x] After both PRs are approved and merged, update the links to the templates on the Guidance for new teams and team members page on Platform Website
- [x] After PW publishes, confirm that links direct you to the correct template with the correct labels

[...]

PR #71301, opened 2023-12-05
Update orientation-epic.md

shiragoodman:
updated per ticket [#70171](https://github.com/department-of-veterans-affairs/va.gov-team/issues/70171)

### DIFF
--- a/.github/ISSUE_TEMPLATE/orientation-epic.md
+++ b/.github/ISSUE_TEMPLATE/orientation-epic.md
@@ -38,7 +38,7 @@ As a new VFS team member, I need to complete onboarding activities so I can succ
 
 ### General
 
-- [ ] For [General orientation](https://depo-platform-documentation.scrollhelp.site/getting-started/calendly-orientation-sessions#OrientationSessions-Generalorientation) sessions, either attend a live session (schedule via Calendly) or watch the latest recorded session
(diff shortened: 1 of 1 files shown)


rferguson13:
@shiragoodman - A couple of things:

- I just noticed that we're inconsistent on how we capitalize "Collaboration Cycle." Sometimes we capitalize "cycle," and sometimes we don't. We should be consistent - and probably capitalize both words.
- It might be nice to link the most recent general orientation video (first bullet under "General").

allison0034 (approved):

rferguson13 (changes requested):
A couple of things:
- I just noticed that we're inconsistent on how we capitalize "Collaboration Cycle." Sometimes we capitalize "cycle," and sometimes we don't. We should be consistent - and probably capitalize both words.
- It might be nice to link the most recent general orientation video (first bullet under "General").

rferguson13 (approved):

shiragoodman:
@rferguson13 thanks for the suggestions! I'm 100% on board with ensuring we write Collaboration Cycle consistently each time. However, I don't want to link the youtube video from the ticket. The Youtube recording changes every 3 months when we host and rerecord a live session (we do this to prevent the content from getting out of date). We have never had an issue with teams not being able to find the recording from Platform Website, also.

Would you mind rolling back your changes? I was going to do it but I think it would be good practice for you :)

rferguson13 (approved):

rferguson13:
@shiragoodman Done! Triple check that I didn't mess something up? 😂

shiragoodman:
all good. Thanks @rferguson13 ! merging now.

PR #71803, opened 2023-12-12
Update dynamic-content-process.md

megzehn:
Hi @RLHecht, 

These are some terms that I think would be helpful to define upfront. Please edit away.... these definitions are my understanding, but I may be off! 

Thank you so much for putting this together!


RLHecht (approved):

RLHecht:
This is a great addition, Megan. Thanks for looking at this!

PR #71874, opened 2023-12-13
Add documentation for MAP STS endpoint

dickdavis:
This PR adds documentation for the MAP STS endpoint in a newly created "Service Account Endpoints" directory.


bosawt (approved):
looks good!

PR #71933, opened 2023-12-14
First draft of AP broken links RFC.

timcosgrove:
AP Broken links epic: https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14454

The Accelerated Publishing team is seeking comment on a proposal to rework how broken link checks are done for CMS content.

Questions, alternate proposals, and requests for clarification are all very welcome. Please do not merge this issue; it is our intention to invite discussion on the pull request and to use that discussion to further shape the RFC.

### DIFF
--- a/platform/cms/accelerated_publishing/rfc/broken-link-checking.md
+++ b/platform/cms/accelerated_publishing/rfc/broken-link-checking.md
@@ -7,7 +7,7 @@
 
 _Explain the current state of the feature._
 
-We currently have two link checking systems in place around CMS and Content Build.
+We currently have two link checking systems in place around CMS and Content Build: a Content Build broken link check plugin and a Drupal module [Node Link Report](#nlb).
 
 ### Content Build broken link checks
 
@@ -16,12 +16,6 @@ Content Build runs a [broken link check](https://github.com/department-of-vetera
 1. Content Build has knowledge of every page & file it is handling in the form of a list of paths
 2. Content Build parses each HTML page, finds any links, images, or other files, and checks each of those links to see if it is in the list from
 3. If it does not find that item in the list from step 1, it is a broken link.
-
-### Node Link Report
-
-[Node Link Report](https://www.drupal.org/project/node_link_report) is a helpful Drupal module that checks in-page content and provides notifications if it contains broken links. It checks internal and external links, with some limitations.
-
-There is no problem with Node Link Report continuing to operate as it is now. This RFC does not propose any changes to the VA.gov CMS's use of Node Link Report.
 
 ## Motivation
 
@@ -124,3 +118,9 @@ _What other alternatives solutions were considered, why weren't they chosen?_
 **We considered an architecture for Content Build and Next Build where the two systems would be more tightly coupled.** In particular, Next Build content would be included as part of the 'files' that Content Build manages. This would effectively allow the Content Build broken link system to continue, because it would have access to all files at once.
 
 We abandoned this idea because it would make Next Build dependent on Content Build, which means we would see no speed gains from Next Build. It is also generally a good idea to ensure that systems are not interdependent on each other, as they become harder to disentangle later. Our goal ultimately is to retire Content Build, and so we do not want to depend on its systems for this critical functionality.
+
+### <a name="nlb"></a>What about Node Link Report?
+
+[Node Link Report](https://www.drupal.org/project/node_link_report) is a helpful Drupal module that checks in-page content and provides notifications if it contains broken links. It checks internal and external links, with some limitations.
+
+There is no problem with Node Link Report continuing to operate as it is now. This RFC does not propose any changes to the VA.gov CMS's use of Node Link Report.
--- a/platform/cms/accelerated_publishing/rfc/broken-link-checking.md
+++ b/platform/cms/accelerated_publishing/rfc/broken-link-checking.md
@@ -14,7 +14,7 @@ We currently have two link checking systems in place around CMS and Content Buil
 Content Build runs a [broken link check](https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/stages/build/plugins/modify-dom/check-broken-links/index.js) for every build via a custom plugin. This works as follows:
 
 1. Content Build has knowledge of every page & file it is handling in the form of a list of paths
-2. Content Build parses each HTML page, finds any links, images, or other files, and checks each of those links to see if it is in the list from
+2. Content Build parses each HTML page, finds any links it thinks are internal to  pages, images, or other files, and checks each of those links to see if it is in the list from
 3. If it does not find that item in the list from step 1, it is a broken link.
 
 ## Motivation
--- a/platform/cms/accelerated_publishing/rfc/broken-link-checking.md
+++ b/platform/cms/accelerated_publishing/rfc/broken-link-checking.md
@@ -1,0 +1,105 @@
+# RFC: Broken Link Checking for VA.gov content
+
+- Date: 2023-12-13
+- Related Issue: [Accelerated Publishing broken link checking (epic)](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/14454)
+
+## Background
+_Explain the current state of the feature._
+
+We currently have two link checking systems in place around CMS and Content Build.
+### Content Build broken link checks
+Content Build runs a [broken link check](https://github.com/department-of-veterans-affairs/content-build/blob/main/src/site/stages/build/plugins/modify-dom/check-broken-links/index.js) for every build via a custom plugin. This works as follows:
+1. Content Build has knowledge of every page & file it is handling in the form of a list of paths
+2. Content Build parses each HTML page, finds any links, images, or other files, and checks each of those links to see if it is in the list from
+3. If it does not find that item in the list from step 1, it is a broken link.
+
+### Node Link Report
+[Node Link Report](https://www.drupal.org/project/node_link_report) is a helpful Drupal module that checks in-page content and provides notifications if it contains broken links. It checks internal and external links, with some limitations.
+
+There is no problem with Node Link Report continuing to operate as it is now. This RFC does not propose any changes to the VA.gov CMS's use of Node Link Report.
+
+## Motivation
+_Why do we want to change the current implementation? What problem(s) does the change solve?_
+
+### Issues with Content Build broken link checking
+The current broken link check plugin for Content Build has a number of limitations:
+1. This process cannot check broken links that are external to the va.gov domain, because it inherently only has knowledge of what Content Build  is currently building.
+2. The process relies of having *total* knowledge of all files. It cannot run broken links checks on a single page, for example, because if it is only building a single page, it has no way to know whether the other destinations are valid.
+3. Because of the above point, it is also not possible to use Content Build link check if some portion of the total site pages are being built by another system.
+4. Inserting this at build time adds time to deploy. It is difficult to determine exactly how long the current broken link check process takes, but it is in the range of 2-3 minutes.
+
+### Reasons motivating the change
+As part of the [Accelerated Publishing Initative](https://github.com/department-of-veterans-affairs/va.gov-cms/issues/6577), we are creating a new build application for VA.gov content ([Next Build](https://github.com/department-of-veterans-affairs/next-build)). The move to the new system poses some challenges for the current broken link checking system:
+
+1. Moving from Content Build to Next Build will happen incrementally. There will be a period during which both systems will be responsible for portions of CMS content. This presents problems for points 2. and 3. above. Content Build _requires_ building the entire set of CMS pages in order to do its link checks; but, Content Build will no longer have the entire set of CMS pages once we start shifting some content to Next Build.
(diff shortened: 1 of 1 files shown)


platform/cms/accelerated_publishing/rfc/broken-link-checking.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Crawler Proposal
+
+A straight-forward way would be to create a site crawler that checks va.gov pages, follows any links found, and reports any that return a non-200/300 response. This would operate as a CI process on VA.gov's published content.
+
+All content published by Content Build and Next Build is listed in sitemap.xml files. These URLs can be read by a site crawler and the contents of the URL scanned.
+
+Advantages:
+
+- This would be relatively simple to create.
```
swirtSJW, line 71: Checking links from within the VA network is not as simple as it sounds.  Bumped my head into this many times.  Checking links to other gov't sites is not as simple as it sounds either.   Many of them return responses that lie.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
```diff
@@ -0,0 +1,126 @@
...
+- May be able to be largely addressed by of-the-shelf software. or entirely. [The Node.js ecosystem has several broken link checking packages](https://www.npmjs.com/search?q=broken%20link), for example.
+- Both internal and external links would be collected together into one report.
```
swirtSJW, line 73: This would be a huge win.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
```diff
@@ -0,0 +1,126 @@
...
+
+Challenges:
+
+- Content would no longer be checked as it is built, which is an expectation VA.gov has due to the current architecture.
+- Determining how frequently to scan will require some experimentation.
```
swirtSJW, line 78: Scanning once per hour if it can handled it ought to be sufficient to cover existing build.  Though it may function better as the way we currently do content releases.  When one full scan completes, it kicks off the next.
Would actually want it to run a bit on the slow side so that we don't have race conditions where someone creates an event  (next-build) with a link to a content build based page.  The event would land first with a link to a page that has not been built yet by content-build.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
```diff
@@ -0,0 +1,126 @@
...
+
+It is only recommended that we do this when we move to an incremental publishing scenario, and also that the link checks happen independently of deployment. In other words, link checks would be triggered by publishing of pages, but they would happen after deployment, rather than before.
+
+Advantages:
+
+- Errors would be caught at build time and rectified sooner.
+
+Challenges:
+
+- Publishing new groups of pages that link to each other could present challenges, as they are not on production yet and cannot be checked there.
```
swirtSJW, line 93: There is a similar race condition problem in either proposal.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
```diff
@@ -0,0 +1,126 @@
...
+
+Some of these risks have been discussed above.
+
+1. The current system (build everything and check everything) is inherently a problem for speedy publishing, but it _does_ check every internal link every publishing run. What we are proposing would actually capture more broken links than the current system, but there is potential for a broken link to be live on VA.gov for longer before it is reported.
+2. The current system has existed in its current form for some time and over time has reached something like stability. We would be building a new system, and inevitably there will be issues that will need to be resolved as the new system comes into use.
+3. Depending on architecture choices, we risk reintroducing slowdown into the publishing process. We want to avoid this as much as possible while ensuring the integrity of linked content on VA.gov.
+
+## Alternatives
+
+_What other alternatives solutions were considered, why weren't they chosen?_
```
swirtSJW, line 116: [A similar proposal was pitched by this design doc](https://github.com/department-of-veterans-affairs/va.gov-team/pull/14092) several years ago.
It accounted for notifications to editors.

This proposal does not seem to account for notification of anyone.  Would this tie into the same notification system we currently have (Slack notifications for CMS Helpdesk)?

swirtSJW (commented):
This looks like a great start.  I added a couple comments and questions.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
```diff
@@ -0,0 +1,126 @@
...
+
+- Content would no longer be checked as it is built, which is an expectation VA.gov has due to the current architecture.
+- Determining how frequently to scan will require some experimentation.
+- This sort of checker would create additional traffic to the site. This should not present a problem to our current S3-as-web server architecture, but we should validate that the reverse proxy will not have trouble with the additional traffic, and we should ensure this traffic is ignored by analytics systems.
+
+### Build time Proposal
+
+We could check links on pages as they are built. However, these checks would need to check against the live production site, which requires making a request to a URL and verifying the response. This is _significantly_ slower than the current Content Build system.
+
+It is only recommended that we do this when we move to an incremental publishing scenario, and also that the link checks happen independently of deployment. In other words, link checks would be triggered by publishing of pages, but they would happen after deployment, rather than before.
```
swirtSJW, line 85: This bit makes it sound like publishing a page is the only source of the broken link.  As in, if I publish a new version of my page, I need all the links on that page checked.  In reality though, I also need all the links TO that page that exist on the site to be checked, because my content may not have added new links, but it might have changed my page location, or I unpublished a page.   So I do not think a build time solution could cover enough all on its own.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
timcosgrove: Fair point; though, for prod at least, this is likely to run outside the network. We're concerned that a Veteran can reach these pages, not that they're accessible from within the VA network, so it would actually be prudent to *not* run the scan from within the VA network.

"Gov't sites that give bad responses" is noted though; can you speak more to specifics?

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
timcosgrove: Yeah, the broken in-links problem from archiving a page can't be handled by a check on the changed pages alone. It requires full scans. I'll update this to be clearer and to call out that specific problem.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
timcosgrove: I will add something to speak to this question. I think addressing notifications is outside the scope of this proposal, and so we would default to replicating the current notification system, I.E. Slack notifications that are monitored by CMS User Support. 

User Support folks have said that the current system is not at all ideal for them, because it's a lot of repetitive notification of the same broken links with every content build, and each one needs to be scanned for new issues mixed in with the existing ones, which presents a significant cognitive load. 

Editor notifications would be a dream, if we can get them to respond to them. It would need to be emails; or at least, the research for outdated content notifications indicated that editors vastly prefer email notification to anything else. 

In either case, I think whatever broken link checks we come up with should produce machine-formatted results that can be ingested by other systems. I think User Support could greatly benefit from a dashboard-type application that helps them keep track of broken links and allows them to clearly see both new broken links as well as their age, i.e. how long it takes to resolve them. 

The same data could also be used to drive editor notifications, though I think a lot of user research would need to be done there to know what the best way to notify editors is.

timcosgrove: I will add something to speak to this question. I think addressing notifications is outside the scope of this proposal, and so we would default to replicating the current notification system, I.E. Slack notifications that are monitored by CMS User Support. 

User Support folks have said that the current system is not at all ideal for them, because it's a lot of repetitive notification of the same broken links with every content build, and each one needs to be scanned for new issues mixed in with the existing ones, which presents a significant cognitive load. 

Editor notifications would be a dream, if we can get them to respond to them. It would need to be emails; or at least, the research for outdated content notifications indicated that editors vastly prefer email notification to anything else. 

In either case, I think whatever broken link checks we come up with should produce machine-formatted results that can be ingested by other systems. I think User Support could greatly benefit from a dashboard-type application that helps them keep track of broken links and allows them to clearly see both new broken links as well as their age, i.e. how long it takes to resolve them. 

The same data could also be used to drive editor notifications, though I think a lot of user research would need to be done there to know what the best way to notify editors is.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
timcosgrove: I think trying to account for race conditions with respect to content publishing/availability between next-build and content-build is a losing game, frankly. The only way to ensure that the two are in sync would be to yoke next-build to content-build; to only allow next-build to release when content-build releases. This effectively would mean we see no speed improvements of any kind until content-build is 100% retired.

I *do* think that there will need to be an education process for CMS User Support in diagnosing broken link issues, since there will be some 'intermittent' broken link issues resulting from them not being yoked together. I.e., a full site scan might report a broken link from a next-build page to a content-build page because content-build is lagging behind. I think this would need to become part of the assessment/triage process for broken link response that currently is happening.

See also the comment below about notifications, a potential broken link dashboard, etc.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
timcosgrove: True, noted and will clarify.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
swirtSJW: There are a lot of site that when you try to curl them return something other than a friendly 200 response.  [Most recent example](https://dsva.slack.com/archives/C02730UEZPS/p1702659644868359)

So the method has to build in an ignore list  similar to the config we have in node_link_report  
image

In fact it might make sense to have it use the same ignore list so that the cms and the FE are in agreement about what they are ignoring.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
swirtSJW: You already laid the groundwork for the scan to be machine readable.  Drupal could easily ingest that (it was supposed to be the next step) and put notifications in a dashboard.  I do not recommend email for this.  You'd send out 5 emails to editors in one section.  one person would fix it and the others would be confused by their copy of the email that cried wolf.

One email a month for outdated content does not create much cognitive overload, but 20 a day about broken links certainly would.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
timcosgrove: We could. We'd have to figure out how to make the CMS list available more publicly, so that it's available to an out-of-VA-network process.

I'd also really like to explore if we can figure out what leads the site to return a 403 to a CURL process but not to a browser request. At the end of the day both are just making HTTP calls with different headers, or at least that's my naive understanding. If we can figure out what allows a browser request to succeed, maybe we can also automate checks to those same domains.

I understand why the domain list is necessary, but it frankly worries me a bit because of not being able to differentiate legitimate 404s/403s from 'unfriendly' ones.

swirtSJW: See that is the beauty of having a public repo.  The data is right there in readable yml format :smiley: 
https://github.com/department-of-veterans-affairs/va.gov-cms/blob/main/config/sync/node_link_report.settings.yml#L7-L98   


Why do sites get goofy.... too many reasons.  Could we masquerade as a browser user agent... we could.  Not really supposed to, but we could.  Though even when I masquerade as chrome, that one example site still gives me a 403.

swirtSJW: Though in fairness, not all of those need ignoring. Some are only ignored because of the TIC intercepting, while the others are truly bad responses.

timcosgrove: Right, and again I think it's much better to be checking this stuff from outside the TIC. I'd be curious to see if we get the same kinds of misbehaving responses from these sites if the TIC is not contributing. 

If the purpose of our scans is to check that a browser can hit a given site, I don't think it's a problem to modify user agent. We're not scraping or spidering except to verify the site works when hit with a browser.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
timcosgrove: Actually, thinking about this: the full site scan would run from the outside. If there are pages interlinking and one is not published and the other is not, that is not the scanner's fault; it's reporting what is actually true on the site at that time.

If that's a persistent problem we can look at grouped releases, but that's a different problem than this.

platform/cms/accelerated_publishing/rfc/broken-link-checking.md
timcosgrove: Regardless, notifications aren't part of this proposal. That'll be its own project at some point.

PR #72741, opened 2023-12-29
Restructure PAW folder in anticipation of future initiatives & to organize info that is/was specific to launch

jilladams:
FYI @cindymerrill @FranECross 

Also note: the QA artifact for PAW has a new URL. I've updated that in the Collab Cycle ticket, in prep for Staging Review.


cindymerrill:
@jilladams Updated link to research folder in Research Repository.

Issue #70474, opened 2023-11-22
[Discovery] Reading/Writing Profile Info in Current Portal


hinzed1127:
## Description

Document where a Veteran's profile info is getting sourced from in the existing portal.

Since we will (at least initially) be existing in a bifurcated state, with a VA.gov BTSSS tool living alongside the current site, we want to make sure they are operating with a single source of truth with respect to a Veteran's Profile

## User Story
As a Veteran, I want my profile information consistently sourced from the same place, so that updating profile information in either VA.gov or the current BTSSS portal leads back to the same origin point.

## Definition of Done
- [ ] Source of Veteran Profile information in BTSSS is defined
- [ ] Flow of Veteran Profile Updates in BTSSS is mapped and defined.

---
## How to configure this issue
- [ ] **Attached to an Epic**

hinzed1127:
Via Tim Hawkins on the Migration team:
> Dynamics is really an autonomous system. I don't think there is any direct correlation between the veteran profile data in Dynamics and any other system outside of the ICN which the portal gets from the user signing in.
>
> From a portal perspective, the user doesn't actually have the ability to update their profile information. When they submit the form on the profile page, it creates a task with the Dynamics, and I believe that gets routed to a travel clerk.

I need to clarify on if there's any other information derived from the ICN, but it sounds like this information is pretty isolated. It would probably be on us to make sure any updates between the dynamics db and external profile info is kept in sync.

hinzed1127:
Lots of good information in some more follow up from Tim.

## The quick summary 
As I understand it, here's the flow of data when a user logs into BTSSS:
- User logs in
- BTSSS Portal receives the ICN and EDIPI from the authentication process, uses them to fetch:
    - Caregiver/donor relationship data
    - ﻿﻿EHRM and VistA appointment records
- ﻿﻿That data, along with claim status info that gets fetched and updated in a nightly job, is stored locally in the Dynamics db
- To reduce external calls post log-in, the dynamics db becomes the source of truth for a user's BTSSS portal session. BTSSS portal uses its own UUID to associate all of this data with a user.
---
## Full email exchange
I asked the following questions:
- Is the ICN just used as a unique ID for storing Veteran info in Dynamics, or does it serve other purposes?
- Is there any other authenticated data that the portal grabs besides the ICN to pre-populate other information in the portal, such as that profile information, or is everything manually entered and self-contained within the Dynamics db once someone is authenticated?

His response:

[...]

PR #73133, opened 2024-01-08
Create data-flow-sequence-diagram.md

hinzed1127:


products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
```diff
@@ -0,0 +1,33 @@
+The basic sequence for most BTSSS is straightforward:
+
+```mermaid
+sequenceDiagram
+autonumber
+actor veteran as Veteran
```
kjduensing, line 6: It might be good to add a simplified version of the authentication sequence here. I wouldn't go into detail, but enough to show there's some authN happening. Maybe a link to the SiS authN sequence diagram: https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/identity/Sign-In%20Service/auth_flows/web_oauth.md.

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
```diff
@@ -0,0 +1,33 @@
...
+participant travel-pay-api as Travel Pay API
+participant dynamics-db as Dynamics DB
+
+veteran->>vets-website: Navigate to travel pay
+
+activate vets-website
+
+vets-website->>vets-api: vets-api endpoints (/claims, /appointments, etc.)
+vets-api->>travel-pay-api: travel-pay endpoints (/claims, /appointments, etc.)
+travel-pay-api->dynamics-db: fetch data via dynamics portal plugins
```
kjduensing, line 20: Keep in mind that these diagrams are from the perspective of our work on VA.gov. Dynamics is (or probably should be) hidden to us. I don't want to miss any nuance though - what's the benefit of us including the Dynamics system as part of this seq diagram?

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
```diff
@@ -0,0 +1,33 @@
...
+vets-api-->>vets-website: vets-api response(s)
+
+vets-website-->>veteran: website UX
+
+deactivate vets-website
+```
+
+What's to be determined is how we deal with any profile information updates. According to the BTSSS team, when a veteran first logs in to the portal, their contact information "is dependent on their data in enrollment services." Any contact updates from that point onward appear to be isolated within the dynamics database used by the portal.
+
+Since we'll have users coming from VA.gov with their information coming from their VA.gov profile/VA Profile, we'll need to account for any discrepancies between the two and update accordingly, with the VA.gov information being the source of truth.
```
kjduensing, line 33: This might be _A Whole Thing_ later, agreed. Also agreed that VA.gov info should be the source of truth. 

From my perspective, the responsibility of maintaining an unsynced database lies on the owner of the database (API team in this case).

When you say
> we'll need to account for any discrepancies between the two and update accordingly

1. Who's we?
2. What are you thinking accounting for discrepancies looks like?
3. What would updating accordingly look like (our side? their side?)

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
```diff
@@ -0,0 +1,33 @@
...
+veteran->>vets-website: Navigate to travel pay
+
+activate vets-website
+
+vets-website->>vets-api: vets-api endpoints (/claims, /appointments, etc.)
+vets-api->>travel-pay-api: travel-pay endpoints (/claims, /appointments, etc.)
+travel-pay-api->dynamics-db: fetch data via dynamics portal plugins
+
+dynamics-db-->>travel-pay-api: claims and appoinments data
+travel-pay-api-->>vets-api: travel-pay endpoint response(s)
```
kjduensing, line 23: Thinking about performance - what would adding in Redis caching look like here? I'm not sold on if we need to do that with claim status responses, but might be good to think about for later.

Also, do you think it's necessary to add caching in?

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
```diff
@@ -0,0 +1,33 @@
+The basic sequence for most BTSSS is straightforward:
```
kjduensing, line 1: Nice high-level diagram! I'm wondering to what level we should address these points on the Readiness Review ticket:

* Data Flow Diagram:
  * This diagram must illustrate the following specifics.
  * What data is collected or used, and where, including information such as credentials used by this system?
  * Where is the data is stored and how, including information such as any encryption used?
  * How is the data transferred, including information such as any encryption used?
  * Who accesses the data and in what capacity (read or read-write)?
  * What is the audit trail of data access and manipulation?

kjduensing: I'm leaning toward answering these in-diagram from our (VA.gov) perspective and not try to answer for the backing API team.

What do you think?

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
hinzed1127: I guess it might not be necessary to include that final bit with Dynamics. My thinking has more to do with what I called out in the [non-diagram portion below](https://github.com/department-of-veterans-affairs/va.gov-team/pull/73133/files/e6401b311d57c70e08f437d2c61b5c71be46a97e#diff-91a82379c52f487631d0c9cfc93ebc184a168052b5634a156a4815d517a770f9). It's mainly to highlight that VA.gov and the existing portal are drawing from different DBs that don't talk to each other.

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
hinzed1127: Since it's likely that the requests made by the travel pay API will have the biggest performance implications, I think caching from there to start is the most important. I'm not sure the impact caching vets-api responses on top of that will be as critical, but it's worth looking into. 

> what would adding in Redis caching look like here? 

I'm not quite sure what the additional complexity/level of effort doing this in `vets-api` entails

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
hinzed1127: A lot of my concerns here are with the fact that there will be at least some period of time where the existing portal will run in parallel with the VA.gov functionality. We want to make sure both are able to function effectively.

> 1. Who's we?

In this case, we is the VA.gov BTSSS team responsible for the `vets-website` and `vets-api` portion of things

> 2. What are you thinking accounting for discrepancies looks like?

Since the BTSSS API team is essentially a proxy for accessing the BTSSS data, travel claims should all be sourced and updated from there. My concern is mostly the profile information. It sounds like profile info is populated upon BTSSS account creation (see https://github.com/department-of-veterans-affairs/va.gov-team/issues/70474 for more details). I'm not sure how profile information works in creating BTSSS appointments or claims, but the problematic scenario I'm imagining is something along these lines:
 
- Veteran profile info on VA.gov doesn't match info on BTSSS
- Veteran goes to create travel claim on VA.gov
- This includes requests to the travel pay api
- The travel pay api receives a request to file a claim and uses erroneous profile information from the BTSSS DB

> 3. What would updating accordingly look like (our side? their side?)

That's where how BTSSS currently utilizes profile info comes into play. If we could bypass any mismatches entirely and just include VA.gov's profile info in the claim initialization, then the travel-pay-api could just take that request data as the source of truth. Assuming VA.gov as the authoritative source, It would be a nice if BTSSS could do an update/insert sort of operation, where a claim is created while also updating any profile information that is a mismatch between VA.gov and BTSSS (i.e, updating the BTSSS profile to match VA.gov's, not the other way around)

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
hinzed1127: I think that's fine, whether that's in the diagram or any additional information below the diagram

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
hinzed1127: They updated the SiS docs and broke that link, but I believe[ this it the same diagram](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/identity/Products/Sign-In%20Service/auth_flows/cookie_oauth.md#sequence-diagram)?

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
kjduensing: Let's add it in the diagram. Let me know if you need help fitting something into diagram form

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
kjduensing: Agreed, let's add a note or some type of reference to the diagram.

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
hinzed1127: Updated [here](9df298bc600a8c8028cc810040100cb4b97b6961) with some additional notes

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
hinzed1127: I removed explicit mention of dynamics

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
hinzed1127: updated in 236c079de4e0976d37505b4cdefb32c454b3ed4c

products/health-care/beneficiary-travel/engineering/data-flow-sequence-diagram.md
hinzed1127: Added in c57c7b79a53ae276626d412da51b99afe996ed27

hinzed1127:
@kjduensing I cleaned up the non-diagram stuff to add in things discussed above, i.e., a few outstanding questions related to certain steps in the diagram, and an appendix linking to the Sign-in Service auth sequence diagram. Considering that sufficient for now.

PR #77303, opened 2024-02-28
Update setup-real-user-monitoring.md

acrollet:
Adds info on how to globally override clicked element titles.


Mottie (approved):
NIce! Thanks for updating this doc

PR #79920, opened 2024-04-03
Update ADR-callback-api-to-receive-status-from-pega.md

cloudmagic80:
This PR adds to the previous version of the ADR file. 
1.  Added controller logic for PEGA callback
2. Updated Routes file
3. Add endpoint


michaelclement (approved):
lgtm

PR #80357, opened 2024-04-09
Update how-to-generate-analytics-report-for-stakeholders.md

benbrasso:
Updated value for 1; please validate


loripusey:
If we have the Uptime value, wouldn't the downtime be: 100 - Uptime? I'm not understanding your calculation.

PR #83832, opened 2024-05-23
Update faqs.md

benbrasso:
Added common question, but needs an answer response


products/health-care/checkin/faqs/faqs.md
```diff
@@ -23,4 +23,5 @@ Created: September 25, 2023 <br/>
 | **Does eCheck-in work if the same phone number is associated with more than 1 Veteran?** | Yes, it does work, but only if the two or more Veterans don't both have appointments on the same day. If they do, VEText will error out and they will be told via VEText upon texting "check in" to 53079 to check in at the desk with an MSA. |
 | **How are Pre-Check-in & Check-in links associated with an appointment?** | Pre-check-in links are associated with a specific appt or appt group, but check-in links do not have a specific appt association, there just has to be at least one appt eligible for check-in for a uuid/link to be created under the current business rules |
 | **What do you 3-digit & 5-digit station codes mean?** | Station with 3 digits would mean a VAMC and all sites that fall under that VAMC; Station with 5 digits is the unique facility either just the VAMC or a specific CBOC. |
+| **Why doesn't PCI display time zones eith appointment times?** | xx |
```
brianseek, line 26:
```suggestion
| **Why doesn't PCI display time zones with appointment times?** | Currently, the data we get back from vista-api has a mix of times with TZ offsets and without. So there is no reliable way to determine TZ for all. |
```

benbrasso-agile6: @brianseek looks good. Is there a quick explanation to add after that, which explains why? E.g., vista-api has trouble identifying time zones because a 3-digit station might have facilities across different time zones and does not have a way to identify the 5-digit station with a correct time zone. 

Is that right? Close? Did I make that up?

brianseek: I don't know the why of it all. I wouldn't say the stations is the reason. I just know that we analyzed this in the past and determined that we couldn't programmatically solve the problem with the data that we get from vista-api.

PR #89536, opened 2024-07-29
[88993] Add POARequest storage proposal

jvcAdHoc:
Proposal for storing Power of Attorney Request Data in vets-api

[88993](https://app.zenhub.com/workspaces/accredited-representation-management-team-64d0dc51d3e8f4788ac6ef96/issues/gh/department-of-veterans-affairs/va.gov-team/88993)


products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
```diff
@@ -0,0 +1,160 @@
...
+Code below is for illustrative purposes. These snippets are not likely to be the final implementation.
+
+### Column level encryption
+The [blind index gem](https://github.com/ankane/blind_index) allows for exact matching queries on encrypted columns. This will be used on fields that users will search or filter by. Existing KMS encryption will be used on the fields that are for display purposes only.
+
+#### The Model
+```ruby
+class PowerOfAttorneyRequest < ApplicationRecord
+  has_kms_key
+  has_encrypted :consent_data, key: :kms_key, **lockbox_options
```
nihil2501, line 24: I think that consent data is a particularly important data point for a given POA request. We have heard that some VSOs intend to reject any request that don't consent to everything. There is a chance that product would like users to be able to filter by this data point.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
```diff
@@ -0,0 +1,160 @@
...
+    self.accepted_or_declined_by = portal_user
+
+    self.save
+  end
+
+  def decline!(portal_user, reason)
+    terminate
+    self.status = 'declined'
+    self.accepted_or_declined_by = portal_user
+    self.decline_reason = reason
```
nihil2501, line 50: I suspect that this could easily have sensitive information. Particularly if there is a free-form input option.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Column level encryption
+The [blind index gem](https://github.com/ankane/blind_index) allows for exact matching queries on encrypted columns. This will be used on fields that users will search or filter by. Existing KMS encryption will be used on the fields that are for display purposes only.
+
+#### The Model
+```ruby
+class PowerOfAttorneyRequest < ApplicationRecord
+  has_kms_key
+  has_encrypted :consent_data, key: :kms_key, **lockbox_options
+  blind_index :first_name, :last_name, :city, :state_code, :zip_code
```
nihil2501, line 25: I am skeptical that the product needs to support "VSO can filter by all 'Adam's". I believe product does want to load data for a particular user. For this there might be a part of the app that requires a sufficient number of inputs to uniquely identify a Veteran/claimant that a VSO could enter that we could get back to the user from using MPI. SEP's version of this feature is described in SEP FAQs [here](https://www.knowva.ebenefits.va.gov/system/templates/selfservice/va_ssnew/help/customer/locale/en-US/portal/554400000001018/content/554400000008534/61-Search-for-a-Veteran).

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
```diff
@@ -0,0 +1,160 @@
...
+  def expire!
+    terminate
+    self.status = 'expired'
+
+    self.save
+  end
+
+  private
+
+  def terminate
```
nihil2501, line 64: It might be nice to make another table that has our ephemeral data. The presence or absence of an association to it can be coupled to whatever our business logic is for "still searchable" (inert due to already being settled or expired). This idea of no-longer-searchable POA requests calls into questions existing designs that surface settled POA requests in the POA request search results table.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+### Short term persistence
+Due to the sensitive nature of this data, the various encrypted fields will only be stored for XX days. After XX days or a terminal status is reached (i.e. the request is accepted) only metadata will be kept in the database.
+
+#### Cleanup Job
+This would be a scheduled job and could run on any cadence. The first choice would be once per day to match `InProgressFormCleaner`.
+
+```ruby
+class PowerOfAttorneyRequestCleaner
```
nihil2501, line 128: I have been curious about other approaches to expiration. Sidekiq is probably fine, but we could be somewhat nervous that Sidekiq is offline for "too long". `pg_cron` to poll from within the DB might be slick, but I'm not sure.

nihil2501:
Nice document you've produced here @jvcAdHoc !

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
nihil2501, line 64: It's possible that a model lifecycle hook is a nice way to let us more confidently ensure coverage on termination. "if such and such is changing now, do termination". Active Record APIs have all the conveniences for this idea if like it.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
```diff
@@ -0,0 +1,160 @@
...
+The model helper methods `accept!` and `decline!` should be used when VA.gov users take terminal actions.
+
+## Risks
+- `blind_index` is susceptible to leakage in the same way any deterministic encryption method is.
+- Mixing elevated PII fields (name + location) makes this dataset a more appealing target to attackers.
+- The medical consent limitations may be considered PHI. However, one could argue that limiting access to certain types of records does not necessarily mean a user has those medical conditions.
+
+## Future Considerations
+- Encryption in transit is not considered in this document. That may be an additional security concern for those sending data to VA.gov.
+- Data access policies (i.e. Pundit) are not considered in this document. That may be an additional security concern for those implementing Accredited Representative Portal users.
```
nihil2501, line 151: Good point.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
```diff
@@ -0,0 +1,160 @@
...
+#### Terminal Actions by Users
+The model helper methods `accept!` and `decline!` should be used when VA.gov users take terminal actions.
+
+## Risks
+- `blind_index` is susceptible to leakage in the same way any deterministic encryption method is.
+- Mixing elevated PII fields (name + location) makes this dataset a more appealing target to attackers.
+- The medical consent limitations may be considered PHI. However, one could argue that limiting access to certain types of records does not necessarily mean a user has those medical conditions.
+
+## Future Considerations
+- Encryption in transit is not considered in this document. That may be an additional security concern for those sending data to VA.gov.
```
ddzz, line 150: Do we need anything beyond standard HTTPS/TLS for encryption in transit?

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
jvcAdHoc: There were a few things like this that were tough to make a call on without a final product decision. Maybe a good compromise would be to keep the consent data jsonb for the specific limitations and make a boolean `has_health_record_limitations` that is `true` if any of the 4 limitations are present at record creation.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
jvcAdHoc: For storage purposes then, switching first and last to KMS would be good? We lose the easy search/filter but would still be able to display the name without an MPI call for each Veteran.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
jvcAdHoc: I like the idea of a better implementation than this. I don't want to be too prescriptive and make people think that everything must be implemented exactly how it is in this document. I hoped `Code below is for illustrative purposes. These snippets are not likely to be the final implementation.` might provide some cover.

vets-api already has [aasm](https://github.com/aasm/aasm) and I've liked using that in the past. I can rewrite some of the model to be more state machine like.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
jvcAdHoc: I honestly don't know if there are different security needs/requirements for reading this data from the front end due to the resultant sets including many Veteran rows instead of a single Veteran.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
raudabaugh: It would simplify things greatly if we dropped support for searching settled requests. Maybe that can be a post-MVP feature.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
raudabaugh: @nihil2501, is your suggestion to use `mpi_service.find_profile_by_attributes(first_name:, last_name:, birth_date:, ssn:)` instead of blind indexing to find a single record?

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
jvcAdHoc: To me, it seems allowing settled requests to be searchable contradicts the data retention plan. Unless the intent is for them to be searchable for XX days post settlement?

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
nihil2501: I think there are two features of the app that concern the details of a single Veteran. One is the POA request detail view. For this we would load the POA request by ID, which in turn is associated to a user that has an ICN which we could use to retrieve Veteran details from MPI.

The other is like the one linked in SEP where the VSO enters in fields including SSN and DOB that I assume they have on hand in this flow. A query of MPI with those fields would yield an ICN that we could use to find a corresponding Veteran user's POA requests. I don't think this feature has been mocked up yet though. Also I'm curious about the same question you have about the required field set in the SEP instructions I linked.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
nihil2501: Sounds good!

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
holdenhinkle: I think Sidekiq should be fine. vets-api relies on Sidekiq for some many things, it autoscales, etc.

Also, I've never used `pg_cron` so I can't comment on it off the top of my head.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
nihil2501: Yeah that makes sense about product ambiguity. And you've called out that this document is meant to be sufficiently illustrative about security concerns. It's possible that these data points correctly fit within that idea of fields (possibly in their own table) that make the POA request non-inert / searchable / actionable and can get the blind-index + expiration treatment.

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
```diff
@@ -0,0 +1,160 @@
...
+      t.string :first_name_bidx
+      t.string :last_name_bidx
+      t.string :city_bidx
+      t.string :state_bidx
+      t.string :zip_code_bidx
+
+      t.text :consent_data_ciphertext
+
+      t.string :status
+      t.datetime :expires_at
```
holdenhinkle, line 92: I'm wondering if we need `expired_at`?

Might it be more flexible to set a constant for the number of days a request expires after `created_at`?

products/accredited-representation-management/technical-documentation/appoint-a-representative/poa_request_proposal.md
holdenhinkle: Off the top of my head, is this what the two options would look like?

### One table
Table: `power_of_attorney_requests`
#### Search:
1. Search for requests with
    - pending status
    - not expired

#### Update: 
1. Update status and terminate the request

### Two tables (ephemeral record option)
`power_of_attorney_requests` and `active_power_of_attorney_requests`
#### Search: 
1. Search `active_power_of_attorney_requests` for requests that are not expired (just in case some requests have expired but the record hasn't been deleted from the table yet)

#### Update:
1. Update status in `power_of_attorney_requests`
2. Delete the record in `active_power_of_attorney_requests` which deletes the association

holdenhinkle (commented):
Great job Jon. Very thorough.

PR #90821, opened 2024-08-15
Creates new Spikes & Decisions folder + adds the interstitial spike document

asg5704:
- [x] Create a new `Spikes & Decisions` folder in our identity director
- [x] Added a Spike document for the interstitial technical analysis


products/identity/Spikes & Decisions/2024-08-15_interstitial-spike.md
```diff
@@ -0,0 +1,79 @@
...
+- **Can we interrupt the flow of users who have logged in with DSLogon, but no further - identifiers?**
+
+  - Unauthenticated
+    - Possible: Yes
+    - How: On button click in the sign in modal/page
+  - Authenticated
+    - Possible: Yes
+    - How: Lookup in Redux `user.profile.signIn.serviceName`
+
+- **Can we interrupt the flow of users who have logged in with MyHealtheVet AND who are known to have a Login.gov/ID.me account that they are not using?**
```
joeniquette, line 29: im not a big fan of this question/request. Where did it come from? Perhaps talking more about opinions on how this may or may not work from a user experience standpoint is warranted but in this document we can leave it at the technical level.

products/identity/Spikes & Decisions/2024-08-15_interstitial-spike.md
```diff
@@ -0,0 +1,79 @@
...
+    - Possible: No
+    - Why: No way to determine unauthenticated
+  - Authenticated
+    - Possible: Maybe
+    - How:
+      - Option 1. Significant BE work would be required to determine if the user has a Login.gov/ID.me account associated (eg. creating new tables & associations)
+      - Option 2. Significant BE work to utilize the ID.me wrapper to determine if an ID.me account is associated with the user.
+- **Are there differences in what we can serve when it is a modal vs a full page?**
+
+  - While there should not be significant differences, there is the possibility that - accessibility issues could arise from focus switching between modals. Additionally, - there could be unexpected issues if we go from sign-in modal => interstitial page => - user clicks back button/link to take them back to select a different CSP => sign-in - modal. It could be a little confusing, and jarring for the user.
```
joeniquette, line 63: I think we need to expand on this further. Are there any restrictions as to secondary login button pages from USiP vs modal. For example if we make "use legacy creds" go to a different login page/view..is this possible or do we run into significant issues...and does it make a difference ifs its a modal vs usip. 

One other thing for modal vs usip capabilities...on the callbacks for either sis or ssoe if we want to show a message to those mhv only or dslogon only users, do we have any concerns regarding interrupting the flow with a new page or a pop up or whatever?

Its ok to state these arent in the scope of this ticket and we can make a follow on ticket. But this is the mindset we need to be in to ensure we provide options to the comms teams.

products/identity/Spikes & Decisions/2024-08-15_interstitial-spike.md
asg5704: These were questions that came from the ticket that Parker created. I think both should be warranted and documented.

products/identity/Spikes & Decisions/2024-08-15_interstitial-spike.md
asg5704: @joeniquette This is hard to answer because all of it depends. Can we link up for this after standup?

asg5704: @joeniquette added the interstitial-logic.png from the sync we had yesterday (8/21)

claytonzook (approved):

Issue #83023, opened 2024-05-14
Access to SOCKS for Lighthouse employees


RachalCassity:
### Summary
Currently, users that need SOCKS access have to go through the Platform Orientation. Lighthouse Engineers and engineers from other services should not have to go through Platform Orientation. 

We need to ensure that the SOCKS access template has an option for a Lighthouse resource.

There may be some training updates that need to be shared with the DevOps CoP and the Support team. 

## Slack Thread
- [Access to SOCKS for Lighthouse Engineers](https://dsva.slack.com/archives/CBU0KDSB1/p1715030438257559)

## Tasks
 - [x] Find out if there is a Lighthouse roster that can be checked.
 - [x] Review and modify [SOCKS access issue template](https://github.com/department-of-veterans-affairs/va.gov-team/issues/new?assignees=&labels=external-request%2Coperations%2Cops-access-request&projects=&template=socks-access-request.yml&title=SOCKS+access+for+%5Bindividual%5D)
 - [x] Update `.github/workflows/pst-amt-roster-check.yml` (devops repo)
 - [x] Updated Platform Website [documentation](https://depo-platform-documentation.scrollhelp.site/getting-started/accessing-internal-tools-via-socks-proxy#AccessingInternalTools(viaSOCKSProxy)-PrerequisitesPrerequisitesforSOCKSProxyconfiguration)
 - [x] Update this [internal documentation](https://vfs.atlassian.net/wiki/spaces/OT/pages/1792114735/Onboard+New+Team+Members+Granting+Access+to+Internal+Tools#SOCKS-Access-Request-(adding-SSH-public-keys))
 - [x] Share changes with the DevOps CoP and the Support team–they're the ones that process the access requests.

## Acceptance Criteria
 - [x] There is a path forward for LH SOCKS access 
 - [x] Platform Website [documentation](https://depo-platform-documentation.scrollhelp.site/getting-started/accessing-internal-tools-via-socks-proxy#AccessingInternalTools(viaSOCKSProxy)-PrerequisitesPrerequisitesforSOCKSProxyconfiguration) is published with the updates.
 - [x] Relevant parties notified of changes

## Validation
- create new [SOCKS access ticket](https://github.com/department-of-veterans-affairs/va.gov-team/issues/new?assignees=&labels=external-request%2Coperations%2Cops-access-request&projects=&template=socks-access-request.yml&title=SOCKS+access+for+%5Bindividual%5D) with the following:
[...]

PR #91088, opened 2024-08-20
Socks template update - replace checkboxes at bottom

rmtolmach:
These instructions/checkboxes at the bottom of the SOCKS template are a bit confusing–who are they for? What is the ask? This PR replaces those instructions (they live [in Confluence](https://vfs.atlassian.net/wiki/spaces/OT/pages/1792114735/Onboard+New+Team+Members+Granting+Access+to+Internal+Tools#SOCKS-Access-Request-(adding-SSH-public-keys)) so they're not gone forever) with two things that need to be verified by a human, which makes more sense for checkboxes.

_Note: The reason they're checkboxes is because Markdown does not get submitted with a yaml form (meaning that md fields can be seen by the person filling out the template but after they submit it, the md is gone)._

- Follow-up to https://github.com/department-of-veterans-affairs/va.gov-team/pull/90881 
### Related issue
-  https://github.com/department-of-veterans-affairs/va.gov-team/issues/83023

### DIFF
--- a/.github/ISSUE_TEMPLATE/socks-access-request.yml
+++ b/.github/ISSUE_TEMPLATE/socks-access-request.yml
@@ -103,9 +103,8 @@ body:
         :x: Don't fill out anything below here :x:
   - type: checkboxes
     attributes:
-      label: User must exist in a roster before SOCKS access can be granted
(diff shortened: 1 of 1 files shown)


rmtolmach:
I got a review from Chris. merging.

rmtolmach:
Validation:
This is what it looks like to a submitter:

image

This is what it looks like after it's been submitted:

image

PR #94487, opened 2024-10-08
Update url-list.md with additional routes

bacitracin:
I needed all the 526 routes for some testing I'm doing. Grabbed these urls from the following `vets-website` files
- https://github.com/department-of-veterans-affairs/vets-website/blob/main/src/applications/disability-benefits/all-claims/config/form.js
- https://github.com/department-of-veterans-affairs/vets-website/blob/main/src/applications/disability-benefits/all-claims/pages/toxicExposure/toxicExposurePages.js


bacitracin:
The team is now using a google spreadsheet, so no need to update this file as well.

PR #96223, opened 2024-10-31
[ART] Create pilot-2122-backend ADR

nihil2501:


products/accredited-representative-facing/engineering/ADRs/pilot-2122-backend.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Decision
+
+For the purposes of efficient pilot implementation, we'll first get an end-to-end 21-22 solution working where POA request information is stored in the VA.gov database, rather than making Lighthouse calls to persist that information elsewhere.
+
+- When a Veteran/VA.gov user creates a new POA request, the request will be saved in the VA.gov database, but no Lighthouse API calls will be made.
+- When an ARP user views POA requests, the VA.gov database will be queried for open requests, but no Lighthouse API calls will be made.
+- When an ARP user accepts or declines a POA request, a Lighthouse API call will be made to change the POA, using v2 of the Power of Attorney API. The VA.gov database will also be queried and modified to indicate that the POA request is closed.
```
raudabaugh, line 17: https://staging-developer.va.gov/explore/api/benefits-claims/docs?version=current

raudabaugh: 1. Does this API need to be modified to return the proc ID from BGS?
2. At what point(s) in development does the team anticipate being blocked by Lighthouse or BGS?

products/accredited-representative-facing/engineering/ADRs/pilot-2122-backend.md
```diff
@@ -0,0 +1,25 @@
...
+  - At first, we should implement a fully synchronous solution, where the VA.gov DB POA request is only set to accepted/rejected after the Lighthouse call is completed.
+  - If the fully synchronous solution is found to not be adequate, we can push the Lighthouse call to a Sidekiq job. In doing so, we have to be sure to follow No Silent Failures best practices.
+
+## Consequences
+
+Technical complexity will be greatly reduced for the pilot. Error handling and monitoring will be simplified, and we will have more control over UX responsiveness.
+
+We'll separately prioritize and roadmap adding the additional Lighthouse calls originally called for in the technical design. In this ADR, we're not proposing that we won't later add those Lighthouse calls. We're also not proposing a particular timeline for adding those calls. We'll depend on our Product Owner, Engineering Lead, and other VA partners for future decisionmaking and discussion there.
```
raudabaugh, line 25: I think it's worth discussing the risks of skipping CorpDB for request storage in our initial release. E.g. might there be conflicts with a 3rd party tool posting to `/forms/2122` on the [v1 API](https://staging-developer.va.gov/explore/api/benefits-claims/docs?version=v1)? Do we expect 3rd party tools to store requests in our database as well?

raudabaugh (changes requested):

products/accredited-representative-facing/engineering/ADRs/pilot-2122-backend.md
nihil2501: Responses from Alex (doesn't have GH access to reply himself yet):
> Does this API need to be modified to return the proc ID from BGS?

I don't believe proc id is relevant anymore to ART with this design, @jvcAdHoc can confirm.

> At what point(s) in development does the team anticipate being blocked by Lighthouse or BGS?

Right now - we currently have tickets assigned to engineers to implement the accept/reject flow, and a target of being feature complete by end of the year, which would give us 2 months to focus on setting the stage for the pilot and getting through hurdles. We also likely need to pull a bunch of our capacity to work on the claims flow by the beginning of next year if we're going to have a claim form in production by June. LH's roadmap currently envisions delivering the APIs by the end of the year, so the beginning of next year would be a good time to reevaluate.

products/accredited-representative-facing/engineering/ADRs/pilot-2122-backend.md
nihil2501: _Response from Alex (doesn't have GH access to reply himself yet):_
> might there be conflicts with a 3rd party tool posting to /forms/2122 on the [v1 API](https://staging-developer.va.gov/explore/api/benefits-claims/docs?version=v1)?


Scenario:
- Someone opens a POA request for organization A
- They then file a 21-22 using Tyler Tech or mail for organization B, which is processed by VA and changes their POA to organization B
- A POA manager from organization A goes to view POA requests

With the design in this ADR, the POA manager from organization A should then be able to accept/reject the request, potentially overwriting POA from organization A to organization B. 

It's possible that the LH APIs would have handled this situation differently — I'm not aware if they were designed to — but they could, when receiving a 21-22 via LH API, close any open POA requests from other organizations.

However, whatever LH APIs did would not apply to mailed in forms, which don't use LH APIs and should still be able to overwrite.

It's not clear to me what the best system behavior should be in the above scenario. 

Hopefully expiring POA requests after a period of time (currently 60 days) and adding the expiration period in copy when a user files a request will eliminate much of the possible confusion.

@raudabaugh @ITS-mharlow @nihil2501 @holdenhinkle @oddball-lindsay — anything I'm missing?

nihil2501: The most obvious race condition that comes to mind is:
```
ARM POA requested date < Tyler Tech POA established date
```
For this scenario, I think we have the opportunity to query for this condition during individual POA request operations (POA request detail view & POA request acceptance). It is also possible to poll on some cadence (even a relaxed one) so that the POA request list view doesn't misreport the situation often.

products/accredited-representative-facing/engineering/ADRs/pilot-2122-backend.md
jvcAdHoc: > Does this API need to be modified to return the proc ID from BGS?

Nope. We shouldn't need a proc id for a pure vets-api implementation.

jvcAdHoc: - If you're referring to LH's auto establishment endpoint, then there isn't a proc id to return. The auto establishment endpoints do not create poa request objects, they skip right to updating the necessary flags for an entity to have the access/permissions they need to act as power of attorney.
- This ADR proposes skipping the poa request creation endpoint which is where the proc id would come into play. TIL LH [already updated](https://github.com/department-of-veterans-affairs/vets-api/pull/19210) the creation endpoint to return proc id so we should be covered either way.

candilemoine:
_On Alex Prokop's behalf_

@nihil2501 — good point — just to confirm what I think you're suggesting:

- The POA manager clicks into a particular POA request on ARP
- We load POA history in that card view
- If we see that POA was updated for the requester since the POA request was filed, we display an extra warning to the POA manager before we let them accept the request

This might be the most optimal way to handle this situation regardless, as long as we don't have reason to mistrust POA manager motivations. It would potentially come at the cost of an extra trip to BGS which would add slowness to each request.

I suggest we file this in the backlog and decide whether to prioritize it later.

products/accredited-representative-facing/engineering/ADRs/pilot-2122-backend.md
candilemoine: _On Alex Prokop's behalf_

@nihil2501 — good point — just to confirm what I think you're suggesting:

- The POA manager clicks into a particular POA request on ARP
- We load POA history in that card view
- If we see that POA was updated for the requester since the POA request was filed, we display an extra warning to the POA manager before we let them accept the request

This might be the most optimal way to handle this situation regardless, as long as we don't have reason to mistrust POA manager motivations. It would potentially come at the cost of an extra trip to BGS which would add slowness to each request.

I suggest we file this in the backlog and decide whether to prioritize it later.

nihil2501: Yeah I think the desired business logic is an open question. One example is what you outlined above.  This could be the most simple if when the user is driven to a declination in this scenario, we don't need extra logic that accounts for the scenario. For example, would the set of declination notifications change in this condition.

Another would be forcing the impacted POA request into a "soft-terminal" state–some concept like "obsolesced by system" that, for example, we also leverage when newer ARM POA requests come (the currently proposed business logic says "only one open POA request at a time").

I agree that we should backlog this open question.

From an architectural standpoint, we should consider the concern of having important application state that lives in another system that we need to sync (the established POA in BGS) one that it is important to model and implement robustly. I think the possible approaches for this are a little messy but still within the realm of viability.

nihil2501: And to be clear, the Tyler Tech edge case (and the concomitant external application state synchronization concern) exists independently of this "shallow backend" proposal.

PR #97754, opened 2024-11-25
[10-10EZ] Added testing helper scripts for 10-10EZ Failure submission emails

coope93:
Added testing helper scripts for 10-10EZ Failure submission emails


allanto-ah (approved):
These testing steps worked for me. Looks good

PR #99656, opened 2024-12-26
Create assign-to-support-project.yml

bdech:
Workflow to automate GitHub issues for Platform Support Project board

### DIFF
--- a/.github/workflows/assign-to-support-project.yml
+++ b/.github/workflows/assign-to-support-project.yml
@@ -39,3 +39,11 @@ jobs:
       with:
         project: 'https://github.com/orgs/department-of-veterans-affairs/projects/1408'
         column_name: 'Access Requests'
+    - name: Add issues with `Offboarding` label to Platform Support project Access Request column
+      uses: srggrs/assign-one-project-github-action@1.2.1
+      if: |
+        github.event.action == 'opened' && 
+        contains(github.event.issue.labels.*.name, 'Offboarding')
+      with:
+        project: 'https://github.com/orgs/department-of-veterans-affairs/projects/1408'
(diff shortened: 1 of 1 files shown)


bdech:
New label added to access request issue template to aid in this automation

https://github.com/department-of-veterans-affairs/va.gov-team/pull/99658

https://github.com/department-of-veterans-affairs/va.gov-team/pull/99661

https://github.com/department-of-veterans-affairs/va.gov-team/pull/99659

.github/workflows/assign-to-support-project.yml
```diff
@@ -0,0 +1,41 @@
+name: Assign to Support Project board
+# Assigns new issues to project based on label
+# See https://github.com/orgs/department-of-veterans-affairs/projects/1408/
+# Organizes GitHub issue into respective column based on label
+
+on:
+  issues:
+    types: [opened]
```
rmtolmach, line 8: This should suffice 👍 
If an issue is opened without one of these labels and then one of the labels is added later, this action won't run, since it only opens on newly-opened issues. But I think that's okay because 'T1-access-request', 'new-vfs-team-member' and 'new-vfs-team' are added automatically when those specific issues are created.

rmtolmach:
@bdech do you want to make any rules for offboarding tickets, or do you not track them on your board?

.github/workflows/assign-to-support-project.yml
bdech: Thanks Rebecca! I added that to this workflow

bdech:
@rmtolmach Thanks Rebecca! I added that to this workflow

PR #102476, opened 2025-02-04
Update cc yaml request template to account for VA Mobile App

it-harrison:
This PR updates the Collaboration Cycle Request issue Template to prompt teams to specify if their products are intended for VA.gov or the VA Mobile app.


KKitagawa-Bosch (approved):
Looks good to me!

PR #102743, opened 2025-02-07
Suggested specifics for vague tasks + generalizing the sign in task

davidmpickett:


cindymerrill:
These revisions look great to me, @davidmpickett !

PR #103514, opened 2025-02-20
dbex/101351: add lost 2023 JourneyMap file

aurora-a-k-a-lightning:
Add Journey Map from 2023.


mayacarrolldesign (commented):
@ahay-agile6 We confirm the name of the file is correct. We think you are putting it in the correct location, but can't seem to verify this visually until the file is merged. The path looks right though.

PR #104934, opened 2025-03-11
Update appointments-content-brief.md

benbrasso:
Added appt reminders information bases on discussion with Privacy Office and details shared in https://dsva.slack.com/archives/C081Y4X4W1L/p1741203125332159

### DIFF
--- a/products/content/content-briefs/appointments-content-brief.md
+++ b/products/content/content-briefs/appointments-content-brief.md
@@ -37,7 +37,7 @@ This brief will help your team create content for your product that aligns with 
 
 Last updated: March 11, 2025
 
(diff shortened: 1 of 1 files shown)


benbrasso-agile6:
@outerpress I think I translated the feedback correctly. Mind giving that a second set of eyes? cc @DanielleThierryUSDSVA @mistymg

mistymg:
@benbrasso-agile6 My only suggestion would be instead of putting first names (Kay, Lauren, etc.), could we say the meeting was between Unified Appointment Experience team (full names) and the Privacy Office? That way anyone who comes in has the context for who these people are.

benbrasso-agile6:
Updated my PR based on Misty's feedback above

PR #107853, opened 2025-04-17
added table of contents

emilyd-oddball:


nhuckleberry6 (approved):
Love that you added a table of contents! Do you think linking to the transcripts under each heading would be overkill? Just curious

emilyd-oddball:
Transcripts would be a nice-to-have, but I would put that at the end of the priority list. A good idea for future projects though.

Issue #107875, opened 2025-04-17
Add guidance for closing Design Intent & Midpoint Review feedback tickets


shiragoodman:
## User Story
As a VFS team member, I want to know when to close the feedback ticket and whose responsibility it is to close it so that I am not leaving the ticket open indefinitely. 

Assignee: @allison0034 
Peer Reviewer: @erinrwhite 

## Description
Governance team has received several questions lately about whose responsibility it is to close the DI and MPR feedback tickets and if it is the responsibility of the VFS team, when can the ticket be closed. The purpose of this ticket is to add that guidance. 

The tickets can and should be closed by the VFS team, and the tickets can be closed post touchpoint, or whenever the feedback has been addressed by the team. We don't track these tickets, so they're free to close at any time if they so choose.

### Live pages
- [Collab Cycle Feedback](https://github.com/department-of-veterans-affairs/va.gov-team/issues/new?template=collab-cycle-feedback.md) issue template

### Drafted pages
- 

## Tasks
- [x] Update the template to include guidance for VFS Teams, specifying that they close the ticket and when they can close the ticket.

## Peer Review
_To be completed by peer reviewer_
- [x] User story and acceptance criteria are met

## Acceptance Criteria
- [x] Collaboration Cycle Feedback template has been updated to include guidance for closing the ticket

## Team Notification
- [ ] Slack post on #platform-governance-team
- [ ] Presentation at Weekly Initiative Meeting
- [x] No notification needed

### Refinement
- [x] Ticket has user story, description, tasks, and acceptance criteria
- [x] Ticket has been sized 
- [x] Once all above items are checked, add the 'Ready' label.
### Planning
_If this ticket is picked up from the Backlog mid-sprint, connect with Shira to ensure the below items are completed correctly_
- [x] Ticket has been discussed as a team at Planning
- [x] Ticket has been assigned to appropriate initiative or quarterly epic
- [x] Ticket has been assigned to a Sprint
- [x] Ticket has been moved to the Planned pipeline in the Governance team board
- [x] Ticket as an assignee and peer reviewer

erinrwhite:
Screenshot of current guidance - we already tell teams to close the ticket "when all feedback has been addressed"
![Image](https://github.com/user-attachments/assets/d7547c45-0994-4bd9-9328-77307c220d9f)

We could update this to "any time after the ticket has been reviewed and/or addressed." @allison0034 ?

erinrwhite:
[...]

PR #108590, opened 2025-04-29
Update closing instructions on collab cycle feedback template

erinrwhite:
Per #107875 updating instructions for VFS teams to be clearer about when they should close the ticket.


allison0034 (approved):
Looks good!

Issue #108716, opened 2025-04-30
[Spike] Update cancel documentation


outerpress:
## Background/Request

In [this thread](https://dsva.slack.com/archives/C023EFZPX4K/p1744312821452759), Kay asked us to document the cancelation documentation [here](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/appointments/va-online-scheduling/feature-reference/tools/tool-cancel.md#requirements).

Please review the requirements section of that document and:

- update any technical details that are out of date
- add all scenarios for the cancelation flag for OH and VistA appointments

### Requirements to Consider

---

## Tasks
- Answer the following questions:
 * How does cancellation work logic-wise?
 * What should a non-technical stakeholder understand about cancellation?

## Time Box
3  hours

## Definition of Done
- Tasks are complete
- Research and recommendations are documented in this ticket

---
## How to configure this issue
- [ ] **Attached to an Epic** (what body of work is this a part of?)
- [ ] **Labeled with Practice Area** (`backend`, `frontend`, `devops`, `design`, `research`, `product`, `ia`, `qa`, `analytics`, `contact center`, `research`, `accessibility`, `content`)

outerpress:
Thanks @JunTaoLuo ! This clarifies things nicely for me. I'll go ahead and close it out, and ping that thread.

PR #109448, opened 2025-05-09
Update cancellation doc

JunTaoLuo:
Updating and reorganizing VAOS cancellation doc.
- I've confirmed the accuracy of the Webapp and API requirement sections 
- We'll need help from the backend to confirm accuracy of the Backend requirement section, but I've tried to capture all that I can from previous the previous doc and slack conversations.
- Confirmed that we don't have any OH specific cancellation logic so they are handled like every other appointment

Related issue: https://github.com/department-of-veterans-affairs/va.gov-team/issues/108716


ala-yna (approved):
This looks great! Thank you John :)

PR #110786, opened 2025-05-29
Create Vets API Integration.md

theodur:
## Description Vets API Integration Writeup

### DIFF
--- a/products/vet-transition-support/engineering/Vets API Integration.md
+++ b/products/vet-transition-support/engineering/Vets API Integration.md
@@ -1,0 +1,48 @@
+# Vets API Integration Discovery
+
+## Overview
+
+The intent of this writeup is to explore the technical requirements, reasonability, and lift for integrating the BDS API with the `Discover your Benefits` app on VA.gov.
+
+## Vets API Overview
+
+All apps on VA.gov that rely on any API have to go through [vets-api](https://github.com/department-of-veterans-affairs/vets-api), the API that powers VA.gov applications. This is because of how VA.gov is configured. The [Cross Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS) for the VA.gov server is setup to only allow requests from [api.va.gov](https://api.va.gov/). This means attempting to access an upstream API service (i.e. one hosted via lighthouse, e.g. BDS) directly isn't possible, for more than one reason.
+
+This means that in order to integrate with the BDS API, we would most likely need to create a module in the `vets-api` repo, which would essentially be it's own API that handles requests to and from the BDS API. Implementing this API in `vets-api` would require work from backend engineers familiar with Ruby on Rails.
+
+## Authentication
+
+Apps on VA.gov do not need to require users to be authenticated in order to use an API from `vets-api`. An example of this can be seen with the [Find VA locations](https://www.va.gov/find-locations) and [GI Bill Comparison Tool](https://www.va.gov/education/gi-bill-comparison-tool) applications. This means we could still manage an unauthenticated experience even when integrating with the BDS API.
+
+### Benefits of authentication
+
+Some potential benefits to having users be authenticated are:
+
+- Having some of their data already pre-populated, so we wouldn't need to ask them as many questions
+- Being able to save their recommended benefits to their account, so they would return and access them at a future time
+
+This would need to be verified by the backend engineer who works on the `vets-api` portion of this work.
+
+### Frontend authentication
+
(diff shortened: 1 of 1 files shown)


products/vet-transition-support/engineering/Vets API Integration.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Overview
+
+The intent of this writeup is to explore the technical requirements, reasonability, and lift for integrating the BDS API with the `Discover your Benefits` app on VA.gov.
+
+## Vets API Overview
+
+All apps on VA.gov that rely on any API have to go through [vets-api](https://github.com/department-of-veterans-affairs/vets-api), the API that powers VA.gov applications. This is because of how VA.gov is configured. The [Cross Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS) for the VA.gov server is setup to only allow requests from [api.va.gov](https://api.va.gov/). This means attempting to access an upstream API service (i.e. one hosted via lighthouse, e.g. BDS) directly isn't possible, for more than one reason.
+
+This means that in order to integrate with the BDS API, we would most likely need to create a module in the `vets-api` repo, which would essentially be it's own API that handles requests to and from the BDS API. Implementing this API in `vets-api` would require work from backend engineers familiar with Ruby on Rails.
```
macalemid, line 11: How big of a lift is creating the base of the app? Would we need another engineer in order to integrate?

products/vet-transition-support/engineering/Vets API Integration.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Another approach, similar to the first option is that after the user decides which workflow to proceed with, instead of routing them to the `Sign-in` application, we render the version of the app wrapped in the `RequiredLoginView`, that way we are enforcing authentication before they continue.
+
+- One last way would be to not ask the user whether they would like to sign in, and just check whether they are authenticated or not, and if they are authenticated, use the authenticated workflow.
+
+## Scalability of Vets API Integration
+
+By integrating `vets-api` into the app, we gain more scalability in the sense that there is room for additions to the app. We could eventually get to a place where we could have some sort of benefit portal where we help veterans track which recommended benefits they've applied to, potentially even showing status updates to them.
```
macalemid, line 40: I really like this idea!

products/vet-transition-support/engineering/Vets API Integration.md
```diff
@@ -0,0 +1,48 @@
...
+
+This works great for the authenticated workflow of our app, but becomes a bit more tricky accounting for both the authenticated and unauthenticated workflows. From my findings, there aren't any apps that support both authenticated and unauthenticated workflows, so our app would potentially be the first to do so. There are a few ways we could potentially handle this:
+
+- The [Sign-in](https://www.va.gov/sign-in) application on the VA.gov website essentially directs the user to the sign-in page so they can authenticate. We could have an entry page to the app that allows the user to choose whether they want to log in and continue the app (explaining the benefits of logging in) or continue unauthenticated, and based on that selection, we would navigate them to the `Sign-in` app so they can authenticate.
+
+- Another approach, similar to the first option is that after the user decides which workflow to proceed with, instead of routing them to the `Sign-in` application, we render the version of the app wrapped in the `RequiredLoginView`, that way we are enforcing authentication before they continue.
```
macalemid, line 34: This is an interesting approach. What are the pros and cons of this approach?

products/vet-transition-support/engineering/Vets API Integration.md
theodur: The BDS API doesn't have that many endpoints, so I'd imagine the module wouldn't be as large as many of the other APIs in the repo. That said, It would still be ideal to have a dedicated backend engineer working on this, as there's a good chunk of leg work for sorting out the API schema and making sure the data flow between our app and BDS is consistent. I feel like the structure I've seen with most VFS teams is they either have dedicated backend engineers on their team that handles the API portion of the app, or if their app is large enough, they will have a separate frontend and backend sub-teams

products/vet-transition-support/engineering/Vets API Integration.md
theodur: The pros would be:
- Easier to ensure they stay within the app after signing in. When routing them to the `Sign in` application, I'm not sure whether they'd be automatically routed back to the app that sent them there
- Implementation would probably be less cumbersome because we wouldn't need to send them to a different app for authentication, we would be supporting it natively in ours

The cons would be:
- We'd have to pay close attention to the state of the app when the user selects an option, because if they were to navigate back to the options screen after already having selected an options, we'd want to make sure we reset the state of their initial choice. And we'd also need to figure out how to handle if the user tried to supersede the option page that asks them whether they'd like to sign in or not. (e.g. do we automatically kick them back to the options page if they haven't selected an option? Do we just show the unauthenticated version of the app?) So it would require more thinking and logic on our end to handle edge cases like that

Good question, I'm gonna add what I said above to the doc, along with pros and cons for the first approach

macalemid (approved):

Issue #105073, opened 2025-03-12
[Documentation] Identify / create dev documentation


jonathanjnelson:
## Description
This is a follow-up ticket to [ticket #101536](https://github.com/department-of-veterans-affairs/va.gov-team/issues/101536). And it supplements ticket [#82786](https://github.com/department-of-veterans-affairs/va.gov-team/issues/82786), which documents technical logic on the landing page.

## User Story
As a Cartography team member, I want to identify the engineering documentation that we consider most important for future work, so that we or the next contract team can easily find them. 

## Notes
- The team has created the [My HealtheVet on VA.gov transition documentation](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/digital-health-modernization/mhv-to-va.gov/documentation-for-health%20portal.md) MVP

- We know need to add a more robust [engineering section.](https://github.com/department-of-veterans-affairs/va.gov-team/blob/master/products/health-care/digital-health-modernization/mhv-to-va.gov/documentation-for-health%20portal.md#engineering-resources)

- This documentation should exclude anything covered by ticket [#82786](https://github.com/department-of-veterans-affairs/va.gov-team/issues/82786), which documents technical logic on the landing page

- This includes, but is **not necessarily limited to**, the AC API, Supply Reorder, Secondary Nav, Route Guard, etc.

## Things to Consider
- What would someone want to know coming to this space?
- Where are we now (what exists, cohesive health portal, goals for 2025, milestones)
- Relevant Murals (including api logic of health portal), etc

## Possible Tasks
- Pull together the resources that we think exist (perhaps from [this link](https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/products/health-care/digital-health-modernization/engineering)?)
- Write summaries where they make sense
- Add to MD file (as an addition to [this page](https://github.com/department-of-veterans-affairs/va.gov-team/tree/master/products/health-care/digital-health-modernization/engineering))

## Acceptance Criteria
- [ ] Documentation added to MD file 
- [ ] Completed peer review and additional items that are missing

michaelbrodsky-a6:
Hey @jonathanjnelson @fmccaf1 @sterkenburgsara @athenabozak – do you think this ticket would still be helpful now that [#82786](https://github.com/department-of-veterans-affairs/va.gov-team/issues/82786) has been prioritized by the devs?

fmccaf1:
[...]

PR #111120, opened 2025-06-03
[105073] Add engineer section of MHV documentation

nichia:
## Summary

Add to engineer section of the MHV documentation

## Related issue(s)

- department-of-veterans-affairs/va.gov-team#105073


dcloud (approved):
Great work overall, thanks for doing this. In particular, I like how you organized things under "Application and Shared Function Documentation"

I have a couple of minor thoughts/ideas:

1. I'm so used to "No VA patient facility registration" itself being an alert that I wonder if we should include a brief note about the non-patient version of the landing page. Or is this described in a document that is linked to this one?
2. The Alert flow could be a [mermaid diagram](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams) if you wanted something more visual. Doesn't have to be, but mentioning that in case it hadn't occurred to you.
3. Looking at the "Application and Shared Function Documentation" section with the "*Path:"… did you consider linking to the file being mentioned? Not sure if that's worth the effort or not

PR #114074, opened 2025-07-10
Update product-outline-template.md

humancompanion-usds:
Per request from Chris & Ryan, adds a question about mobile so that we can remember to talk about mobile app during PO Sync.


rtwell (commented):
Can we and an ", and why?" at the end? it would be helpful to know how they decided for or against instead of just a "no".

PR #114077, opened 2025-07-10
Update initiative-brief-template.md

humancompanion-usds:
Adds a question about the mobile app. Fixes some formatting.


rtwell (commented):
I'd like to add an ", and why?" here as well

Issue #114229, opened 2025-07-14
discovery(526): Document why cypress did not catch submission validation errors


LinnJS:
## Description

Before proceeding with implementation, we need to conduct a focused discovery effort to understand **why the Cypress maximal test did not catch regressions** when required pages were removed or broken. 

This discovery work will inform the necessary changes to how we follow test patterns, and will include a Technical Analysis, a Test Gap Analysis, and a Preventative Measures section so outline how best to avoid this kind of failure in the future. 

## Goals

* Create an analysis of the gaps in test coverage that took place, and how the [Post Mortem](https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/blob/master/Postmortems/2025/2025-06_30_Missing_Conditions_Page.md) work after the incident fixed the issue.
* Determine whether there are **gaps in the maximal test's current expectations** that need addressing.
* Clarify the failure modes to ensure we are fixing the right problems in implementation.

## Acceptance Criteria

* [x] Review the current implementation of `cypress.helpers.js` (see [reference line](https://github.com/department-of-veterans-affairs/vets-website/blob/main/src/applications/disability-benefits/all-claims/tests/cypress.helpers.js#L300)) and identify why it did not fail during the problematic PR test run.
* [x] Analyze the Cypress maximal test to confirm whether it explicitly checks for all required pages being rendered.
* [x] Reproduce the test run against commit `8b89640` and capture detailed logs to understand the behavior.
* [x] Document findings in a **root cause analysis** (add to ticket comments or link to Confluence).
* [x] Provide a clear recommendation for next steps (e.g., helper refactor, new assertions, test coverage expansion).

ARPerry87:
I started on this, and will link up with Alex next week so that we can dig further! I'm excited about working on this ✨

ARPerry87:
Link to the document: https://github.com/department-of-veterans-affairs/va.gov-team/pull/116209

ARPerry87:
The PR attached to this will be getting an overhaul to follow [this slack](https://aquiaworkspace.slack.com/archives/C08QHMTCATE/p1754666051805969) convo based on the knowledge that Team 1 finished the actual testing and code, while our document will serve as a way to discuss preventative measures and future testing patterns.

PR #116209, opened 2025-08-06
Discovery: Document why cypress did not catch submission validation errors

ARPerry87:
## Summary
Documentation of why Cypress testing did not catch submission validation errors

## Related issue
- https://github.com/department-of-veterans-affairs/va.gov-team/issues/114229

### DIFF
--- a/products/disability/526ez/engineering_research/526_cypress_not_failing_in_regression.md
+++ b/products/disability/526ez/engineering_research/526_cypress_not_failing_in_regression.md
@@ -1,0 +1,205 @@
+# Discovery Document: [Disability Benefits/Discovery: Document why cypress did not catch submission validation errors]
+
+## Overview
+
+Before proceeding with implementation, we need to conduct a focused discovery effort to understand why the Cypress maximal test did not catch regressions when required pages were removed or broken. The core issue involves the cypress.helpers.js file not failing as expected, but the root cause remains unclear.
+
+This discovery work will inform the necessary changes to the maximal test suite and prevent premature fixes without fully understanding the problem.
+
+- Owner: Alicia Perry
+- Date: 2025-08-06
+- Related Docs: [Github issue]([https://github.com/orgs/department-of-veterans-affairs/projects/1660/views/5?pane=issue&itemId=119831819&issue=department-of-veterans-affairs%7Cva.gov-team%7C114229]
+)
+
+## Goals
+
+To understand this failure, and to figure out how to best formulate and understand testing in order to avoid regressions like this in the future. Then, once understanding has been reached, we will then want to create guidelines and suggestions for the future testing patterns for Cypress.
+
+- Discover what exactly caused the regression and what tests should have caught the error.
+- Discover how to adjust the test suite and patterns so that we pick up the failures more robustly.
+
+## Problem Statement
+
+We want to figure out how to tweak our test suite to catch such failures in the future so that it doesn't take down the form.
+
+- in July, a [commit](https://github.com/department-of-veterans-affairs/vets-website/commit/8b896400558334f2aee0799595c1d6f847f96556) was released and merged into main of the vets-website repo. This change caused a regression and an overall down period that was discovered by a PM looking at the VA subreddit. From there, members of the Disability Benefits team swarmed the problem, and tracked down the above commit as the reason for the failure. It was noted that there was a failure on the CI/CD to pick up this regression, having passed both Cypress and Unit Tests. This ticket and discovery are set out to discover how to make it to where we can test for such failures again, and how to avoid running into this situation once more.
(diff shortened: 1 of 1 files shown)


products/disability/526ez/engineering_research/526_cypress_not_failing_in_regression.md
```diff
@@ -0,0 +1,92 @@
...
+
+- in July, a [commit](https://github.com/department-of-veterans-affairs/vets-website/commit/8b896400558334f2aee0799595c1d6f847f96556) was released and merged into main of the vets-website repo. This change caused a regression and an overall down period that was discovered by a PM looking at the VA subreddit. From there, members of the Disability Benefits team swarmed the problem, and tracked down the above commit as the reason for the failure. It was noted that there was a failure on the CI/CD to pick up this regression, having passed both Cypress and Unit Tests. This ticket and discovery are set out to discover how to make it to where we can test for such failures again, and how to avoid running into this situation once more.
+- All veteran users on the VA attempting to complete or begin the disabilities process, form 0781 and 526ez, were faced with 5XX and 4XX errors in regards to attempting to route to their respective pages.
+
+## Discovery Questions
+
+Key questions we need to answer during this phase.
+
+- What was the likely root cause inside of the changes within the commit?
```
tommasina-va, line 32: The root cause can be found in the postmortem document here: https://github.com/department-of-veterans-affairs/va.gov-team-sensitive/blob/master/Postmortems/2025/2025-06_30_Missing_Conditions_Page.md

ARPerry87: Omg thank you so much! I couldn't find that, and was trying to wrap my head around it. I'll post this in the ticket as well! You rock <3

products/disability/526ez/engineering_research/526_cypress_not_failing_in_regression.md
```diff
@@ -0,0 +1,92 @@
...
+- It is possible that the object being returned to the submit flow was a different object or contained a different type that didn't match with what the backend expected to receive. As almost all OOP languages use objects as a form of message between methods, views, and the backend that contain a payload that is being validated against, should any part of the payload not match the expected message, it will cause a failure.
+- One of the proposed causes was that PTSD was being called by the term 'condition' as opposed to 'name', which is what is expected in a user or patient that already has the existing, historic condition, as outlined in: ratedDisabilities has a name not a condition in src/applications/disability-benefits/all-claims/tests/initialData.js:41 and in src/applications/disability-benefits/all-claims/tests/utils/form0781.unit.spec.jsx:49. This type difference, or even key name difference, could have caused the failure. So when, how do we test for it in a way that is meaningful and would cause the CI/CD to fail in a way that would restrict a similar regression from happenign again.
+
+- Solution brainstorming
```
tommasina-va, line 50: It might be helpful to review the latest PR resolving the initial issue: https://github.com/department-of-veterans-affairs/vets-website/pull/37482/files In it, I added a detailed step-by-step addition to the Cypress test which now fails if the follow-up conditions page isn't shown.

tommasina-va: Sorry that our work seems to have overlapped. I'm not sure why you all had the testing ticket to begin with. What I think could be really useful is to establish a pattern for breaking out Cypress tests into easier chunks for writing more specific use-case assertions. Maybe something like what's in `src/applications/discharge-wizard/tests/cypress/flows/discharge_unjust_flow.cypress.spec.js`?

tommasina-va (changes requested):
This is a good start! I'd recommend updating to reflect the information from the postmortem and the current state of the Cypress tests. 

A few other thoughts:
The initial work completed involved the form config page which establishes the form flow. There isn't really a good way to unit test that, unless methods are extracted as utility functions for unit testing, which still limits scope and doesn't reflect the overall form flow. That approach (which was in the initial ticket) in combination with the new updated Cypress test could have caught this issue. The Cypress test is sufficient in this case however, since we are just looking to make sure the condition follow up pages (which add the data required to submit) show up as expected. 

I liked the questioning of validation on the review and submit page. Why didn't users see an error message on each listed condition about the missing follow up data? I'm not sure that's within the task of this testing discovery ticket, but could be useful to understand down the line.

products/disability/526ez/engineering_research/526_cypress_not_failing_in_regression.md
ARPerry87: That's a marvelous idea! And something we've actually been discussing a lot on Team 2, which is awesome that the two teams are both kind of working in sync with each other like this. And I don't mind the overlap at all! I just started on the project like a month and a half ago, so this helped me understand some of the depth of our project.

tommasina-va (approved):
Great refactors. Really liked seeing the code examples and ways to make the tests stronger in the future. One tweak is just to take my name out of it. These are process improvements we're making as a team, so I don't need any extra credit.

PR #118628, opened 2025-09-04
Update create-accessible-PDF.md

SarahKay8:
Martha's ask was to write up ADE's PDF audit process


artsymartha68 (approved):
Looks fabulous, thanks for asking me!

PR #118773, opened 2025-09-05
Create discovery-claim-title-mapping.md

williamphelps13:
Issue: [#117659](https://github.com/department-of-veterans-affairs/va.gov-team/issues/117659)

### DIFF
--- a/products/claim-appeal-status/engineering/discovery-claim-title-mapping.md
+++ b/products/claim-appeal-status/engineering/discovery-claim-title-mapping.md
@@ -1,0 +1,812 @@
+# Discovery: Claim Title Mapping Implementation Refactoring
+
+Issue: [#117659](https://github.com/department-of-veterans-affairs/va.gov-team/issues/117659)
+
+- [Problem Statement](#problem-statement)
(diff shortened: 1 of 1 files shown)


nathangthomas (commented):
Excellent job documenting and explaining this issue! You clearly demonstrated the user impact of the current problem.

I tend to lean toward more comprehensive do it once do it right refactors, but am also learning that approach has its place and this may not be it. I agree with your analysis here and support your hybrid approach. 

The technical gains aren't worth the additional complexity. As you said, option one adds an unfamiliar pattern for marginal performance gains and the current bottleneck isn't lookup performance. The real issue here is maintainability and accuracy which your proposed hybrid approach solves. 

My only pause was with the 100% test coverage target in the success metrics.  1% coverage is clearly inadequate, but jumping to 100% of all 1,696 + combinations seems expensive to maintain and I suspect we will start to see diminishing returns with many combinations being trivial formatting variations. In the realm of testing I think we can revise the success metric to focus on "meaningful coverage" rather than "exhaustive" testing.

PR #119204, opened 2025-09-10
Manifest products v1

humancompanion-usds:
This is the start of the products side of the Manifest tool that will replace the VFS Product Directory. The initiative is here: https://github.com/department-of-veterans-affairs/va.gov-team/issues/59141


naomimarcussen (approved):
Looks good!

PR #123338, opened 2025-10-27
Update collaboration-cycle-request.yml

allison0034:
Add touchpoints markdown


.github/ISSUE_TEMPLATE/collaboration-cycle-request.yml
erinrwhite: These need to be reversed - the details tag closed, then the `## Collaboration Cycle touchpoints` after that

PR #124241, opened 2025-11-04
686c picklist release plan

alwerner:


steele-lm (commented):
Is there any testing or coordination planned with RBPS to review the json payload submissions to ensure claims are processed (or off-ramped) as expected?

alwerner:
> Is there any testing or coordination planned with RBPS to review the json payload submissions to ensure claims are processed (or off-ramped) as expected?

@hnvt1989 will be payload testing during QA. We expect the payload to be the same before it gets submitted. If we're not comfortable releasing without testing with RBPS, we'll need to wait until the shutdown is over before releasing.

steele-lm (approved):

PR #125151, opened 2025-11-13
Billchapman fix shared research findings to slack

BillChapmanUSDS:
Refactored the task that shared research findings to Slack 
- Used a purpose specific Slack Token 
- Refactored the inline bash in to modular python scripts
- added tests for the python scripts (mostly copilot) 
- Successful test run is here: https://github.com/department-of-veterans-affairs/va.gov-team/actions/runs/19342339273 
- Fixed the JSON to escape interpolated variables properly 

See Shane Strassberg for use case and intent of the original feature.


sstrassberg:
@little-oddball Let me know if you have any questions about this!

PR #129958, opened 2026-01-13
[959] adds `lychee` link checking

bramleyjl:
This PR adds a `link-validation` Github workflow that checks all links (including HTTP) in updated documents and reports ones that don't work. The link checking is performed by [lychee](https://github.com/lycheeverse/lychee).

To test the lychee link-checking functionality modify (create, edit, move, rename, or delete) files in the `va.gov-team` documentation repository under the  `origin/959-fix-internal-links-lychee` Git branch & create a PR to trigger the workflow. New & changed files should be validated, deleted ones should be ignored. An example workflow can be seen [here](https://github.com/department-of-veterans-affairs/va.gov-team/actions/runs/21458123660/job/61803990330?pr=129958).

This check is non-blocking - its purpose is to alert documentation maintainers about which links need updating.

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -12,9 +12,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        env:
+          GIT_LFS_SKIP_SMUDGE: "1"
         with:
           fetch-depth: 0
           lfs: false
+          sparse-checkout: |
+            **/*.md
+            **/*.mdx
+            **/*.txt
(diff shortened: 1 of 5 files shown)


emilykim13 (approved):
Image

I created a PR with a new file that contained 3 links. In my first screenshot, the check was accurate with 3 proper links.

Then I made an update with 2 proper links and 1 broken link. Which can be seen in the second screenshot.

Image

This check is really neat and can be useful for a project like this, lgtm!

little-oddball (approved):

PR #130844, opened 2026-01-23
Creating documentation for the new VSC API endpoint

matt-guest-wilcore:
Documenting the old and new logic surrounding how we present the Veteran Status Card (VSC) in the mobile app. I outline the different paths we could face as well as noting all the possible SSC codes and their results.

Original Ticket: [1053](https://github.com/department-of-veterans-affairs/va-mobile-feature-support/issues/1053)


oddballdave (approved):
LGTM!

PR #131310, opened 2026-01-28
Update feedback process in Architecture Intent template

shiragoodman:
updated some verbiage re: ticket names and who to expect feedback from


erinrwhite (approved):
Thanks for these clarifying changes. Looking good to me.

PR #135245, opened 2026-03-05
Update veteran-status-api-logic.md

NatalieDavied:
Updated the PDF with an updated flow that include the correct content for ineligibility messages and updated status codes.


matt-guest-wilcore (approved):
lgtm!

Issue #132913, opened 2026-02-11
Automate finishing touches (phase 3)


shiragoodman:
## User Story
As a Governance team member supporting Collab cycle, I want to automate many of the manual steps currently part of our kickoff process so that we can ensure tickets are generated consistently without requiring manual steps to be completed by Governance. 

Assignee: @rsmithadhoc 
Peer Reviewer: @jasonday 

## Description
This automation phase 3 of the Collaboration Cycle Kickoff improvements and simplifying the framework includes misc. wrap-up, dependent on our touchpoint templates being created. This ticket includes the following:
- Dynamic numbering emoji
- Add code block content for Engineering/Security only SR.
   - Note: this may have to be skipped for now as there isn't currently a process for teams to follow if/when they need an Engineering/Security only SR. Eng/Sec folks are still addressing other high priority problems, and this is likely not one of them.
   - Are there other suggestions for how we might workaround this problem in the interim?
   - Per conversation at Sprint Planning on 3/11, we will populate the normal SR guidance if/when we determine a eng/sec only SR is needed. 
- Replace "Initiate" links with ticket links (Example: "**PO Sync ticket:** #131829")
- Update "Initiate (touchpoint)" links from draft links to real links, when those are ready. Define URL parameters to pass in (example: a default title).
- Add link to Collab Cycle ticket to touchpoint ticket when created via "Initiate (touchpoint)" link is selected
- Remove content in the post-submission ticket re: CMS Request under  "Other Required tasks" per ticket #135820.
   - update content in **Other required tasks** per [this thread](https://dsva.slack.com/archives/GNF5GSD9N/p1773239433690259). The designs are being updated via ticket #135943.
- Figure out the answers to these comments and address them in the process
    - Needs Shira's input: https://github.com/department-of-veterans-affairs/va.gov-team/pull/132894/changes#r2793939962
    - Needs finalization: https://github.com/department-of-veterans-affairs/va.gov-team/pull/132894/changes#r2794217699

### Live pages

_Live guidance docs to be be edited for this ticket_

- 

### Drafted pages

_All draft documentation created for this ticket_

- 

## Tasks
- [ ] _Placeholder task_

## Peer Review
_To be completed by peer reviewer_
- [ ] User story and acceptance criteria are met

## Acceptance Criteria
- [ ] We have automated the population of the dynamic numbering emoji for better ticket organization
- [ ] We have replaced the "initiate touchpoint" links with the link to the touchpoint ticket.
[...]

PR #136501, opened 2026-03-17
CC intake automation phase 3

rsmithadhoc:
https://github.com/department-of-veterans-affairs/va.gov-team/issues/132913#issuecomment-4118599514 for details


jasonday (approved):
@rsmithadhoc left one comment, which does not hold up approval. Nice work with the automations and the consideration of error states, etc.  Looks good.

PR #138380, opened 2026-04-03
Revise and expand incident response playbook

simiadebowale:
Updated the Appointments incident response playbook to simplify the on-call flow and make it easier to follow during incidents.


JunTaoLuo (approved):
💯

